PR #27, opened 2021-06-15
Fixes implemented

aleksMooncascade:


widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
```diff
@@ -216,31 +216,31 @@ public void startChat(
...
+                    chatHeadsController.onNavigatedToChat(
+                            new ChatHeadInput(
+                                    companyName,
+                                    queueId,
+                                    contextUrl,
+                                    this.theme
+                            ),
+                            activity instanceof ChatActivity,
+                            activity instanceof ChatActivity && useOverlays,
+                            controller.isMediaQueueingOnGoing()
```
aleksMooncascade, line 240: Is needed so the chatHeadsController knows of an ongoing media queueing. Start audio call / Start video call queueing.

widgetssdk/src/main/java/com/glia/widgets/glia/GliaQueueForEngagementUseCase.java
```diff
@@ -14,31 +19,70 @@
...
+    public void ticketLoaded(String ticket) {
+        typeOfOngoingQueueing = TypeOfOngoingQueueing.NONE;
+    }
+
+    public TypeOfOngoingQueueing getTypeOfOngoingQueueing(){
+        return typeOfOngoingQueueing;
+    }
+
+    public enum TypeOfOngoingQueueing {
+        NONE, CHAT, MEDIA
```
aleksMooncascade, line 86: Tells subscribers what type of queueing is ongoing when execute is called.

deniss-mc:
Will need some more work later, need to move engagement queue-ing prop to repository and all the null checks to use-cases

Will do for now :)

deniss-mc (approved):

PR #42, opened 2021-07-14
MUIC-237: File download

mustafa-ogun-ozturk:


app/src/main/java/com/glia/exampleapp/MainFragment.java
```diff
@@ -74,7 +74,7 @@ private void setNavigationIntentData(Intent intent, SharedPreferences sharedPref
         intent.putExtra(GliaWidgets.COMPANY_NAME,
                 Utils.getStringFromPrefs(R.string.pref_company_name, "", sharedPreferences, getResources()));
         intent.putExtra(GliaWidgets.QUEUE_ID,
-                Utils.getStringFromPrefs(R.string.pref_queue_id, getString(R.string.queue_id), sharedPreferences, getResources()));
+                Utils.getStringFromPrefs(R.string.pref_queue_id, "febd00bc-5ac6-4544-916b-b01fcb714549", sharedPreferences, getResources()));
```
deniss-mc, line 77: Don't forget to remove this

widgetssdk/src/main/java/com/glia/widgets/chat/ChatController.java
```diff
@@ -319,9 +322,9 @@ public void sendMessagePreview(String message) {
         }
     }
 
-    public boolean sendMessage(String message) {
+    public boolean sendMessage(String message, int numOfAttachment) {
```
deniss-mc, line 325: I don't think that adapterItemCount is required here

You could move the validity check to the sendMessageUseCase and introduce separate CallbackListener for that one

In order to check whether it has attachments to send you could ask this info from a AttachmentsProvider

deniss-mc: this way we will keep the controller clean :)

the sendMessageUseCase has 3 execution methods

basically one should remain execute(String message, CustomCallback callbacks)
with callbacks handling the success case, operator missing error etc. - you don't have to map the error case yet as it will be separate task in the future

others you can make private for now and map to them from this new execute(String message, CustomCallback callbacks)

as this method should be used in case of empty message:
public void execute(FilesAttachment filesAttachment, Listener listener)

widgetssdk/src/main/java/com/glia/widgets/chat/ChatController.java
```diff
@@ -673,56 +710,123 @@ private void changeDeliveredIndex(List<ChatItem> currentChatItems, VisitorMessag
...
+        List<SingleChoiceOption> singleChoiceOptions = null;
+
+        if (attachment instanceof SingleChoiceAttachment) {
+            singleChoiceOptions = Arrays.asList(((SingleChoiceAttachment) attachment).getOptions());
+        }
+
+        return singleChoiceOptions;
+    }
+
+    private boolean isMessageValid(String message, int numOfAttachment) {
```
deniss-mc, line 816: this one will be not needed after extraction to usecase

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
```diff
@@ -156,6 +164,9 @@ public void onItemRangeInserted(int positionStart, int itemCount) {
         }
     };
 
+    private Handler mainHandler = null;
```
deniss-mc, line 167: those two needs proper naming

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
```diff
@@ -1073,15 +1135,121 @@ private static Uri chooseUriByRequestCode(int requestCode, Uri galeryImgUri, Uri
...
+
+    private void downloadFile(OperatorAttachmentItem clickedItem, File file) {
+        if (clickedItem.attachmentFile.isDeleted()) {
+            Toast.makeText(this.getContext(), this.getContext().getString(R.string.chat_file_download_failed_msg), Toast.LENGTH_SHORT).show();
+            return;
+        }
+
+        Glia.fetchFile(clickedItem.attachmentFile, (fileInputStream, gliaException) -> new Thread(() -> {
+            try {
+                try (OutputStream output = new FileOutputStream(file)) {
```
deniss-mc, line 1164: extract this to utils

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
```diff
@@ -1073,15 +1135,121 @@ private static Uri chooseUriByRequestCode(int requestCode, Uri galeryImgUri, Uri
...
+        downloadFile(clickedItem, file);
+    }
+
+    private void downloadFile(OperatorAttachmentItem clickedItem, File file) {
+        if (clickedItem.attachmentFile.isDeleted()) {
+            Toast.makeText(this.getContext(), this.getContext().getString(R.string.chat_file_download_failed_msg), Toast.LENGTH_SHORT).show();
+            return;
+        }
+
+        Glia.fetchFile(clickedItem.attachmentFile, (fileInputStream, gliaException) -> new Thread(() -> {
```
deniss-mc, line 1162: there could be a FileProviderRepository made for that case - so that single object would be responsible for fetching the files

deniss-mc: tho I see that there could be problems with messageList atm. might need to sync later

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/ChatAdapter.java
```diff
@@ -151,10 +177,10 @@ public VisitorMessageViewHolder(@NonNull View itemView, UiTheme uiTheme) {
         public void bind(VisitorMessageItem item) {
             content.setText(item.getMessage());
             deliveredView.setVisibility(item.isShowDelivered() ? View.VISIBLE : View.GONE);
-            bindAttachments(item);
+            //bindAttachments(item);
         }
 
-        private void bindAttachments(VisitorMessageItem item) {
+        /*private void bindAttachments(VisitorMessageItem item) {
```
deniss-mc, line 183: just delete this

widgetssdk/src/main/res/layout/chat_visitor_message_layout.xml
```diff
@@ -33,7 +33,7 @@
         app:layout_constraintStart_toStartOf="@id/start_guideline"
         app:layout_constraintTop_toTopOf="parent" />
 
-    <androidx.recyclerview.widget.RecyclerView
+    <!--<androidx.recyclerview.widget.RecyclerView
```
deniss-mc, line 36: delete this

deniss-mc (changes requested):

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
mustafa-ogun-ozturk: In this case, too many listeners need to be involved since the use case cannot be called from ChatView directly. I kinda do not want to do it since it just creates unnecessary complexity.

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
mustafa-ogun-ozturk: I think this naming is ok because it represents the main thread?

deniss-mc: oki, let it be for now

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
deniss-mc: it could react on the itemslist state change aswell invoked after files get fetched

but its not totally "stable" atm. so I see your concerns

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
mustafa-ogun-ozturk: I think it does not change much. This call should not be here at all but the current implementation is full of such calls.

deniss-mc: would make the chatview bit shorter at least - this is sort of file handling methods anyway

widgetssdk/src/main/java/com/glia/widgets/chat/ChatController.java
mustafa-ogun-ozturk: I think if the message is here the count can be here too, logic is not so different. The controller is a huge mess anyway and this is a very minor thing to improve. If we want to move the count we need to move the logic for the message too which I do not want to atm. Let's leave it as it is since it works otherwise we will be dealing some other issues

widgetssdk/src/main/java/com/glia/widgets/chat/ChatController.java
mustafa-ogun-ozturk: I think the same as above

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
mustafa-ogun-ozturk: Alright, will move it then.

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
mustafa-ogun-ozturk: Totally, so agreed on keeping it as it is?

deniss-mc (approved):

PR #53, opened 2021-08-04
MUIC-406: [Android] User sees blank view instead of picture from his …

deniss-mc:
…camera

closes:
MOB-801: [Android] Upgrading from chat to video call in visitor side is shown blank withe screen

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
@@ -550,9 +550,7 @@ public class CallController implements
     public void onNewVisitorMediaState(VisitorMediaState visitorMediaState) {
         Logger.d(TAG, "newVisitorMediaState: " + visitorMediaState.toString());
         emitViewState(callState.visitorMediaStateChanged(visitorMediaState));
-        if (callState.is2WayVideoCall()) {
-            Logger.d(TAG, "newVisitorMediaState: video");
-        }
+        Logger.d(TAG, "newVisitorMediaState- is2WayVideo:" + callState.is2WayVideoCall());
     }
 
     @Override
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
@@ -330,13 +330,6 @@ public class CallController implements
     public void setViewCallback(CallViewCallback callViewCallback) {
         Logger.d(TAG, "setViewCallback");
         this.viewCallback = callViewCallback;
-        viewCallback.emitState(callState);
-
-        if (callState.isVideoCall()) {
-            startOperatorVideo(callState.callStatus.getOperatorMediaState());
-            if (callState.is2WayVideoCall())
-                startVisitorVideo(callState.callStatus.getVisitorMediaState());
-        }
         emitViewState(callState.landscapeControlsVisibleChanged(!callState.isVideoCall()));
     }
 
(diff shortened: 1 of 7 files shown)


widgetssdk/src/main/java/com/glia/widgets/call/CallView.java
```diff
@@ -444,6 +431,40 @@ public void startVisitorVideoView(MediaState visitorMediaState) {
                 .getScreenSharingController(screenSharingCallback);
     }
 
+    private void shouldShowVisitorVideo(CallState state) {
+        if (
```
gersonnoboa, line 435: Could this, and the operator video if, be changed in a way that for example the `Media.Status.PLAYING` is contained into a variable? So that it reads "if state is 2 way video call and isVisitorVideoStatusPlaying and visibility is gone, do this". cause right now that PLAYING part is not the easiest to read.

widgetssdk/src/main/java/com/glia/widgets/call/CallView.java
```diff
@@ -444,6 +431,40 @@ public void startVisitorVideoView(MediaState visitorMediaState) {
                 .getScreenSharingController(screenSharingCallback);
     }
 
+    private void shouldShowVisitorVideo(CallState state) {
```
gersonnoboa, line 434: `should` tells to me that this should return a boolean. Maybe there is another way to name these two methods.

widgetssdk/src/main/java/com/glia/widgets/call/CallView.java
```diff
@@ -444,6 +431,40 @@ public void startVisitorVideoView(MediaState visitorMediaState) {
...
+                        state.callStatus
+                                .getVisitorMediaState()
+                                .getVideo()
+                                .getStatus() == Media.Status.PLAYING &&
+                        visitorVideoContainer.getVisibility() == GONE
+        ) {
+            visitorVideoContainer.setVisibility(VISIBLE);
+            showVisitorVideo(state.callStatus.getVisitorMediaState());
+        }
+        if (!state.is2WayVideoCall() && visitorVideoContainer.getVisibility() == VISIBLE) {
```
gersonnoboa, line 446: This if could be an else if. If I understand correctly, either only the first or only the second can happen. Same in `shouldShowOperatorVideo`

gersonnoboa (commented):
I ran it on device and it is working fine. However you can see (hopefully) that there are some things after the rounded corners, it's as if the background of the container is not fully transparent. Could that also be fixed here?

imagen

dukhovnyi (approved):
Looks good for me, just a few minor suggestions.

deniss-mc:
> I ran it on device and it is working fine. However you can see (hopefully) that there are some things after the rounded corners, it's as if the background of the container is not fully transparent. Could that also be fixed here?
> 
> imagen

https://salemove.atlassian.net/browse/MUIC-433 created issue to handle that

widgetssdk/src/main/java/com/glia/widgets/call/CallView.java
deniss-mc: extracted to state

widgetssdk/src/main/java/com/glia/widgets/call/CallView.java
deniss-mc: renamed to handleOperatorVideoState

widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
```diff
@@ -573,7 +552,6 @@ public void onNewVisitorMediaState(VisitorMediaState visitorMediaState) {
         emitViewState(callState.visitorMediaStateChanged(visitorMediaState));
         if (callState.is2WayVideoCall()) {
             Logger.d(TAG, "newVisitorMediaState: video");
```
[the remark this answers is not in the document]
deniss-mc: fixed

widgetssdk/src/main/java/com/glia/widgets/call/CallView.java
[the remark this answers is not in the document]
deniss-mc: fixed

PR #93, opened 2021-10-01
MUIC-506: Typing indicator space is always visible

mustafa-ogun-ozturk:


widgetssdk/src/main/res/layout/chat_view.xml
```diff
@@ -15,7 +15,7 @@
         android:id="@+id/chat_recycler_view"
         android:layout_width="match_parent"
         android:layout_height="0dp"
-        app:layout_constraintBottom_toTopOf="@id/divider_view"
+        app:layout_constraintBottom_toTopOf="@id/operator_typing_animation_view"
```
deniss-mc, line 18: when view is gone does it behave ok?

deniss-mc (changes requested):

widgetssdk/src/main/res/layout/chat_view.xml
mustafa-ogun-ozturk: Depends on what is okay. I did not see anything weird.

PR #104, opened 2021-10-14
Implement Gradle script for uploading Android Widgets to Maven Central

dukhovnyi:
MOB-873

### DIFF
--- a/scripts/publish-module.gradle
+++ b/scripts/publish-module.gradle
@@ -1,0 +1,58 @@
+apply plugin: 'maven-publish'
+apply plugin: 'signing'
+
+group = PUBLISH_GROUP_ID
+version = PUBLISH_VERSION
+
+afterEvaluate {
+    publishing {
+        publications {
+            release(MavenPublication) {
+
+                groupId PUBLISH_GROUP_ID
+                artifactId PUBLISH_ARTIFACT_ID
+                version PUBLISH_VERSION
+
+                // Two artifacts, the `aar` (or `jar`) and the sources
+                if (project.plugins.findPlugin("com.android.library")) {
+                    from components.release
+                } else {
+                    from components.java
+                }
+
+                pom {
+                    name = PUBLISH_ARTIFACT_ID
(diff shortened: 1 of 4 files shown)


scripts/publish-module.gradle
```diff
@@ -0,0 +1,58 @@
...
+                if (project.plugins.findPlugin("com.android.library")) {
+                    from components.release
+                } else {
+                    from components.java
+                }
+
+                pom {
+                    name = PUBLISH_ARTIFACT_ID
+                    description = 'Glia Android Widgets SDK'
+                    url = 'https://www.glia.com'
```
gugalo, line 26: NIT: I think it would be nice to extract the description, URL and company name in the same style as `PUBLISH_GROUP_ID`, `PUBLISH_ARTIFACT_ID`.

scripts/publish-root.gradle
```diff
@@ -0,0 +1,34 @@
+// Create variables with empty default values
```
gugalo, line 1: I don't know if `publish-root` has some meaning to it but for the person who doesn't know it, it sounds very cryptic.
Maybe some more clear name could be used. For example: `publish-settings`, `publish-config`, `publish-properties`, `publish-variable`...

widgetssdk/build.gradle
```diff
@@ -73,3 +73,11 @@ dependencies {
 
     implementation "com.airbnb.android:lottie:$lottieVersion"
 }
+
+ext {
+    PUBLISH_GROUP_ID = 'com.glia'
+    PUBLISH_VERSION = android.defaultConfig.versionName
+    PUBLISH_ARTIFACT_ID = 'android-widgets'
+}
```
gugalo, line 81: NIT: Why not add these to `publish-root` gradle script?

scripts/publish-module.gradle
dukhovnyi: fixed

scripts/publish-root.gradle
dukhovnyi: Renamed to `publish-settings`

PR #117, opened 2021-11-03
Release 1.6.8

deniss-mc:


widgetssdk/build.gradle
```diff
@@ -7,8 +7,8 @@ android {
     defaultConfig {
         minSdkVersion 24
         targetSdkVersion 30
-        versionCode 12
-        versionName "1.6.7"
+        versionCode 13
+        versionName "1.7.8"
```
dukhovnyi, line 11: looks like a typo, should be **1.6.8** or did I miss something?

widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsConfig.java
```diff
@@ -14,18 +14,6 @@
     private final String region;
     private final int requestCode;
 
-    /**
-     * Deprecated since SDK version 1.6.5. Please use {@link GliaWidgetsConfig#GliaWidgetsConfig(String, String, Context, String, int)} instead.
-     */
-    @Deprecated
-    public GliaWidgetsConfig(String appToken, String apiToken, String siteId, Context context, String region, int requestCode) {
```
dukhovnyi, line 21: Do I understand correctly that all integrators will get a compilation error in case of removing this initializer? If so, let's avoid making breaking changes and let it as is with deprecation note.

dukhovnyi (changes requested):

widgetssdk/build.gradle
deniss-mc: attributes count will change - it would be better to update to 1.7

Id replace 8 there to be same as versionCode, what do you think?

widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsConfig.java
deniss-mc: yup, hence the 1.7 update

deniss-mc: - returned the constructor and setter with the deprecation notice

widgetssdk/build.gradle
deniss-mc: version code set to 1.6.8

dukhovnyi (approved):
Looks good 👍

PR #121, opened 2021-11-12
App crashes when try to turn off the video during a video call

VadymMobileDev:
MUIC-556

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
@@ -496,7 +496,7 @@ public class CallController implements
         Logger.d(TAG, "videoButtonClicked");
         if (callState.isCallOngoingAndOperatorConnected()) {
             VisitorMediaState currentMediaState = callState.callStatus.getVisitorMediaState();
-            Logger.d(TAG, "videoButton status:" + currentMediaState.getAudio().getStatus().toString());
+            Logger.d(TAG, "videoButton status:" + currentMediaState.getVideo().getStatus().toString());
             if (currentMediaState.getVideo().getStatus() == Media.Status.PAUSED) {
                 currentMediaState.getVideo().resume();
                 if (currentMediaState.getVideo().getStatus() == Media.Status.PLAYING) {


dukhovnyi (approved):

gugalo (approved):

gugalo:
Should it also somewhere check the nullability of the `getVideo`?

widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
```diff
@@ -496,7 +496,7 @@ public void videoButtonClicked() {
         Logger.d(TAG, "videoButtonClicked");
         if (callState.isCallOngoingAndOperatorConnected()) {
             VisitorMediaState currentMediaState = callState.callStatus.getVisitorMediaState();
-            Logger.d(TAG, "videoButton status:" + currentMediaState.getAudio().getStatus().toString());
+            Logger.d(TAG, "videoButton status:" + currentMediaState.getVideo().getStatus().toString());
```
jakob-mc, line 499: Is it necessary to keep the log code in the source? Ideally you would not have logging in production code.

dukhovnyi: @jakob-mc  it makes sense, but this logging is under `debug` level, so it will be ignored for production code.
I could be wrong, so if I missed anything, please clarify once more your suggestion

jakob-mc: I was thinking if anybody benefits from the logs after the feature/fix have been implemented and tested? If they are only helpful during the task development, then should be thrown out before merging the finished task.

deniss-mc:
LGTM - but please investigate why mediastate.getAudio() was null

widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
dukhovnyi: imho, logging is useful for debugging purposes anytime. It should not go to the production with info/debug/trace levels, but I'm not sure that we need to remove logging at all.

deniss-mc (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
deniss-mc: I'd prefer to go with logging as it has saved us a lot of time previously :)

jakob-mc: clear, just wanted to make sure you really benefit from them.

jakob-mc (approved):

PR #124, opened 2021-11-15
MUIC-565: [Android] Disable overlay permission dialog (if bubble disabled)

deniss-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-565

**Additional info**
Will use `GliaConfigurationManager` for Configuration holder (will need to move whitelabel stuff there too)
At some point would love to introduce some sort of public interface with all the non-ui related configurations (or maybe add some to `GliaConfig` class - some will come from the core-sdk "purchased package" anyway in the future.

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/MainFragment.java
+++ b/app/src/main/java/com/glia/exampleapp/MainFragment.java
@@ -80,6 +80,6 @@ public class MainFragment extends Fragment {
         UiTheme uiTheme = Utils.getUiThemeByPrefs(sharedPreferences, getResources());
         intent.putExtra(GliaWidgets.UI_THEME, uiTheme);
         // use to set the bubble functionality
-        // intent.putExtra(GliaWidgets.USE_OVERLAY, false);
+        intent.putExtra(GliaWidgets.USE_OVERLAY, Utils.getUseOverlay(sharedPreferences, getResources()));
     }
 }
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
@@ -68,7 +68,7 @@ public class CallActivity extends AppCompatActivity {
         queueId = intent.getStringExtra(GliaWidgets.QUEUE_ID);
         runtimeTheme = intent.getParcelableExtra(GliaWidgets.UI_THEME);
         contextUrl = intent.getStringExtra(GliaWidgets.CONTEXT_URL);
-        useOverlays = intent.getBooleanExtra(GliaWidgets.USE_OVERLAY, true);
+        useOverlays = intent.getBooleanExtra(GliaWidgets.USE_OVERLAY, Dependencies.getSdkConfigurationManager().isUseOverlay());
         callView.setTheme(runtimeTheme);
         callView.setOnBackClickedListener(onBackClickedListener);
         callView.setOnEndListener(onEndListener);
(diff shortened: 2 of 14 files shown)


dukhovnyi:
could we introduce this kind of setting in our TestingApp also to have possibility to test it by QAs?

dukhovnyi (commented):
could we introduce this kind of setting in our TestingApp also to have possibility to test it by QAs?

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
```diff
@@ -68,7 +68,7 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
         queueId = intent.getStringExtra(GliaWidgets.QUEUE_ID);
         runtimeTheme = intent.getParcelableExtra(GliaWidgets.UI_THEME);
         contextUrl = intent.getStringExtra(GliaWidgets.CONTEXT_URL);
-        useOverlays = intent.getBooleanExtra(GliaWidgets.USE_OVERLAY, true);
+        useOverlays = intent.getBooleanExtra(GliaWidgets.USE_OVERLAY, Dependencies.getSdkConfigurationManager().isUseOverlay());
```
jakob-mc, line 71: I suggest to use a separate method to get default value, improves readability.

widgetssdk/src/main/java/com/glia/widgets/call/CallView.java
```diff
@@ -201,6 +201,7 @@ public void startCall(String companyName,
                           String contextUrl,
                           boolean useOverlays,
                           Engagement.MediaType mediaType) {
+        Dependencies.getSdkConfigurationManager().setUseOverlay(useOverlays);
```
jakob-mc, line 204: `startCall` method is doing many different things here: setting configuration, initCall, onNavigatedToCall. Method could be broken down into multiple methods with less responsibilities.

jakob-mc: Is there any more appropriate layer (presenter, controller etc) to put that kind of code?

jakob-mc (changes requested):

deniss-mc:
> could we introduce this kind of setting in our TestingApp also to have possibility to test it by QAs?

Added `Use Overlay` option to settings (enabled by default)- 
To test it:
1) Wipe data
2) Disable in settings
3) Go to ChatView

Expected:
Overlay Permissions dialog is not displayed

app/src/main/java/com/glia/exampleapp/MainFragment.java
```diff
@@ -80,6 +80,6 @@ private void setNavigationIntentData(Intent intent, SharedPreferences sharedPref
         UiTheme uiTheme = Utils.getUiThemeByPrefs(sharedPreferences, getResources());
         intent.putExtra(GliaWidgets.UI_THEME, uiTheme);
         // use to set the bubble functionality
-        // intent.putExtra(GliaWidgets.USE_OVERLAY, false);
```
jakob-mc, line 83: cleanup

dukhovnyi (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
deniss-mc: Introduced Configuration class - moved stuff around a bit

jakob-mc (changes requested):

jakob-mc (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallView.java
deniss-mc: Prefer not to do it atm. 
- breaking down is not related to this commit (will add some sort of technical debt task for that)
- its extra usecase for setting one variable

Atm. Views public interface requires variables

Will refac to using Configuration class ASAP (check PR notes)

PR #136, opened 2021-11-29
Update Visitor Information

VadymMobileDev:
MOB-1019

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
@@ -5,9 +5,15 @@ import android.content.Intent;
 
 import com.glia.androidsdk.Glia;
 import com.glia.androidsdk.GliaConfig;
+import com.glia.androidsdk.visitor.VisitorInfoUpdateRequest;
+import com.glia.widgets.core.visitor.GliaWidgetException;
+import com.glia.widgets.core.visitor.GliaVisitorInfo;
+import com.glia.widgets.core.visitor.VisitorInfoUpdate;
 import com.glia.widgets.di.Dependencies;
-import com.glia.widgets.view.head.ChatHeadsController;
 import com.glia.widgets.helper.Logger;
+import com.glia.widgets.view.head.ChatHeadsController;
+
+import java.util.function.Consumer;
 
 /**
  * This class is a starting point for integration with Glia Widgets SDK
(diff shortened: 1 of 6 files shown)


widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -137,4 +141,23 @@ public static void clearVisitorSession() {
         Dependencies.getControllerFactory().destroyControllers();
         Glia.clearVisitorSession();
     }
+
+    private static final Consumer<GliaException> updateVisitorExceptionConsumer = exception -> {
+        if (exception != null) {
+            Logger.e(TAG, "Visitor Info update failed: " + exception.toString());
+        } else {
+            Logger.e(TAG, "Visitor Info updated success: " + exception.toString());
```
jakob-mc, line 149: should the log be error-level in case of success?

VadymMobileDev: Yes you are right, I'll delete it

jakob-mc: can't see any changes?

VadymMobileDev: Do you mean this, or I something misunderstood?
`        if (exception != null) {
            Logger.e(TAG, "Visitor Info update failed: " + exception.toString());
        } else {
            Logger.e(TAG, "Visitor Info updated Success");
        }`

gugalo: Wanted to confirm, is there some reason we are not notifying the integrator of the request result? I think additionally to this we can also let the integrator register his own callback if he wants to know if the request was made successfully?

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -137,4 +141,23 @@ public static void clearVisitorSession() {
         Dependencies.getControllerFactory().destroyControllers();
         Glia.clearVisitorSession();
     }
+
+    private static final Consumer<GliaException> updateVisitorExceptionConsumer = exception -> {
+        if (exception != null) {
+            Logger.e(TAG, "Visitor Info update failed: " + exception.toString());
+        } else {
+            Logger.e(TAG, "Visitor Info updated Success");
+        }
```
gugalo, line 150: Wanted to confirm, is there some reason we are not notifying the integrator of the request result? I think additionally to this we can also let the integrator register his own callback if he wants to know if the request was made successfully?

VadymMobileDev: as I understand it was necessary to add only a possibility without implementation.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
jakob-mc: this line yes. what I ment was that should error be logged in case of a success. Debug level could be more proper `Logger.d`

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -137,4 +141,23 @@ public static void clearVisitorSession() {
         Dependencies.getControllerFactory().destroyControllers();
         Glia.clearVisitorSession();
     }
+
+    private static final Consumer<GliaException> updateVisitorExceptionConsumer = exception -> {
+        if (exception != null) {
+            Logger.e(TAG, "Visitor Info update failed: " + exception.toString());
+        } else {
+            Logger.e(TAG, "Visitor Info updated Success");
```
dukhovnyi, line 149: The issue that @jakob-mc has mentioned is not resolved.
Please correct the logger level to debug.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -137,4 +141,23 @@ public static void clearVisitorSession() {
...
+        }
+    };
+
+    /**
+     * Updates the visitor's information
+     *
+     * Updates the visitor's information stored on the server. This information will also be displayed to the operator.
+     * */
+    public static void updateVisitorInfo(VisitorInfoUpdateRequest visitorInfoUpdateRequest) {
+        Logger.d(TAG, "updateVisitorInfo");
```
dukhovnyi, line 159: This log message looks not so descriptive, at least for me.
Perhaps we could improve it in sense of adding some additional not sensitive information or remove it if this log doesn't provide any useful debug information.
Does it make sense?

dukhovnyi (changes requested):

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
gugalo: You mean to add the callback to the `VisitorInfoUpdateRequest.Builder` ?

I think that there should be some way to notify the integrator of the request was successful or not. And having a callback in the update method is the simplest solution. Though having it in the builder is also good.
@dukhovnyi

widgetssdk/src/main/java/com/glia/widgets/core/visitor/GliaVisitorException.java
```diff
@@ -0,0 +1,26 @@
+package com.glia.widgets.core.visitor;
+
+import com.glia.androidsdk.GliaException;
+
+public class GliaVisitorException extends RuntimeException{
```
gugalo, line 5: This class is almost a duplicate of the `GliaException`. Also, I think it would be better to use it instead of this one there is no need in my opinion to have a dedicated Exception. Or if you want to wrap it to have a separate exception for Widget SDK then it might have a more generic name. For example `GliaWidgetException`

AndriiHorishniiMOC:
Why don't you use the reformat code to follow the code style?

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/core/visitor/VisitorInfoUpdate.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    public String getPhone() {
+        return phone;
+    }
+
+    @Override
+    public String getNote() {
+        return null;
```
dukhovnyi, line 61: why do we return here `null` instead of `note`?

gugalo (approved):
The `null` thing that Yurii mentioned. Nothing else from me

dukhovnyi (approved):

PR #137, opened 2021-11-29
MUIC-601: [Android] Vertical CTA button position

deniss-mc:
# https://glia.atlassian.net/browse/MUIC-601

Note: 
Rough implementation of Vertical Alignment support

Can test with enabling toggle in settings screen

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/view/Dialogs.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/view/Dialogs.java
@@ -6,7 +6,6 @@ import android.content.res.ColorStateList;
 import android.graphics.Typeface;
 import android.view.LayoutInflater;
 import android.view.View;
-import android.widget.Button;
 import android.widget.ImageButton;
 import android.widget.ImageView;
 import android.widget.TextView;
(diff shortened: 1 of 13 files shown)


deniss-mc:
rebased onto develop

widgetssdk/src/main/java/com/glia/widgets/view/Dialogs.java
```diff
@@ -33,9 +33,17 @@ public static AlertDialog showOptionsDialog(Context context,
                                                 View.OnClickListener positiveButtonClickListener,
                                                 View.OnClickListener negativeButtonClickListener,
                                                 DialogInterface.OnCancelListener cancelListener) {
+
+        boolean isUseAlertDialogButtonVerticalAlignment = Utils.getGliaAlertDialogButtonUseVerticalAlignment(theme);
```
jakob-mc, line 37: this layout code is duplicated 3 times in the class. can you separate it into a method like `getLayout()`?

jakob-mc (approved):

widgetssdk/src/main/java/com/glia/widgets/view/Dialogs.java
deniss-mc: different layouts

it will be getLayout(boolean, resid, resid) not sure if worth it

PR #159, opened 2021-12-28
Use FileUploadDenied error instead of InternalError

dukhovnyi:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-639

**Additional info:**
Show proper error message when file uploading for visitor is disabled from Glia Hub side

**Screenshots:**


deniss-mc (approved):
Looks good - might need to hide the attachment button if it is disabled in first case?

dukhovnyi:
> Looks good - might need to hide the attachment button if it is disabled in first case?

We will introduce the feature to hide the attachment button based on Site configuration soon.

PR #164, opened 2021-12-30
Fix incorrect "Audio call" title during connection with operator

AndriiHorishniiMOC:
Fixed bug where when an operator upgrades to two-way video, a visitor sees an Audio call screen title

Used the requested media type to set the title on the call screen.

[MOB-1037](https://glia.atlassian.net/browse/MOB-1037)

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
@@ -631,7 +631,7 @@ public class ChatController implements
                 Logger.d(TAG, "upgradeOfferChoiceSubmitSuccess");
                 if (submitter == MediaUpgradeOfferRepository.Submitter.CHAT) {
                     String requestedMediaType;
-                    if (offer.video != null) {
+                    if (offer.video != null && offer.video != MediaDirection.NONE) {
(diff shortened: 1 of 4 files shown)


dukhovnyi (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
```diff
@@ -631,7 +631,7 @@ public void upgradeOfferChoiceSubmitSuccess(
                 Logger.d(TAG, "upgradeOfferChoiceSubmitSuccess");
                 if (submitter == MediaUpgradeOfferRepository.Submitter.CHAT) {
                     String requestedMediaType;
-                    if (offer.video != null) {
+                    if (offer.video != null && offer.video != MediaDirection.NONE) {
```
deniss-mc, line 634: what is this `MediaDirection.NONE` - aka when video offer can end up in this state?

deniss-mc (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
AndriiHorishniiMOC: When we get an offer to update to the audio.

PR #175, opened 2022-01-14
MUIC-634 Fix choice card ui issues

jakob-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-634

**Additional info:**
- Improved the ui to match the design and iOS
- Choice card image height is fixed to 200px just like on the iOS. This solves the issue with card height jumping after image loading finished.
- Broke down a larger method to improve the readability.

**Screenshots:**

Before:
Screenshot 2022-01-14 at 17 38 40

After:
Screenshot 2022-01-14 at 17 37 48

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
@@ -9,7 +9,6 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.pm.PackageManager;
-import android.content.res.ColorStateList;
 import android.content.res.Resources;
 import android.content.res.TypedArray;
 import android.database.Cursor;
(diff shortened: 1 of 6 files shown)


deniss-mc (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
```diff
@@ -289,7 +281,8 @@ public void startChat(
                                     contextUrl,
                                     this.theme
                             ),
-                            activity instanceof ChatActivity, activity instanceof ChatActivity && useOverlays
```
deniss-mc, line 292: Revert this, I'll remove it in ChatHead solution

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
jakob-mc: it's the second time I run into these warnings and I don't see the reason to keep it. If you need that code for reference, you can always use git history.

PR #179, opened 2022-01-20
Version update

deniss-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-649


dukhovnyi (commented):
The version in commit message is incorrect.

PR #198, opened 2022-02-28
MUIC-494 - Chat - Keyboard may freeze when visitor is typing fast in the message text box

jakob-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-494

**Additional info:**
- remove unnecessary callback that caused infinite loop when typing fast
- minor code cleanup


deniss-mc (approved):
Seems ok

DavDo (approved):

PR #204, opened 2022-03-02
Remove keys from repository

deniss-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-666

### DIFF
--- a/app/src/master/res/values/setup.xml
+++ b/app/src/master/res/values/setup.xml
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools">
-    <string name="queue_id" tools:ignore="ExtraTranslation">55e5ca69-04a9-453a-88c1-73b13e83a227</string>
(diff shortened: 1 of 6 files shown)


app/src/master/res/values/setup.xml
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools">
-    <string name="queue_id" tools:ignore="ExtraTranslation">55e5ca69-04a9-453a-88c1-73b13e83a227</string>
+    <string name="default_queue_id" tools:ignore="ExtraTranslation">55e5ca69-04a9-453a-88c1-73b13e83a227</string>
```
gugalo, line 3: I don't think there is a point in leaving the default queue ID. This is open source project and it might be confusing why there is a default queue ID that is site-specific and won't work with just any site.

gugalo (commented):
As I understand this will require an update of the CI/CD as well?

app/src/master/res/values/setup.xml
deniss-mc: https://glia.atlassian.net/browse/MUIC-667

gugalo (approved):

PR #208, opened 2022-03-04
MUIC-647 Disable attachment download when already started

jakob-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-647


deniss-mc (approved):
Seems OK

PR #217, opened 2022-03-15
Click handles diff with Math.abs, action solution removed

deniss-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-664

**Additional info:**
Previous solution worked well on emulator but not on real devices sadly :)

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/view/ViewHelpers.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/view/ViewHelpers.java
@@ -13,7 +13,6 @@ public class ViewHelpers {
         private int initialY;
         private float initialTouchX;
         private float initialTouchY;
-        private int lastAction;
         private final OnRequestInitialCoordinates onRequestInitialCoordinates;
         private final OnMoveListener onMoveListener;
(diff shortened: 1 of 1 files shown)


widgetssdk/src/main/java/com/glia/widgets/view/ViewHelpers.java
```diff
@@ -40,22 +39,19 @@ public boolean onTouch(View v, MotionEvent event) {
...
                     initialTouchY = event.getRawY();
-
-                    lastAction = event.getAction();
                     return true;
                 case MotionEvent.ACTION_UP:
-                    if (lastAction == MotionEvent.ACTION_DOWN) {
-                        v.performClick();
+                    int xDiff = (int) Math.abs(event.getRawX() - initialTouchX);
+                    int yDiff = (int) Math.abs(event.getRawY() - initialTouchY);
+                    if (xDiff < 20 && yDiff < 20) {
```
jakob-mc, line 46: pls use constant

jakob-mc (changes requested):

jakob-mc (approved):

PR #219, opened 2022-03-16
MUIC-682 Fix screensharing notification crash

jakob-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-682

**Additional info:**

Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.
    Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles.


deniss-mc (approved):
Seems ok

DavDo (approved):

PR #221, opened 2022-03-17
Part 1. Add screen sharing mode to the settings, pass to activities.

jakob-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-562

**Additional info:**

- please review commit by commit so you can separately review code clean up and new feature code

**Screenshots:**
Screenshot 2022-03-17 at 11 32 10

Screenshot 2022-03-17 at 11 32 17

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/MainFragment.java
+++ b/app/src/main/java/com/glia/exampleapp/MainFragment.java
@@ -33,12 +33,8 @@ public class MainFragment extends Fragment {
     public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceState) {
         super.onViewCreated(view, savedInstanceState);
         NavController navController = NavHostFragment.findNavController(this);
-        view.findViewById(R.id.settings_button).setOnClickListener(view1 -> navController.navigate(R.id.settings));
-        /* todo https://salemove.atlassian.net/browse/MUIC-362
-        view.findViewById(R.id.integrator_chat).setOnClickListener(view1 -> {
-            navController.navigate(R.id.chat);
-        });
-         */
+        view.findViewById(R.id.settings_button).setOnClickListener(view1 ->
+                navController.navigate(R.id.settings));
         view.findViewById(R.id.chat_activity_button).setOnClickListener(v ->
                 navigateToChat());
         view.findViewById(R.id.audio_call_button).setOnClickListener(v ->
--- a/app/src/main/java/com/glia/exampleapp/MainFragment.java
+++ b/app/src/main/java/com/glia/exampleapp/MainFragment.java
@@ -15,7 +15,6 @@ import androidx.navigation.fragment.NavHostFragment;
 import androidx.preference.PreferenceManager;
 
 import com.glia.widgets.GliaWidgets;
-import com.glia.widgets.UiTheme;
 import com.glia.widgets.call.CallActivity;
 import com.glia.widgets.chat.ChatActivity;
 import com.glia.widgets.view.head.ChatHeadLayout;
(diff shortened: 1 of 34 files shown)


app/src/main/java/com/glia/exampleapp/MainFragment.java
```diff
@@ -39,42 +33,34 @@ public View onCreateView(@NonNull LayoutInflater inflater,
     public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceState) {
         super.onViewCreated(view, savedInstanceState);
         NavController navController = NavHostFragment.findNavController(this);
-        view.findViewById(R.id.settings_button).setOnClickListener(view1 -> {
-            navController.navigate(R.id.settings);
-        });
-        SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(requireContext());
-        /* https://salemove.atlassian.net/browse/MUIC-362
+        view.findViewById(R.id.settings_button).setOnClickListener(view1 -> navController.navigate(R.id.settings));
+        /* todo https://salemove.atlassian.net/browse/MUIC-362
```
deniss-mc, line 37: Not todo - task that removed it

app/src/main/java/com/glia/exampleapp/Utils.java
```diff
@@ -230,4 +231,24 @@ public static Integer getTypefaceFromPrefs(SharedPreferences sharedPreferences,
...
+        );
+        String unbounded = resources.getString(R.string.screen_sharing_mode_unbounded);
+        String appBounded = resources.getString(R.string.screen_sharing_mode_app_bounded);
+
+        if (value.equals(unbounded)) {
+            return ScreenSharing.Mode.UNBOUNDED;
+        } else if (value.equals(appBounded)) {
+            return ScreenSharing.Mode.APP_BOUNDED;
+        } else {
+            return null;
```
deniss-mc, line 251: else here seems unnecessary - return default

deniss-mc: why not use toggle instead?

deniss-mc (changes requested):

app/src/main/java/com/glia/exampleapp/MainFragment.java
jakob-mc: is this comment expected to be there forever or is it assumed to be removed one day?

jakob-mc: moreover, should we have refs to Jira in open source code?

app/src/main/java/com/glia/exampleapp/Utils.java
jakob-mc: I assumed we need a default value as for some of the other settings. But if this is not needed, sure toggle is better, and the default value can be set somewhere deeper in the SDK.

app/src/main/java/com/glia/exampleapp/MainFragment.java
deniss-mc: Screenshot 2022-03-18 at 12 32 06
Well we are having them in pull requests which are visible aswell

Guess just removing the whole comment would be better approach here

dukhovnyi (changes requested):
I'd like to propose removing the ticket number from the commit title and moving it to the commit description. This change will make git history and changelog more consistent and more readable.

app/src/main/java/com/glia/exampleapp/Utils.java
```diff
@@ -211,7 +212,7 @@ public static Integer getTypefaceFromPrefs(SharedPreferences sharedPreferences,
         String delius = resources.getString(R.string.font_delius_value);
         String expletus = resources.getString(R.string.font_expletus_value);
         String tangerine = resources.getString(R.string.font_tangerine_value);
-        String defaultPrefValue = resources.getString(R.string.settings_value_default);
+        String defaultPrefValue = resources.getString(R.string.settings_value_default); // todo Wrong string resource is used. should be R.string.default_value
```
gugalo, line 215: NIT: I think it would be better if all TODO notes would have some related Jira tickets attached. Otherwise, these improvements/changes will not get the required attention in the future.

gugalo (approved):

PR #222, opened 2022-03-17
Fix screen sharing dialog not shown bug

jakob-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-457

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/ChatFragment.java
+++ b/app/src/main/java/com/glia/exampleapp/ChatFragment.java
@@ -61,7 +61,7 @@ public class ChatFragment extends Fragment {
 
     @Override
     public void onResume() {
-        chatView.onResume();
+        chatView.onResume(true);
         super.onResume();
     }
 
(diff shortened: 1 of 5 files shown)


deniss-mc (changes requested):
Returns bug when CallView is on top of ChatView - onResume is called

jakob-mc:
> Returns bug when CallView is on top of ChatView - onResume is called when orientation is changed

how does the bug manifest itself?

jakob-mc:
> Returns bug when CallView is on top of ChatView - onResume is called when orientation is changed

I reverted the isFinishing boolean logic only for ChatHead controller since other controllers seem to work well. I also made dialog controller lifecycle-aware. Now, if the view is paused, it does not try to display any dialogs.

dukhovnyi (changes requested):
I'd like to propose removing the ticket number from the commit title and moving it to the commit description. This change will make git history and changelog more consistent and more readable.

jakob-mc:
> I'd like to propose removing the ticket number from the commit title and moving it to the commit description. This change will make git history and changelog more consistent and more readable.

done

dukhovnyi:
looks like you have changed the PR name, but not commit message. or do I miss something?

andrews-moc (approved):

dukhovnyi (approved):

deniss-mc (approved):

PR #228, opened 2022-03-23
Devs want to show/hide attachment button depends on site configuration

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-1191?atlOrigin=eyJpIjoiOTZlOTE1YjFmMjAxNDhiOThiMDY0NmZhZTI2MDc2OGEiLCJwIjoiaiJ9

MOB-1191

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
@@ -257,6 +257,7 @@ public class ChatController implements
                 .setLastTypedText("")
                 .setChatInputMode(ChatInputMode.ENABLED_NO_ENGAGEMENT)
                 .setIsAttachmentButtonNeeded(false)
+                .setIsAttachmentAllowed(true)
                 .setIsChatInBottom(true)
                 .setMessagesNotSeen(0)
                 .setPendingNavigationType(null)
@@ -1278,11 +1279,12 @@ public class ChatController implements
     private void fileDownloadSuccess(AttachmentFile attachmentFile) {
         if (viewCallback != null) viewCallback.fileDownloadSuccess(attachmentFile);
     }
+
     private void updateAllowFileSendState() {
         siteInfoUseCase.execute((siteInfo, e) -> onSiteInfoReceived(siteInfo));
     }
 
     private void onSiteInfoReceived(@Nullable SiteInfo siteInfo) {
-        emitViewState(chatState.allowSendAttachmentStateChanged(siteInfo != null && siteInfo.getAllowedFileSenders().isVisitorAllowed()));
+        emitViewState(chatState.allowSendAttachmentStateChanged(siteInfo == null || siteInfo.getAllowedFileSenders().isVisitorAllowed()));
     }
 }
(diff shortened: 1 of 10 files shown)


gugalo (approved):

AndriiHorishniiMOC (approved):

deniss-mc (changes requested):
Change target branch to develop (rebase)

deniss-mc:
This will need to be combined with the Configuration approach at some point @jakob-mc 

At the moment Looks OK

@dukhovnyi  what should happen if information query will fail? does it need to retry? should it default to some parameters?

deniss-mc (approved):

dukhovnyi:
> This will need to be combined with the Configuration approach at some point CC: @jakob-mc
> 
> At the moment Looks OK
> 
> @dukhovnyi what should happen if information query will fail? does it need to retry? should it default to some parameters?
> 
> We would like to move configuration initialization out of Activity classes at some point - this way we will know if we can put callback to application level or still need to query those each time we enter the Glia Activities

@deniss-mc , in case of failed request we need to show the button and give the possibility to visitors to try to upload files. In case it is unavailable, visitors get the error message as it works now. Does it make sense?
cc: @DavDo

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
```diff
@@ -251,7 +256,7 @@ public ChatController(
                 .setChatItems(new ArrayList<>())
                 .setLastTypedText("")
                 .setChatInputMode(ChatInputMode.ENABLED_NO_ENGAGEMENT)
-                .setIsAttachmentButtonVisible(false)
+                .setIsAttachmentButtonNeeded(false)
```
deniss-mc, line 259: id still call it visible :)

deniss-mc:
> > This will need to be combined with the Configuration approach at some point CC: @jakob-mc
> > At the moment Looks OK
> > @dukhovnyi what should happen if information query will fail? does it need to retry? should it default to some parameters?
> > We would like to move configuration initialization out of Activity classes at some point - this way we will know if we can put callback to application level or still need to query those each time we enter the Glia Activities
> 
> @deniss-mc , in case of failed request we need to show the button and give the possibility to visitors to try to upload files. In case it is unavailable, visitors get the error message as it works now. Does it make sense? cc: @DavDo

I ment if the SiteInfo request from hub fails (there could be more configuration options like white label thingy, call disable etc. endless possibilities :) ). In case of attachments it should default to 'show' then - at the moment I see

`siteInfo == null` = don't show

DavDo:
> > This will need to be combined with the Configuration approach at some point CC: @jakob-mc
> > At the moment Looks OK
> > @dukhovnyi what should happen if information query will fail? does it need to retry? should it default to some parameters?
> > We would like to move configuration initialization out of Activity classes at some point - this way we will know if we can put callback to application level or still need to query those each time we enter the Glia Activities
> 
> @deniss-mc , in case of failed request we need to show the button and give the possibility to visitors to try to upload files. In case it is unavailable, visitors get the error message as it works now. Does it make sense? cc: @DavDo

@dukhovnyi In case of request with **App Token** we always receive null for SiteInfo that for I think we need to hide button only when `isVisitorAllowed() == false`

deniss-mc (approved):

PR #230, opened 2022-03-25
Some callback comming back on computation thread

deniss-mc:
Added post to handle view update on UI Thread
Also should fix blinking issue

**Jira issue:**
https://glia.atlassian.net/browse/MUIC-688


widgetssdk/src/main/java/com/glia/widgets/view/head/ChatHeadLayout.java
```diff
@@ -102,12 +102,12 @@ public boolean isInChatView() {
 
     @Override
     public void show() {
-        this.setVisibility(VISIBLE);
+        post(() -> setVisibility(VISIBLE));
```
jakob-mc, line 105: there seems to be a deeper problem with threading. current solution to patch the crashes by adding `post` to the different places inside the views is not very scalable.

jakob-mc (approved):

PR #248, opened 2022-04-07
Feature survey

AndriiHorishniiMOC:
Integrate Survey feature into Widgets SDK following design

https://glia.atlassian.net/browse/MOB-1168

Screenshot 2022-04-11 at 18 00 17

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/survey/SurveyActivity.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/survey/SurveyActivity.java
@@ -1,0 +1,46 @@
+package com.glia.widgets.survey;
+
+import android.os.Bundle;
+
+import androidx.appcompat.app.AppCompatActivity;
+
+import com.glia.androidsdk.engagement.Survey;
+import com.glia.widgets.GliaWidgets;
+import com.glia.widgets.R;
+import com.glia.widgets.UiTheme;
+import com.glia.widgets.di.Dependencies;
+
+public class SurveyActivity extends AppCompatActivity {
+    private static final String TAG = SurveyActivity.class.getSimpleName();
+
+    private SurveyView surveyView;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.survey_activity);
+        prepareSurveyView();
+    }
+
+    private void prepareSurveyView() {
+        surveyView = findViewById(R.id.survey_view);
+
+        SurveyContract.Controller surveyController =
+                Dependencies.getControllerFactory().getSurveyController();
+        surveyView.setController(surveyController);
+        Bundle extras = getIntent().getExtras();
+        UiTheme uiTheme = extras.getParcelable(GliaWidgets.UI_THEME);
+        surveyView.setTheme(uiTheme);
+
+        Survey survey = extras.getParcelable(GliaWidgets.SURVEY);
+        surveyController.init(survey);
+    }
+
+    @Override
+    protected void onDestroy() {
+        if (surveyView != null) {
+            surveyView.onDestroyView();
+        }
+        super.onDestroy();
+    }
+}
--- a/widgetssdk/src/main/java/com/glia/widgets/survey/SurveyAdapter.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/survey/SurveyAdapter.java
@@ -1,0 +1,496 @@
+package com.glia.widgets.survey;
+
+import static java.util.Arrays.asList;
+
+import android.content.Context;
+import android.content.res.ColorStateList;
+import android.graphics.Color;
+import android.graphics.Typeface;
+import android.graphics.drawable.Drawable;
+import android.graphics.drawable.GradientDrawable;
+import android.graphics.drawable.LayerDrawable;
+import android.text.Editable;
+import android.text.Html;
+import android.text.TextWatcher;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.Button;
+import android.widget.EditText;
+import android.widget.LinearLayout;
+import android.widget.RadioButton;
+import android.widget.RadioGroup;
+import android.widget.TextView;
+
+import androidx.annotation.LayoutRes;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.core.content.ContextCompat;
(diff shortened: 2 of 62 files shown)


widgetssdk/src/main/java/com/glia/widgets/survey/SurveyAdapter.java
```diff
@@ -25,20 +26,37 @@
...
+        void onAnswer(@NonNull Survey.Answer answer);
+    }
+
     private static final int SURVEY_SCALE = 1;
     private static final int SURVEY_YES_NO = 2;
     private static final int SURVEY_SINGLE_CHOICE = 3;
     private static final int SURVEY_OPEN_TEXT = 4;
 
-    private List<QuestionItem> items;
+    private final List<QuestionItem> list = new ArrayList<>();
```
jakob-mc, line 48: `questionItems` would describe better than `list` here

widgetssdk/src/main/java/com/glia/widgets/survey/SurveyAdapter.java
```diff
@@ -67,84 +85,16 @@ private static View createNewLayout(ViewGroup parent, @LayoutRes int layoutId) {
...
-        questionItem.setAnswer(Survey.Answer.makeAnswer(questionItem.getQuestion().getId(), response));
-    }
-
-    private void setAnswer(QuestionItem questionItem, boolean response) {
-        questionItem.setAnswer(Survey.Answer.makeAnswer(questionItem.getQuestion().getId(), response));
-    }
-
-    private void setAnswer(QuestionItem questionItem, String response) {
-        questionItem.setAnswer(Survey.Answer.makeAnswer(questionItem.getQuestion().getId(), response));
+        ((SurveyViewHolder) viewHolder).onBind(questionItem, question, answer, listener);
```
jakob-mc, line 94: seems no point to cache question and answer to a separate vars since the whole questionItem is already passed here.

widgetssdk/src/main/java/com/glia/widgets/survey/SurveyAdapter.java
```diff
@@ -158,37 +108,104 @@ public int getItemViewType(int position) {
...
+            titleTextView = itemView.findViewById(R.id.tv_title);
+            requiredError = itemView.findViewById(R.id.required_error);
         }
 
-        TextView titleTextView;
-        List<Button> buttons;
-        OnSurveyScaleClickListener listener;
+        void onBind(QuestionItem questionItem,
+                    Survey.Question question,
+                    Survey.Answer answer,
```
jakob-mc, line 130: answer is unused?

jakob-mc (changes requested):

widgetssdk/src/main/java/com/glia/widgets/survey/SurveyAdapter.java
```diff
@@ -433,7 +422,7 @@ void showRequiredError(boolean error) {
                 ColorStateList strokeColor =
                         error ? ContextCompat.getColorStateList(context, R.color.glia_system_negative_color) :
                                 ContextCompat.getColorStateList(context, R.color.glia_base_shade_color);
-                int width = (int) Math.round(context.getResources().getDimension(R.dimen.glia_px));
+                int width = Math.round(context.getResources().getDimension(R.dimen.glia_px));
```
jakob-mc, line 425: better way to fetch dimensions is using `context.getResources().getDimensionPixelSize(R.dimen.glia_px)`

jakob-mc (approved):
@deniss-mc pls review as well.

widgetssdk/src/main/res/layout/survey_view.xml
```diff
@@ -0,0 +1,75 @@
...
+            android:layout_width="match_parent">
+
+            <TextView
+                android:id="@+id/survey_title"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_marginTop="@dimen/glia_large_x_large"
+                android:text="@string/glia_survey_screen_title"
+                android:gravity="center"
+                android:textSize="20sp"
```
deniss-mc, line 26: text appearance? but i guess customization comes later so let it be

deniss-mc (approved):

widgetssdk/src/main/java/com/glia/widgets/survey/SurveyActivity.java
```diff
@@ -0,0 +1,42 @@
...
+        setContentView(R.layout.survey_activity);
+
+        surveyView = findViewById(R.id.survey_view);
+
+        SurveyContract.Controller surveyController =
+                Dependencies.getControllerFactory().getSurveyController();
+        surveyView.setController(surveyController);
+
+        Bundle extras = getIntent().getExtras();
+        Survey survey = extras.getParcelable(GliaWidgets.SURVEY);
```
deniss-mc, line 30: Survey can be non-parcelable

You could pass "survey id" (or identification for pending survey if even needed) with intent to survey activity
and query pending survey on controller init.

So this way Controller would get it from Repository, not the Activity

deniss-mc: Suggestion from https://github.com/salemove/android-sdk-widgets/pull/259#discussion_r855957103

AndriiHorishniiMOC: A survey can be got from the engagement object. Is it possible to get already ended engagement?

AndriiHorishniiMOC: Also, the engagement can not contain a survey. What in this case can we do? Open the survey activity and then finish it?

deniss-mc: `GliaSurveyRepository.class` repository receives the pending survey object.
You can hold it there instead of cascading down to Chat or Call activity.

example:
SurveyRepository - ResultStorage object is storing the pending survey already
ChatController - `viewCallback.navigateToSurvey(survey);` at line 1180 - remove the "survey" part from callback
`viewCallback.navigateToSurvey();`  <- after that You will see what needs to be removed.

SurveyActivity - remove parcelable survey part
SurveyController remove survey part from init
SurveyController on init could ask SurveyRepository for pending survey object

If survey was null then according to null checks in Chat and Call controller it shouldn't even get to SurveyActivity and would show endengagement dialog instead.

This way You will not need to open/close it :)

PR #261, opened 2022-04-14
Fix crash caused by build time theme being null in chat head controller.

jakob-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-715

**Additional info:**
I added a null check to avoid the crash. There might be a deeper problem but I'm hesitant to change the logic and break something.


DavDo (approved):
I hope this is a temporary solution and you will take care to find the cause of this problem.

jakob-mc:
> I hope this is a temporary solution and you will take care to find the cause of this problem.

yes, I need to consult with Deniss, he's the mastermind behind that logic.

deniss-mc (approved):
This will require some focus later, looks good atm

PR #262, opened 2022-04-14
Fix chat head visibility issue on screen rotation

jakob-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-710

**Additional info:**
     We need to keep track of the currently active (topmost) view. This can be either ChatView
     or CallView. CallView has translucent theme and this changes the usual lifecycle of an activity.
     When the translucent CallActivity is on top of the ChatActivity and config change happens (e.g screen rotation)
     then ChatActivity onResume and onPause are called right after CallActivity's onResume. Current
     solution ignores such onResume calls because another activity is already in resumed state.

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/ChatFragment.java
+++ b/app/src/main/java/com/glia/exampleapp/ChatFragment.java
@@ -61,7 +61,7 @@ public class ChatFragment extends Fragment {
 
     @Override
     public void onResume() {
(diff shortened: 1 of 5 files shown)


gugalo (approved):

deniss-mc (approved):
Seems good, lets hope translucent will cause us less issues from now on :)

PR #267, opened 2022-04-26
Make SDK compatible with compileSDKVersion 30

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-721

**Additional info:**
Some google libraries that we use in Widget SDK require clients to have compileSDKVersion to be 31+. And even though devs can override the library version it was agreed to downgrade those libraries to the latest version that would work with compileSDKVersion 30.

Changed the test app compile SDK version to 30 to make sure that no breaking changes would be introduced.

**Screenshots:**
No visual changes

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,12 +1,12 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 31
+    compileSdkVersion 30
     buildToolsVersion "30.0.3"
     defaultConfig {
         applicationId "com.glia.exampleapp"
         minSdkVersion 24
-        targetSdkVersion 31
+        targetSdkVersion 30
         versionCode 1
         versionName "1.0"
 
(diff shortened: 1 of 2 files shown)


gugalo:
I was not able to start a video call either from the visitor or the operator's side. However, the same issue is reproducible from the latest commit in the master branch.
https://drive.google.com/file/d/1JKzwwZGSREfDGXy3tE9OyHOpw-GQ1nlD/view?usp=sharing

And also have seen one more issue with media not being restored after app restart. But not sure if that is even expected (also reproducible from the master branch).
https://drive.google.com/file/d/1Rwv54JdqxUkx-5u97yFcjhYz8ssU4EiU/view?usp=sharing

If I need to create a bug for those items or this are already known?

dukhovnyi (approved):

andrews-moc (approved):

AndriiHorishniiMOC (approved):

jakob-mc (approved):

gugalo:
!merge

PR #272, opened 2022-05-05
Remove Call and Chat tasks from device's recents menu when engagement…

jakob-mc:
… ends.

**Jira issue:**
https://glia.atlassian.net/browse/MUIC-711

**Additional info:**

In case the engagement ends, Activity is removed from the device's Recents menu to avoid app users to accidentally start queueing for another call when they resume the app from the Recents menu and the app's backstack was empty. Backstack can be empty if the app was started from the bubble.

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallView.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallView.java
@@ -102,6 +102,7 @@ public class CallView extends ConstraintLayout {
 
     private OnBackClickedListener onBackClickedListener;
     private OnEndListener onEndListener;
+    private OnMinimizeListener onMinimizeListener;
     private OnNavigateToChatListener onNavigateToChatListener;
     private CallView.OnNavigateToSurveyListener onNavigateToSurveyListener;
 
@@ -171,9 +172,6 @@ public class CallView extends ConstraintLayout {
         minimizeButton.setOnClickListener(v -> {
             if (callController != null) {
                 callController.minimizeButtonClicked();
-            }
-            if (onEndListener != null) {
-                onEndListener.onEnd();
             }
         });
         muteButton.setOnClickListener(v -> {
@@ -380,6 +378,13 @@ public class CallView extends ConstraintLayout {
             public void destroyView() {
                 if (onEndListener != null) {
                     onEndListener.onEnd();
+                }
+            }
+
+            @Override
+            public void minimizeView() {
+                if (onMinimizeListener != null) {
+                    onMinimizeListener.onMinimize();
                 }
             }
         };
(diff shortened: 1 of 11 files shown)


widgetssdk/src/main/java/com/glia/widgets/call/CallView.java
```diff
@@ -172,9 +173,6 @@ private void setupViewActions() {
             if (callController != null) {
                 callController.minimizeButtonClicked();
             }
-            if (onEndListener != null) {
```
jakob-mc, line 175: Removed that because it's unnecessary as controller already does that.

andrews-moc (commented):
I tested the changes from MUIC-711 branch but the issue is still reproduced - visitor is queueing by opening the App from Android task manager.
What could be the reason for it?

jakob-mc:
> 

QA tested that and was working fine. Can you provide exact steps to reproduce?

deniss-mc (approved):

andrews-moc:
@jakob-mc Did you have a chance to check my video recording?The app shouldn't make a call at the end of the video, should it?

jakob-mc:
> @jakob-mc Did you have a chance to check my video recording? The app shouldn't make a call at the end of the video, should it?

Hey, I'll do it latest on Thursday (my Glia dev day). Is that ok?

andrews-moc:
> > @jakob-mc Did you have a chance to check my video recording? The app shouldn't make a call at the end of the video, should it?
> 
> Hey, I'll do it latest on Thursday (my Glia dev day). Is that ok?

Okay, no problem. Was just concerned if you haven't missed my comment. ;-)

jakob-mc:
> > > @jakob-mc Did you have a chance to check my video recording? The app shouldn't make a call at the end of the video, should it?
> > 
> > 
> > Hey, I'll do it latest on Thursday (my Glia dev day). Is that ok?
> 
> Okay, no problem. Was just concerned if you haven't missed my comment. ;-)
Activity was not removed from the Recents menu in case the survey was shown to the user. Pushed a fix.

andrews-moc:
@jakob-mc I'm not sure if it's an issue, let me describe the case which doesn't work.
Do all the steps like I've shown in the video until the survey is shown.
On the survey, instead of tapping the "Cancel", tap device "Back" button.
In this case, call view is still in the Recents and if user selects it, visitor starts calling again automatically, unfortunately.

jakob-mc:
> @jakob-mc I'm not sure if it's an issue, let me describe the case which doesn't work. Do all the steps like I've shown in the video until the survey is shown. On the survey, instead of tapping the "Cancel", tap device "Back" button. In this case, call view is still in the Recents and if user selects it, visitor starts calling again automatically, unfortunately.

I missed the back button click. It should behave the same way as pressing "Cancel". Pushed the fix. Thanks for pointing it out.

andrews-moc (approved):

PR #278, opened 2022-05-09
Fix on hold operatorstatusview

deniss-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-758

**Notes:**
There will be task to change Chat New Message indicator profile image content padding at some point - currently padding size: 0 (as in ios)


widgetssdk/src/main/java/com/glia/widgets/view/OperatorStatusView.java
```diff
@@ -151,6 +166,29 @@ public void setShowRippleAnimation(boolean show) {
         }
     }
 
+    private void updatePlaceholderView(
+            int size,
+            int contentPadding,
+            int visibility
+    ) {
+        placeholderView.post(() -> {
```
jakob-mc, line 174: why using post if all the invokers of this method manipulate ui without post?

jakob-mc (changes requested):

widgetssdk/src/main/java/com/glia/widgets/view/OperatorStatusView.java
```diff
@@ -63,14 +66,16 @@ public OperatorStatusView(@NonNull Context context, @Nullable AttributeSet attrs
             );
         }
 
-        int imageSize = (int) typedArray.getDimensionPixelSize(
+        IMAGE_SIZE_LARGE = getResources().getDimensionPixelSize(R.dimen.glia_ripple_animation_size);
+        PLACEHOLDER_ICON_PADDING_LARGE = IMAGE_SIZE_LARGE / 4;
```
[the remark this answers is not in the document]
andrews-moc: I would suggest doing the following:
- add separate dimens for image size and paadding large, default (to get rid of the magic number 4 and a not obvious relation between image size and glia_ripple_animation_size)
- remove constants as they don't behave like constants

widgetssdk/src/main/java/com/glia/widgets/view/OperatorStatusView.java
deniss-mc: this is related to .setConentPadding which is not currently working properly.

see:
https://github.com/material-components/material-components-android/issues/2063

it might get resolved in future releases.
I can extract other method calls from this post method and leave only content padding part for now with comment related to this issue

what do you think? @jakob-mc

jakob-mc: sounds good. + add a comment.

jakob-mc (approved):

andrews-moc (approved):

PR #280, opened 2022-05-09
Fix orientation change

AndriiHorishniiMOC:
Fix orientation change

[MUIC-763](https://glia.atlassian.net/browse/MUIC-763)

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/survey/SurveyController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/survey/SurveyController.java
@@ -2,6 +2,7 @@ package com.glia.widgets.survey;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
 
 import com.glia.androidsdk.engagement.Survey;
 import com.glia.widgets.core.survey.domain.GliaSurveyAnswerUseCase;
@@ -17,8 +18,12 @@ public class SurveyController implements SurveyContract.Controller {
     }
 
     private SurveyContract.View view;
-    private Survey survey;
-    private SurveyState state = new SurveyState.Builder().createSurveyState();
+
+    @VisibleForTesting
+    Survey survey;
+
+    @VisibleForTesting
+    SurveyState state = new SurveyState.Builder().createSurveyState();
 
     private final GliaSurveyAnswerUseCase gliaSurveyAnswerUseCase;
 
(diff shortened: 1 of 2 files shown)


widgetssdk/src/main/java/com/glia/widgets/survey/SurveyController.java
```diff
@@ -39,6 +44,22 @@ public void init(@Nullable Survey survey) {
         setQuestions(survey);
     }
 
+    private boolean isAlreadyInit(@Nullable Survey survey) {
+        if (this.survey == null) {
+            return false;
+        }
+        if (survey == null) {
+            return false;
+        }
```
andrews-moc, line 53: I would suggest to unite these two so that it looked like:
```
if (this.survey == null || survey == null) {
      return false;
}
```

widgetssdk/src/main/java/com/glia/widgets/survey/SurveyController.java
```diff
@@ -39,6 +44,22 @@ public void init(@Nullable Survey survey) {
...
+        if (!this.survey.getId().equals(survey.getId())) {
+            return false;
+        }
+        if (!this.survey.getEngagementId().equals(survey.getEngagementId())) {
+            return false;
+        }
+        return state.questions == null || !state.questions.isEmpty();
```
andrews-moc, line 60: Sorry, but I didn't got this condition. It seems to me these two parts are controversial:
If `state.questions` does not exist (null) => already initialized (I would expect that survey is not initialized in this case).
If `state.questions` not empty (there are some questions) there => already initialized (agree)

widgetssdk/src/main/java/com/glia/widgets/survey/SurveyController.java
```diff
@@ -111,6 +136,7 @@ public void onAnswer(@NonNull Survey.Answer answer) {
     public void onCancelClicked() {
         if (view != null) {
             view.finish();
+            controllerFactory.destroySurveyController();
```
deniss-mc, line 139: could reset the controller here instead of destroy in controller factory

widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
```diff
@@ -200,6 +200,11 @@ private void destroyChatController() {
         }
     }
 
+    public void destroySurveyController() {
```
deniss-mc, line 203: not needed

deniss-mc (changes requested):

widgetssdk/src/main/java/com/glia/widgets/survey/SurveyController.java
AndriiHorishniiMOC: Added tests for the `isAlreadyInit`

widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
AndriiHorishniiMOC: removed

widgetssdk/src/main/java/com/glia/widgets/survey/SurveyController.java
AndriiHorishniiMOC: Used `resetController()` instead of `destroySurveyController()`.

widgetssdk/src/main/java/com/glia/widgets/survey/SurveyController.java
AndriiHorishniiMOC: done

widgetssdk/src/main/java/com/glia/widgets/survey/SurveyController.java
andrews-moc: Cool 👍

andrews-moc (approved):

PR #281, opened 2022-05-09
Fix: hide keyboard on survey screen

AndriiHorishniiMOC:
Fix: hide keyboard if the visitor interacts with other elements after the text field

[MUIC-764](https://glia.atlassian.net/browse/MUIC-764)

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/survey/SurveyController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/survey/SurveyController.java
@@ -90,21 +90,36 @@ public class SurveyController implements SurveyContract.Controller {
                 .filter(item -> item.getQuestion().getId().equals(answer.getQuestionId()))
                 .findFirst()
                 .ifPresent(item -> {
-                    item.setAnswer(answer);
-                    if (item.isShowError()) {
-                        boolean showError;
-                        try {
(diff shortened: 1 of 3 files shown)


widgetssdk/src/main/java/com/glia/widgets/survey/SurveyController.java
```diff
@@ -104,6 +104,9 @@ public void onAnswer(@NonNull Survey.Answer answer) {
                             item.getAnswerCallback().answerCallback(showError);
                         }
                     }
+                    if (item.getQuestion().getType() != Survey.Question.QuestionType.TEXT) {
```
jakob-mc, line 107: Lot of different stuff and nesting happening in `public void onAnswer()`. Maybe you can refactor it a bit? For example move everything in `.ifPresent()` into a separate method.

andrews-moc (approved):
Agree with Jacob's suggestion for function extraction. Other than that - it's looking good. 👍

widgetssdk/src/main/java/com/glia/widgets/survey/SurveyController.java
AndriiHorishniiMOC: resolved

jakob-mc (approved):

deniss-mc (approved):

PR #282, opened 2022-05-10
[Android] Chat/Audio/Video – Operator status fixes

deniss-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-769

**Additional info:**
Also fixes issue when during image loading operator status is showing only brand primary and icon is gone

Needs validating:
https://glia.atlassian.net/browse/MUIC-733

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/view/OperatorStatusView.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/view/OperatorStatusView.java
@@ -18,6 +18,7 @@ import com.airbnb.lottie.model.KeyPath;
 import com.glia.widgets.R;
 import com.glia.widgets.UiTheme;
 import com.google.android.material.imageview.ShapeableImageView;
+import com.squareup.picasso.Callback;
 import com.squareup.picasso.Picasso;
 
 public class OperatorStatusView extends ConstraintLayout {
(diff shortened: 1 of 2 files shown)


AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/view/OperatorStatusView.java
```diff
@@ -115,22 +116,40 @@ public void showPlaceHolderWithIconPaddingOnConnect() {
...
 
     public void showProfileImage(String profileImgUrl) {
         updateProfilePictureViewSize(operatorImageSize);
-        Picasso.get().load(profileImgUrl).into(profilePictureView);
-        updatePlaceholderView(
-                operatorImageSize,
-                NO_PADDING,
-                GONE
-        );
+        Picasso.get().load(profileImgUrl).into(profilePictureView, new Callback() {
```
andrews-moc, line 138: According to [Picasso docs](https://square.github.io/picasso/2.x/picasso/com/squareup/picasso/RequestCreator.html#into-android.widget.ImageView-com.squareup.picasso.Callback-):
`Note: The Callback param is a strong reference and will prevent your Activity or Fragment from being garbage collected. If you use this method, it is strongly recommended you invoke an adjacent Picasso.cancelRequest(android.widget.ImageView) call to prevent temporary leaking.
`

So, it looks like, with the usage of callback, there's a need to make `Picasso.cancelRequest` at some point, isn't it?

deniss-mc: cancels request on detached

andrews-moc (approved):

PR #291, opened 2022-05-12
Fix end engagement dialog

AndriiHorishniiMOC:
Fix: Not show the "End engagement" dialog if the engagement ends by a Visitor

[MOB-1284](https://glia.atlassian.net/browse/MOB-1284)
[MUIC-768](https://glia.atlassian.net/browse/MUIC-768)

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
@@ -126,6 +126,7 @@ public class CallController implements
 
     private final String TAG = "CallController";
     private volatile CallState callState;
+    private boolean isVisitorEndEngagement = false;
 
     private final CompositeDisposable disposables = new CompositeDisposable();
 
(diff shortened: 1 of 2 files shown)


andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/core/engagement/domain/ShouldShowEndEngagementDialogUseCase.java
```diff
@@ -0,0 +1,13 @@
+package com.glia.widgets.core.engagement.domain;
+
+public class ShouldShowEndEngagementDialogUseCase {
```
jakob-mc, line 3: A good usecase behaves like a pure function. It has no globally mutable state. Has only one public method, usually called `execute()`, `invoke()` etc and the arguments are passed through that method. This makes the usecase easy to understand, predictable and easily testable. 

Here in this situation, I think usecase is not needed at all.

AndriiHorishniiMOC: UseCase was removed.

jakob-mc (approved):

PR #294, opened 2022-05-17
Tests for survey answers validation logic

andrews-moc:
**Related Jira issue:**
[Survey - Operator doesn't get survey report if Visitor answers only for required question](https://glia.atlassian.net/browse/MOB-1283)

**Additional info:**
The reason for the issue was in invalid answers to the survey.
Backend returns error if any answer is incorrect.
It was decided to write tests to make sure the business logic is correct.

### DIFF
--- a/widgetssdk/src/test/java/android/text/TextUtils.java
+++ b/widgetssdk/src/test/java/android/text/TextUtils.java
@@ -1,0 +1,7 @@
+package android.text;
+
+public class TextUtils {
+    public static boolean isEmpty(CharSequence str) {
+        return str == null || str.length() == 0;
+    }
+}
--- a/widgetssdk/src/test/java/android/text/TextUtils.java
+++ b/widgetssdk/src/test/java/android/text/TextUtils.java
@@ -1,5 +1,9 @@
 package android.text;
 
+/**
+ * Mock implementation of isEmpty() in android.text.TextUtils.
+ * It's needed for unit tests.
+ */
 public class TextUtils {
     public static boolean isEmpty(CharSequence str) {
         return str == null || str.length() == 0;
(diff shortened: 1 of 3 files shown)


AndriiHorishniiMOC (approved):

widgetssdk/src/test/java/android/text/TextUtils.java
```diff
@@ -0,0 +1,7 @@
+package android.text;
+
+public class TextUtils {
```
gugalo, line 3: It would be nice to have some comment here explaining that his class is here for test mocking

widgetssdk/src/test/java/com/glia/widgets/core/survey/domain/GliaSurveyAnswerUseCaseTest.java
```diff
@@ -0,0 +1,676 @@
...
+        when(question.isRequired()).thenReturn(false);
+        when(question.getType()).thenReturn(Survey.Question.QuestionType.TEXT);
+
+        Survey.Answer answer = null;
+        QuestionItem item = new QuestionItem(question, answer);
+        try {
+            gliaSurveyAnswerUseCase.validate(item);
+        } catch (Exception e) {
+            Assert.fail("Should not have thrown any exception");
+        }
```
gugalo, line 49: NIT: why do you wrap the tests in exception. Is it something from styling/guides? I mean the test will fail even if the method just throws the exception

gugalo (approved):

widgetssdk/src/test/java/com/glia/widgets/core/survey/domain/GliaSurveyAnswerUseCaseTest.java
andrews-moc: @gugalo Yes, the test will fail even if the method just throws an exception. The difference is, that with `Assert.fail` the test will fail with a readable assertion exception, and without it, the test will fail because of the crash.

widgetssdk/src/test/java/android/text/TextUtils.java
andrews-moc: Good suggestion, thank you. Added a comment. 58cf5850598c10d1508fe585fd7f44c82a29679a

gugalo (approved):

PR #298, opened 2022-05-30
Small fixes for ADA

AndriiHorishniiMOC:
Small fixes for ADA:
- fixed bug with buttons contend descriptions on the CallView.
- moved content description labels from `donottranslate.xml` to `string.xml`

### DIFF
--- a/widgetssdk/src/main/res/values/donottranslate.xml
+++ b/widgetssdk/src/main/res/values/donottranslate.xml
@@ -2,11 +2,5 @@
 <resources>
 
     <string name="glia_outlined_options_icon_content_description">Outlined options icon description</string>
-    <string name="glia_chat_media_upgrade_item_mic_content_description">Timer microphone </string>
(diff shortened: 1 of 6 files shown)


widgetssdk/src/main/res/values/donottranslate.xml
```diff
@@ -3,10 +3,5 @@
 
     <string name="glia_outlined_options_icon_content_description">Outlined options icon description</string>
     <string name="glia_chat_media_upgrade_item_mic_content_description">Timer microphone </string>
```
jakob-mc, line 5: should move those to strings.xml as well?

andrews-moc (approved):

widgetssdk/src/main/res/values/donottranslate.xml
AndriiHorishniiMOC: Fixed.
Also, it looks like the OutlinedOptionView is unused.

PR #307, opened 2022-06-10
MUIC-831 Unit tests for FilePreview feature

jakob-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-831

### DIFF
--- a/widgetssdk/src/androidTest/java/com/glia/widgets/DownloadsFolderDataSourceTest.java
+++ b/widgetssdk/src/androidTest/java/com/glia/widgets/DownloadsFolderDataSourceTest.java
@@ -19,31 +19,58 @@ import java.io.InputStream;
 @RunWith(AndroidJUnit4.class)
 public class DownloadsFolderDataSourceTest {
 
-    private Context appContext;
     private DownloadsFolderDataSource dataSource;
 
     @Before
     public void setUp() {
-        appContext = InstrumentationRegistry.getInstrumentation().getTargetContext();
+        Context appContext = InstrumentationRegistry.getInstrumentation().getTargetContext();
         dataSource = new DownloadsFolderDataSource(appContext);
     }
 
     @Test
-    public void putImageToDownloads_returnsCompletableCompletes() {
+    public void putImageToDownloads_completesSuccessfully_whenValidArguments() {
         dataSource.putImageToDownloads(IMAGE_NAME, BITMAP)
                 .test()
                 .assertComplete();
     }
 
     @Test
-    public void getImageFromDownloadsFolder_returnsMaybeOnError() {
+    public void putImageToDownloads_completesSuccessfully_whenNullImageNameArgument() {
+        dataSource.putImageToDownloads(null, BITMAP)
+                .test()
+                .assertComplete();
+    }
+
+    @Test
+    public void putImageToDownloads_emitsNullPointerException_whenNullBitmapArgument() {
+        dataSource.putImageToDownloads(IMAGE_NAME, null)
+                .test()
+                .assertError(NullPointerException.class);
+    }
+
+    @Test
+    public void putImageToDownloads_emitsNullPointerException_whenNullArguments() {
+        dataSource.putImageToDownloads(null, null)
+                .test()
+                .assertError(NullPointerException.class);
+    }
+
+    @Test
+    public void getImageFromDownloadsFolder_emitsFileNotFoundException_whenImageNotInDownloadsFolder() {
         dataSource.getImageFromDownloadsFolder(IMAGE_NAME)
                 .test()
                 .assertError(FileNotFoundException.class);
     }
 
     @Test
-    public void downloadFileToDownloads_returnsCompletableCompletes() {
+    public void getImageFromDownloadsFolder_emitsIllegalArgumentException_whenNullArgument() {
+        dataSource.getImageFromDownloadsFolder(null)
+                .test()
+                .assertError(IllegalArgumentException.class);
+    }
+
+    @Test
+    public void downloadFileToDownloads_completesSuccessfully_whenValidArguments() {
         dataSource.downloadFileToDownloads(
                         IMAGE_NAME,
                         "content_type",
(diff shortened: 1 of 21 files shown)


widgetssdk/src/androidTest/java/com/glia/widgets/DownloadsFolderDataSourceTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    public void putImageToDownloads_returnsCompletableCompletes() {
+        dataSource.putImageToDownloads(IMAGE_NAME, BITMAP)
+                .test()
+                .assertComplete();
+    }
+
+    @Test
+    public void getImageFromDownloadsFolder_returnsMaybeOnError() {
```
gugalo, line 39: If the test method says "returnsMaybeOnError" I would expect to see some check that also validates that it returns Maybe. But even though it does return the Maybe it is not what this test checks so I think it can be changed to just "getImageFromDownloadsFolder_returnsError" which more clearly describes the test case. What is more, it would be nice to understand why we expect an error in this test.

widgetssdk/src/androidTest/java/com/glia/widgets/DownloadsFolderDataSourceTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.io.ByteArrayInputStream;
+import java.io.FileNotFoundException;
+import java.io.InputStream;
+
+@RunWith(AndroidJUnit4.class)
+public class DownloadsFolderDataSourceTest {
```
gugalo, line 20: Why there is only a single test case for every method. Is it impossible to cover this class with more tests?

DavDo (changes requested):
I would like to suggest having some naming convention for method names first of all, as in most cases you are using something like this **MethodName_ExpectedBehavior_StateUnderTest** I would like to suggest following this naming strategy.
And the second thing I would suggest is to have a more descriptive naming of methods, for example -
```
 @Test
 public void setView() {}
 @Test
 public void onSharePressed() {}
```
It is unclear what these methods are testing and what result is expected from them.
Also 
```
@Test
public void putBitmap_getBitmapById_returnsTheSameBitmap_whenBitmapUpdated(){}
```
no need to include preparation steps (putBitmap) to the method name.

```
@Test
public void putImageToDownloads_returnsCompletableCompletes() {}
```
missing state under the test/condition, and I think `returnsCompletableCompletes()` is not clear

widgetssdk/src/test/java/com/glia/widgets/filepreview/data/GliaFileRepositoryTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.InputStream;
+
+import io.reactivex.Completable;
+import io.reactivex.Maybe;
+
+public class GliaFileRepositoryTest {
```
gugalo, line 28: Same test functions naming change suggestions as in other class.

widgetssdk/src/test/java/com/glia/widgets/filepreview/data/GliaFileRepositoryTest.java
```diff
@@ -0,0 +1,149 @@
...
+        Maybe<Bitmap> result = gliaFileRepository.loadImageFromCache(FILENAME);
+        result.test().assertResult(BITMAP);
+    }
+
+    @Test
+    public void loadImageFromCache_returnsMaybeOnError() {
+        doThrow(RUNTIME_EXCEPTION).when(bitmapCache).getBitmapById(FILENAME);
+        Maybe<Bitmap> result = gliaFileRepository.loadImageFromCache(FILENAME);
+        result.test().assertError(RUNTIME_EXCEPTION);
+    }
```
gugalo, line 62: Should also test a case when Bitmap returned by the `bitmapCache.getBitmapById` was `null`

gugalo (commented):
I think the Bitrise workflow that should run the Unit Tests and the Instrumented tests are missing. I would be nice to have it before we merge this PR. @dukhovnyi

gugalo:
I have updated the triggers on the Bitrise side. It should try running the tests next time you update this branch

gugalo:
I also have the same issue with the build for UI tests as the Bitrise. @jakob-mc It seems this issue was not introduced by your changes, do you also have the problem when building for the instrumented tests because of [this line](https://github.com/salemove/android-sdk-widgets/blob/e7911ab3109ec0c3bec813c71a2649e3bdb3320f/widgetssdk/src/androidTest/java/com/glia/widgets/CallActivityTest.java#L328)?

jakob-mc:
> I have the same issue with the build for UI tests as the Bitrise. @jakob-mc It seems this issue was not introduced by your changes, do you also have the problem when building for the instrumented tests because of [this line](https://github.com/salemove/android-sdk-widgets/blob/e7911ab3109ec0c3bec813c71a2649e3bdb3320f/widgetssdk/src/androidTest/java/com/glia/widgets/CallActivityTest.java#L328)?

Yes, I have the same compile error, and was not introduced by me.

jakob-mc:
> Could you look into fixing these tests? You might want to have a separate Jira ticket for that.

yes, I'll try to fix it.

AndriiHorishniiMOC (approved):

gugalo:
> it builds now but looks like theres another error in Bitrise logs. Looks like ci conf issue?
I will look into this today

gugalo:
Fixed the CI. But it seems that there is still some issue:
```
NullPointerException: Attempt to invoke virtual method 'int com.glia.widgets.helper.ResourceProvider.getDimension(int)' on a null object reference
```
Issue is also reproducible for me locally when I try to run instrumentation tests.
Full error report can be found here: https://addons-testing.bitrise.io/builds/b8d0cee4-a5c9-4ed6-a8e6-0448988cd66b/testreport/ftl/testsuite/0/testcases?status=failed

jakob-mc:
> Fixed the CI. But it seems that there is still some issue:
> 
> ```
> NullPointerException: Attempt to invoke virtual method 'int com.glia.widgets.helper.ResourceProvider.getDimension(int)' on a null object reference
> ```
> 
> Issue is also reproducible for me locally when I try to run instrumentation tests. Full error report can be found here: https://addons-testing.bitrise.io/builds/b8d0cee4-a5c9-4ed6-a8e6-0448988cd66b/testreport/ftl/testsuite/0/testcases?status=failed

Looks like it takes a bit more time to fix it. I'll create a new task for that. Can you approve this PR?

PR #317, opened 2022-06-24
Navigate to chat or call screen when engagement is ended to ensure that survey was shown (if enabled)

andrews-moc:
**Jira issue:**
[Survey with broken buttons appears if a Visitor starts another call with opened survey](https://glia.atlassian.net/browse/MOB-1383)


AndriiHorishniiMOC (approved):
Looks good, but I have alternative solution https://github.com/salemove/android-sdk-widgets/pull/322

DavDo (approved):

gugalo (approved):

PR #318, opened 2022-06-27
Call Activity Configuration Impl

deniss-mc:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-318

**Additional info:**
Part of MUIC-318 implementation - created a builder class for Call Activity in order to hold Activity configuration and ease up serialisation process

Note for future implementation: I propose to remove WidgetsConfiguration from Activity Initialisation and receive it via some Provider instead. It will require deprecation notice and a lot of rework on how things are tied up. But the configuration will be held and will not be lost in case activity is started from service (or deeplink)

(with current implementation it will be matter of moving sdkConfiguration out of the builder and activity configuration to proper widgets configuration holder class :) )

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
@@ -21,6 +21,7 @@ import com.glia.widgets.R;
 import com.glia.widgets.chat.ChatActivity;
 import com.glia.widgets.core.configuration.GliaSdkConfiguration;
 import com.glia.widgets.helper.Logger;
+import com.glia.widgets.helper.Utils;
 import com.glia.widgets.survey.SurveyActivity;
 
 import java.security.InvalidParameterException;
@@ -225,7 +226,7 @@ public class CallActivity extends AppCompatActivity {
                 Configuration.Builder
                         .builder()
                         .setWidgetsConfiguration(sdkConfiguration)
-                        .setMediaType(toMediaType(mediaType))
+                        .setMediaType(Utils.toMediaType(mediaType))
                         .build()
         );
     }
@@ -238,17 +239,6 @@ public class CallActivity extends AppCompatActivity {
                 .from(context)
                 .setConfiguration(configuration)
                 .getIntent();
-    }
-
-    public static Engagement.MediaType toMediaType(String mediaType) {
-        switch (mediaType) {
-            case GliaWidgets.MEDIA_TYPE_VIDEO:
-                return Engagement.MediaType.VIDEO;
-            case GliaWidgets.MEDIA_TYPE_AUDIO:
-                return Engagement.MediaType.AUDIO;
-            default:
-                throw new InvalidParameterException("Invalid Media Type");
-        }
     }
 
     public static class Configuration {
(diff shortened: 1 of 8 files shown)


widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
```diff
@@ -187,71 +187,219 @@ private boolean isPermissionDataMediaRequestCode(Pair<Integer, Integer[]> permis
...
+            case GliaWidgets.MEDIA_TYPE_VIDEO:
+                return Engagement.MediaType.VIDEO;
+            case GliaWidgets.MEDIA_TYPE_AUDIO:
+                return Engagement.MediaType.AUDIO;
+            default:
+                throw new InvalidParameterException("Invalid Media Type");
+        }
+    }
+
+    public static class Configuration {
```
jakob-mc, line 254: does the Configuration and all other static builder classes need to be inner classes? They should not bloat the CallActivity.java.

widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
```diff
@@ -187,71 +187,219 @@ private boolean isPermissionDataMediaRequestCode(Pair<Integer, Integer[]> permis
...
+            return new IntentBuilder(context);
+        }
+
+        public IntentBuilder setConfiguration(@NonNull Configuration configuration) {
+            this.configuration = configuration;
+            return this;
+        }
+
+        public Intent getIntent() {
+            if (configuration == null) throw new RuntimeException("Configuration " +
```
jakob-mc, line 342: would be more readable if you move the `throw new...` to the next line.

widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
```diff
@@ -187,71 +187,219 @@ private boolean isPermissionDataMediaRequestCode(Pair<Integer, Integer[]> permis
...
+                        .setIsUpgradeToCall(configuration.isUpgradeToCall);
+            }
+
+            public static Builder builder() {
+                return new Builder();
+            }
+        }
+    }
+
+    private static class IntentBuilder {
```
jakob-mc, line 324: do you really need all that overhead? could it be done by a single method in activity class `makeIntent(configuration);`

widgetssdk/src/main/java/com/glia/widgets/view/head/ChatHeadView.java
```diff
@@ -143,13 +143,15 @@ public void navigateToChat() {
...
     public void navigateToCall() {
-        getContext().startActivity(
-                getNavigationIntent(
-                        getContext(),
-                        CallActivity.class,
-                        sdkConfiguration
-                )
-        );
+        CallActivity.Configuration activityConfig =
+                CallActivity.Configuration.Builder.builder()
```
jakob-mc, line 147: imo `new CallActivity.Configuration.Builder()` looks better than `.Builder.builder()`

jakob-mc (changes requested):

jakob-mc (approved):

PR #323, opened 2022-07-01
Deep link init

AndriiHorishniiMOC:
Init Glia SDK with deep link

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/Application.java
+++ b/app/src/main/java/com/glia/exampleapp/Application.java
@@ -1,5 +1,12 @@
 package com.glia.exampleapp;
 
+import android.app.Activity;
+import android.os.Bundle;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.glia.androidsdk.Glia;
 import com.glia.widgets.GliaWidgets;
 
 public class Application extends android.app.Application {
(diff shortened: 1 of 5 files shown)


DavDo (approved):

app/src/main/java/com/glia/exampleapp/Application.java
```diff
@@ -8,6 +15,38 @@ public class Application extends android.app.Application {
...
         super.onCreate();
         GliaWidgets.onAppCreate(this);
+
+        initGliaWidgets();
     }
 
+    private void initGliaWidgets() {
+        registerActivityLifecycleCallbacks(new ActivityLifecycleCallbacks() {
+            @Override
+            public void onActivityCreated(@NonNull Activity activity, @Nullable Bundle savedInstanceState) {
```
gugalo, line 25: I would add some comment to eplain why it is done like that. Otherwise it might be very confusing if you don't know about the deep link logic inside the main activity

app/src/main/java/com/glia/exampleapp/GliaWidgetsConfigManager.java
```diff
@@ -46,17 +46,21 @@ public static GliaWidgetsConfig obtainConfig(@NonNull Intent intent, @NonNull Co
...
+        saveSiteIdToPrefs(data, applicationContext);
 
 //        We're not checking availability for every query param separately because this is for acceptance tests and we assume that link would be correct
-        if (SECRET_KEY.equals(data.getLastPathSegment()))
+        if (SECRET_KEY.equals(data.getLastPathSegment())) {
+            saveSiteApiKeyAuthToPrefs(data, applicationContext);
             return obtainConfigFromSecretDeepLink(data, applicationContext);
-        else if (TOKEN_KEY.equals(data.getLastPathSegment()))
+        } else if (TOKEN_KEY.equals(data.getLastPathSegment())) {
+            saveAppTokenAuthToPrefs(data, applicationContext);
```
gugalo, line 59: NIT: As far as I understood we can remove the app token logic. But maube it can be done in separate PR

gugalo (approved):
Seem good but I added a couple of minor comments.

PR #332, opened 2022-07-23
Disable sending message during engagement transferring

DavDo:
MOB-1456

**Jira issue:**
https://glia.atlassian.net/browse/MOB-1456

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/model/ChatState.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/model/ChatState.java
@@ -154,6 +154,17 @@ public class ChatState {
                 .createChatState();
     }
 
+    public ChatState transferring() {
+        return new Builder()
+                .copyFrom(this)
+                .setOperatorName(null)
+                .setOperatorProfileImgUrl(null)
+                .setEngagementRequested(true)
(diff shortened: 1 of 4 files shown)


AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/model/ChatState.java
```diff
@@ -484,5 +495,18 @@ public Builder setVisitorContextAssetId(String visitorContextAssetId) {
             this.visitorContextAssetId = visitorContextAssetId;
             return this;
         }
+
+        public Builder enableChatPanel() {
+            setChatInputMode(ChatInputMode.ENABLED);
```
andrews-moc, line 500: I see `setShowSendButton(false)` in `disableChatPanel()`.
May be we need to add `setShowSendButton(true)` to `enableChaaatPanel()` function accordingly?

DavDo: It seems send button needs to be shown after text input.

andrews-moc (approved):

PR #338, opened 2022-08-09
Glia overlay issue in Android restricting to tap on back button

DavDo:
MOB - 1477

**Jira issue:**
https://glia.atlassian.net/browse/MOB-1477

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.java
@@ -49,10 +49,20 @@ public class DialogController {
         return dialogsState instanceof DialogsState.OverlayPermissionsDialog;
     }
 
+    public boolean isExitQueueDialogShown() {
+        return dialogsState instanceof DialogsState.ExitQueueDialog;
+    }
+
+    public void dismissCurrentDialog() {
+        Logger.d(TAG, "Dismiss current dialog");
+        emitDialogState(new DialogsState.NoDialog());
+        runPostponedOperation();
+    }
+
     public void dismissDialogs() {
         Logger.d(TAG, "Dismiss dialogs");
         emitDialogState(new DialogsState.NoDialog());
-        runPostponedOperation();
+        postponedOperation = null;
     }
 
     private synchronized void emitDialogState(DialogsState state) {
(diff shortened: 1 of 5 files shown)


widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.java
```diff
@@ -49,10 +49,20 @@ public boolean isOverlayDialogShown() {
         return dialogsState instanceof DialogsState.OverlayPermissionsDialog;
     }
 
+    public boolean isExitingDialogShown() {
```
andrews-moc, line 52: Minor suggestion: I think `isExitQueueDialogShown()` is a little bit better name because "exiting" is very much close to "exciting". 🙂

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.java
```diff
@@ -49,10 +49,20 @@ public boolean isOverlayDialogShown() {
...
+        Logger.d(TAG, "Dismiss current dialog");
+        emitDialogState(new DialogsState.NoDialog());
+        runPostponedOperation();
+    }
+
     public void dismissDialogs() {
         Logger.d(TAG, "Dismiss dialogs");
         emitDialogState(new DialogsState.NoDialog());
-        runPostponedOperation();
+        postponedOperation = null;
```
andrews-moc, line 65: Not sure about this change. It looks like without a call to `runPostponedOperation()` this function doesn't dismiss the dialog anymore. Is it okay?

DavDo: @andrews-moc actually, I renamed the old method to `dismissCurrentDialog()` which at the end runs the postponed one, and also created a new method `dismissDialogs()` which dismisses all dialogs including the postponed one, I'm using the second one only when user clicks `yes` on `exitFromQueue`  dialog.

andrews-moc: Okay ✔️

gugalo (approved):

PR #342, opened 2022-08-15
Fix video is not started if an operator makes a request twice

DavDo:
MOB - 1453

**Jira issue:**
https://glia.atlassian.net/browse/MOB-1453

**Additional info:**
Before these changes, we were keeping only one instance of `DialogsState` and were able to show only one dialog at a time, which means in some cases we lost important information or had a broken logic e.g. when some dialog was shown and the second was arriving the same time we just kept the first one and lost the second that's why the video or audio was not started if an operator makes a request twice (we were answering to the old request) and so on. Also, I changed a lot of implementations of DialogState to @intDef which is faster. 

**Screenshots:**


widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogManager.java
```diff
@@ -0,0 +1,79 @@
...
+        stateDeque = new LinkedBlockingDeque<>();
+    }
+
+    public void offer(DialogState state) {
+        if (isDialogShown()) {
+            emitNone();
+        }
+
+        boolean removeLastoccurence = stateDeque.removeLastOccurrence(state);
+        boolean offer = stateDeque.offer(state);
```
andrews-moc, line 27: Looks like `removeLastoccurence` and `offer` variables aren't used, are they?
And minor catch - `removeLastoccurence` => `removeLastOccurence`

DavDo: > Looks like `removeLastoccurence` and `offer` variables aren't used, are they?
> And minor catch - `removeLastoccurence` => `removeLastOccurence`

You're right, I added variables for test purposes and forgot to delete them.

andrews-moc:
@DavDo Could you also please describe the general idea behind these changes? It will help to understand and review the changes better.

DavDo:
> @DavDo Could you also please describe the general idea behind these changes? It will help to understand and review the changes better.

Updated` Additional Info:` section

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.java
```diff
@@ -571,34 +571,41 @@ private void updateNewMessageOperatorStatusView(String operatorProfileImgUrl) {
...
-                post(this::showNoMoreOperatorsAvailableDialog);
-            } else if (dialogsState instanceof DialogsState.EngagementEndedDialog) {
-                post(this::showEngagementEndedDialog);
-            } else if (dialogsState instanceof DialogsState.StartScreenSharingDialog) {
-                post(this::showScreenSharingDialog);
-            } else if (dialogsState instanceof DialogsState.EnableNotificationChannelDialog) {
-                post(this::showAllowNotificationsDialog);
-            } else if (dialogsState instanceof DialogsState.EnableScreenSharingNotificationsAndStartSharingDialog) {
-                post(this::showAllowScreenSharingNotificationsAndStartSharingDialog);
+        dialogCallback = dialogState -> {
```
AndriiHorishniiMOC, line 574: I don't like those code duplications in ChatView and CallView. Maybe we should refactor it in the future.

DavDo: > I don't like those code duplications in ChatView and CallView. Maybe we should refactor it in the future.

I'll try to refactor tomorrow.

andrews-moc (approved):

gugalo (approved):

PR #346, opened 2022-08-23
Fix: show Overlay permissions Dialog

AndriiHorishniiMOC:
Fix: show current dialog state on add callback to the DialogController

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.java
@@ -109,13 +109,16 @@ public class DialogController {
     }
 
     public void showEnableScreenSharingNotificationsAndStartSharingDialog() {
-            Logger.d(TAG, "Show Enable Notification Channel Dialog");
+        Logger.d(TAG, "Show Enable Notification Channel Dialog");
         dialogManager.offer(new DialogState(Dialog.MODE_ENABLE_SCREEN_SHARING_NOTIFICATIONS_AND_START_SHARING));
     }
 
     public void addCallback(Callback callback) {
         Logger.d(TAG, "addCallback");
         viewCallbacks.add(callback);
+        if (viewCallbacks.size() == 1) {
+            dialogManager.showNext();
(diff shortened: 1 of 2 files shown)


widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.java
```diff
@@ -109,13 +109,16 @@ public void showEnableCallNotificationChannelDialog() {
...
     public void showEnableScreenSharingNotificationsAndStartSharingDialog() {
-            Logger.d(TAG, "Show Enable Notification Channel Dialog");
+        Logger.d(TAG, "Show Enable Notification Channel Dialog");
         dialogManager.offer(new DialogState(Dialog.MODE_ENABLE_SCREEN_SHARING_NOTIFICATIONS_AND_START_SHARING));
     }
 
     public void addCallback(Callback callback) {
         Logger.d(TAG, "addCallback");
         viewCallbacks.add(callback);
+        if (viewCallbacks.size() == 1) {
```
andrews-moc, line 119: Maybe it makes sense to check `if >= 1`?

DavDo: I think with `size() == 1` we are checking that this is the first callback added, if size > 1, it means that some view already handled that callback, and no need to trigger it again.

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.java
AndriiHorishniiMOC: Yeah. The idea is to show a dialog if the state was added before the callback registered.

andrews-moc: It might be good to add a comment about this I think. 😉

gugalo (approved):

AndriiHorishniiMOC:
!merge

gugalo (approved):

PR #347, opened 2022-08-23
Fix: show dialogs by list

AndriiHorishniiMOC:
Added possibility to show dialogs one by one.
Used to show the No More Operators Dialog (returned to previous logic).

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogManager.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogManager.java
@@ -1,6 +1,7 @@
 package com.glia.widgets.core.dialog;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.VisibleForTesting;
 
 import com.glia.widgets.core.dialog.model.DialogState;
 
@@ -10,7 +11,8 @@ import java.util.concurrent.LinkedBlockingDeque;
 public final class DialogManager {
     @NonNull
     private final Callback callback;
-    private final Deque<DialogState> stateDeque;
+    @VisibleForTesting
+    final Deque<DialogState> stateDeque;
     private final DialogState none = new DialogState(Dialog.MODE_NONE);
 
     public DialogManager(@NonNull Callback callback) {
@@ -18,7 +20,13 @@ public final class DialogManager {
         stateDeque = new LinkedBlockingDeque<>();
     }
 
-    public void offer(DialogState state) {
+    /**
+     * Used to add state to the queue.
+     *
+     * It will emit a new state immediately. The previous state is saved in the queue
+     * and will be emitted after the {@link #dismissCurrent()} is called.
+     */
+    public void addAndEmit(@NonNull DialogState state) {
         if (isDialogShown()) {
             emitNone();
         }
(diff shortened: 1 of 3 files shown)


andrews-moc (approved):

DavDo (approved):

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogManager.java
```diff
@@ -28,12 +28,23 @@ public void offer(DialogState state) {
...
+    }
+
     private DialogState remove() {
-        return stateDeque.pollLast();
+        return stateDeque.pollFirst();
     }
 
     private DialogState getCurrent() {
-        return stateDeque.peekLast();
+        return stateDeque.peekFirst();
```
AndriiHorishniiMOC, line 47: Guys, maybe I made a mistake on it. Need to write tests for it.

DavDo: The previous implementation was working with the `LIFO` method, you changed it to `FIFO`.
I thought that new dialogs/events/warnings are more important than already opened ones.

AndriiHorishniiMOC: I did it cause the Deque can add the element only on the end(or start) of the stack. To use LIFO in our add method we should add an element to the penultimate place.
So, to use LIFO we can:
1. Use hack
```
public void add(DialogState state) {
        ...
        DialogState currentState = remove();
        stateDeque.offer(state);
        stateDeque.offer(currentState);
        ...
    }
```
2. Use List instead of Deque

DavDo: Agree with you, but I'm not sure that we need to keep `OverlayDialog` or `ExitQueueDialog` in case of `NoMoreOperatorsAvailable` received because after that there is no need for `OverlayDialog` or `ExitQueueDialog` and we must close the chat screen anyways. In this case no need for the `add` method anymore.

AndriiHorishniiMOC: Maybe you are right about ExitQueueDialog. But OverlayDialog we can show only once. After a user interacts with NoMoreOperatorsAvailable it disappears.
Anyway, I returned to the previous logic of displaying dialogs, may be it made sence according to the requirements.

DavDo: That was only my thoughts I really don't know about the requirements, in this case maybe it makes sense to remove only `ExitQueueDialog` and keep the `add` method as it was.

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogManager.java
AndriiHorishniiMOC: @DavDo but this is your changes regarding `ExitQueueDialog`
https://github.com/salemove/android-sdk-widgets/pull/338/files#diff-23ee758ebee1fe03bc71ea9297219470e1fb17c242291fa8f016be1e926d939dR137

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogManager.java
```diff
@@ -50,17 +85,26 @@ private void emitNone() {
...
+     * Used to remove the current state and emit the next state from the queue.
+     */
     public void dismissCurrent() {
         emitNone();
         remove();
         showNext();
     }
 
+    /**
+     * Used clear the queue and emit {@link Dialog#MODE_NONE} state.
```
andrews-moc, line 98: Minior suggestion:
Used clear the queue => Used to clear the queue

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogManager.java
```diff
@@ -10,15 +11,22 @@
...
         this.callback = callback;
         stateDeque = new LinkedBlockingDeque<>();
     }
 
-    public void offer(DialogState state) {
+    /**
+     * Used to add state to the queue.
+     *
+     * It will emit a new state immediately. The previous state saved in the queue
+     * and will emit after the {@link #dismissCurrent()} called.
```
andrews-moc, line 27: Minor suggestion:
The previous is state saved in the queue and will be emitted after the {@link #dismissCurrent()} is called.

andrews-moc (approved):

andrews-moc:
@AndriiHorishniiMOC Good job 👍

dukhovnyi:
!squash

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogManager.java
DavDo: I'm sorry I forgot about that bug.

dukhovnyi:
!merge

PR #350, opened 2022-09-05
update chat icons and add operator caching mechanism regarding ChatMe…

DavDo:
…ssage changes

MOB - 1542

**Jira issue:**
https://glia.atlassian.net/browse/MOB-1542

**Additional info:**

**Screenshots:**
image

### DIFF
--- a/widgetssdk/build.gradle
+++ b/widgetssdk/build.gradle
@@ -40,7 +40,6 @@ android {
 }
 
 dependencies {
-
     implementation "androidx.appcompat:appcompat:$appCompatVersion"
     implementation "com.google.android.material:material:$materialVersion"
     implementation "androidx.constraintlayout:constraintlayout:$constraintLayoutVersion"
@@ -55,7 +54,6 @@ dependencies {
     // https://square.github.io/leakcanary/changelog/
     // debugImplementation "com.squareup.leakcanary:leakcanary-android:$leakCanaryVersion"
 
-    // Replaced with a local AAR file
     api "com.glia:android-sdk:$gliaSdkVersion"
     // Glia sdk aar
     // implementation(name: "glia-core-sdk-$gliaSdkVersion", ext: 'aar')
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/model/history/OperatorAttachmentItem.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/model/history/OperatorAttachmentItem.java
@@ -7,19 +7,24 @@ import com.glia.widgets.helper.Utils;
 
 import java.util.Objects;
 
-public class OperatorAttachmentItem extends ChatItem {
+public class OperatorAttachmentItem extends OperatorChatItem {
 
-    public final boolean showChatHead;
     public final AttachmentFile attachmentFile;
-    public final String operatorProfileImgUrl;
     public final boolean isFileExists;
     public final boolean isDownloading;
 
-    public OperatorAttachmentItem(String chatItemId, int viewType, boolean showChatHead, AttachmentFile attachmentFile, String operatorProfileImgUrl, boolean isFileExists, boolean isDownloading) {
-        super(chatItemId, viewType);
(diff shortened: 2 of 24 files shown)


widgetssdk/build.gradle
```diff
@@ -55,7 +54,6 @@ dependencies {
     // https://square.github.io/leakcanary/changelog/
     // debugImplementation "com.squareup.leakcanary:leakcanary-android:$leakCanaryVersion"
 
-    // Replaced with a local AAR file
```
andrews-moc, line 58: I think this line shouldn't be deleted

DavDo: to be honest, I don't understand why we need this comment, in fact, dependency is not replaced with local aar.

widgetssdk/src/main/java/com/glia/widgets/chat/model/history/OperatorAttachmentItem.java
```diff
@@ -7,51 +7,51 @@
...
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
+        if (!(o instanceof OperatorAttachmentItem)) return false;
         if (!super.equals(o)) return false;
         OperatorAttachmentItem that = (OperatorAttachmentItem) o;
-        return showChatHead == that.showChatHead &&
-                Objects.equals(attachmentFile, that.attachmentFile) &&
-                Objects.equals(operatorProfileImgUrl, that.operatorProfileImgUrl) &&
```
andrews-moc, line 35: Just for my understanding, what is the reason for removing `Objects.equals(operatorProfileImgUrl, that.operatorProfileImgUrl)` ?

widgetssdk/src/main/java/com/glia/widgets/chat/model/history/OperatorAttachmentItem.java
```diff
@@ -7,51 +7,51 @@
...
-                Objects.equals(isFileExists, that.isFileExists) &&
-                Objects.equals(isDownloading, that.isDownloading);
+        return isFileExists == that.isFileExists && isDownloading == that.isDownloading && Objects.equals(attachmentFile, that.attachmentFile);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(super.hashCode(), showChatHead, attachmentFile, operatorProfileImgUrl, isFileExists, isDownloading);
```
andrews-moc, line 42: Just for my understanding, what is the reason for removing `showChatHead`, `operatorProfileImgUrl`?

widgetssdk/build.gradle
DavDo: we can add the local arr file and disable dependency manually, but this is only for test purposes on the local environment, that's why I found that line pointless.

widgetssdk/src/main/java/com/glia/widgets/chat/model/history/OperatorAttachmentItem.java
DavDo: The `operatorProfileImgUrl` field itself is removed and moved to the superclass

widgetssdk/build.gradle
andrews-moc: I see.

Maybe it makes sense to edit the comment?

I mean change "Replaced with a local AAR file" to something like "Disable this line if dependency shouldbe taken from local AAR file".

Although it's not related to the current feature. 🙂

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/model/history/OperatorAttachmentItem.java
DavDo: moved to the `OperatorChatItem.java` class, which is the supertype for OperatorAttachmentItem and OperatorMessageItem.

AndriiHorishniiMOC (approved):

PR #361, opened 2022-09-20
Add workflow to increment Core SDK version

gersonnoboa:
Using the previous work done in Gradle, a new workflow has been added to `bitrise.yml` that increments the Core SDK version. A pull request is created also by directly calling the GitHub API. This will be used in the Core SDK release process, in order to increment the Core SDK version used in the widgets once the Core SDK is released. This workflow requires an environment variable called `NEW_VERSION`.

MOB-1510

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -231,6 +231,39 @@ workflows:
     - opts:
         is_expand: false
       INTEGRATOR_VARIANT: debug
+  upgrade_dependencies:
+    steps:
+    - activate-ssh-key@4: {}
+    - git-clone@6: {}
+    - cache-pull@2: {}
+    - install-missing-android-tools@3:
+        inputs:
+        - gradlew_path: $PROJECT_LOCATION/gradlew
+    - gradle-runner@2:
+        inputs:
+        - gradlew_path: $PROJECT_LOCATION/gradlew
+        - gradle_file: $PROJECT_LOCATION/build.gradle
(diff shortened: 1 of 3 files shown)


bitrise.yml
```diff
@@ -231,6 +231,39 @@ workflows:
...
       INTEGRATOR_VARIANT: debug
+  upgrade_dependencies:
+    steps:
+    - activate-ssh-key@4: {}
+    - git-clone@6: {}
+    - cache-pull@2: {}
+    - install-missing-android-tools@3:
+        inputs:
+        - gradlew_path: $PROJECT_LOCATION/gradlew
+    - gradle-runner@2:
```
gugalo, line 242: I think it would be nice if you add some meaningful titles to this and the next step to make it easier to read. You can find example in this file

bitrise.yml
```diff
@@ -231,6 +231,39 @@ workflows:
...
+        inputs:
+        - gradlew_path: $PROJECT_LOCATION/gradlew
+        - gradle_file: $PROJECT_LOCATION/build.gradle
+        - gradle_task: saveCoreSdkVersion --coreSdkVersion=$NEW_VERSION
+    - script@1:
```
gugalo, line 247: Does this step do the same thing as 'scripts/create-pr-for-increment-project-version.sh'?
In this case, maybe you can just run that script not to duplicate logic?

gugalo (approved):

gersonnoboa:
Merged as part of this PR https://github.com/salemove/android-sdk-widgets/pull/368/files

PR #364, opened 2022-09-29
Use RemoteConfiguration in UiTheme

AndriiHorishniiMOC:
Create a base mechanism to use RemoteConfiguration in UiTheme

MOB-1571

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/UiTheme.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/UiTheme.java
@@ -6,10 +6,13 @@ import android.os.Parcelable;
 import androidx.annotation.ColorRes;
 import androidx.annotation.DrawableRes;
 import androidx.annotation.FontRes;
+import androidx.annotation.NonNull;
 
 import com.glia.widgets.view.configuration.ButtonConfiguration;
 import com.glia.widgets.view.configuration.ChatHeadConfiguration;
 import com.glia.widgets.view.configuration.TextConfiguration;
+import com.glia.widgets.view.configuration.call.CallStyle;
+import com.glia.widgets.view.configuration.chat.ChatStyle;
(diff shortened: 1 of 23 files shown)


DavDo (approved):

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/UiTheme.java
```diff
@@ -1404,202 +1424,252 @@ public String getAppBarTitle() {
         return appBarTitle;
     }
 
+    @Deprecated
```
andrews-moc, line 1427: You will probably need to add docs here to explain the alternative for the deprecated function(s).

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/UiTheme.java
AndriiHorishniiMOC: Yes and no.
We don't have documentation for those methods, so they are our internal methods.
For now, we don't have the alternatives, I just mark the methods that should be moved from here.

But I can link this method with new in the next PRs of the RemoteConfiguration after those methods will be done.

widgetssdk/src/main/java/com/glia/widgets/core/configuration/GliaSdkConfiguration.java
```diff
@@ -17,6 +18,7 @@ public class GliaSdkConfiguration {
     private final String contextAssetId;
     @Deprecated
     private final String contextUrl;
+    private final RemoteConfiguration remoteConfiguration;
```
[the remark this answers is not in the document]
AndriiHorishniiMOC: Good idea!

AndriiHorishniiMOC:
!merge

PR #368, opened 2022-10-04
Add workflows to automate version increases

gersonnoboa:
There's more info in each commit, but TL:DR there was work required in the Gradle scripts to have the jobs run correctly, and then some work in Bitrise was added so that both project version increase and Core SDK version increase are automated.

As always, copying bitrise.yml into our repository brings a plethora of unwanted changes, but I cannot do anything about it.

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -1,6 +1,6 @@
 ---
 format_version: '8'
-default_step_lib_source: 'https://github.com/bitrise-io/bitrise-steplib.git'
+default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
 project_type: android
 workflows:
   browserstack_upload:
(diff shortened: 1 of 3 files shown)


bitrise.yml
```diff
@@ -35,57 +35,103 @@ workflows:
...
-            /usr/lib/jvm/java-11-openjdk-amd64/bin/java
-
+        - content: |-
+            sudo update-alternatives --set javac /usr/lib/jvm/java-11-openjdk-amd64/bin/javac
+            sudo update-alternatives --set java /usr/lib/jvm/java-11-openjdk-amd64/bin/java
             export JAVA_HOME='/usr/lib/jvm/java-11-openjdk-amd64'
-
-            envman add --key JAVA_HOME --value
-            '/usr/lib/jvm/java-11-openjdk-amd64'
+            envman add --key JAVA_HOME --value '/usr/lib/jvm/java-11-openjdk-amd64'
```
gugalo, line 42: I wonder why this script is needed. Seems like a fix. But I guess this was not introduced by recent changes

gugalo (approved):

bitrise.yml
gersonnoboa: I have no idea indeed, seems weird because android-sdk doesn't have anything similar but it works so I didn't want to touch anything.

gersonnoboa:
!merge

PR #370, opened 2022-10-05
Increment project version to 1.9.0

sm-deployer:


gradle.deps
```diff
@@ -0,0 +1 @@
+118d697fd46fae8c18dfe7c5b4ef9e89a15a9f769f22d112ce7498443fb80f4f95bc040b7c51663ec2e0f4e646e834bd230bf7e8c5301e0978e0696eba25be3ea2e3fedcc589f82179f8a8dd86957dcc9d8f6fd9af7cfe7033906182d7ed04126e36bbcbbdd0d17fa1779b108001849c
```
gugalo, line 1: I don't think this should be committed. I can't even figure out why this have be generated

gersonnoboa: That is strange because `gradle.deps` is part of the [.gitignore](https://github.com/salemove/android-sdk-widgets/blob/a744f7ff5da693ab75ae334e582bc8e75e3a3dc6/.gitignore#L96). Why would it still be here?

gersonnoboa (approved):

gugalo (approved):

gersonnoboa:
!merge

PR #371, opened 2022-10-10
Android App crashes when the visitor/operator ends the engagement call

DavDo:
MOB - 1586

**Jira issue:**
https://glia.atlassian.net/browse/MOB-1586
I can't figure out how, but in some cases, Survey styles are requested before we set it, that's why I added the initial style to the constructor of SurveyAdapter.
**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/survey/SurveyAdapter.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/survey/SurveyAdapter.java
@@ -58,6 +58,8 @@ public class SurveyAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder>
 
     public SurveyAdapter(SurveyAdapterListener listener) {
         this.listener = listener;
+        // initialize style with default empty SurveyStyle, to make sure that style usage is safe even if new style is not set.
+        this.style = new SurveyStyle.Builder().build();
     }
 
     public void submitList(@Nullable List<QuestionItem> items) {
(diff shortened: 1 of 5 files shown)


widgetssdk/src/main/java/com/glia/widgets/survey/SurveyAdapter.java
```diff
@@ -56,8 +56,9 @@ interface SurveyAdapterListener {
     private final List<QuestionItem> questionItems = new ArrayList<>();
     private final SurveyAdapterListener listener;
 
-    public SurveyAdapter(SurveyAdapterListener listener) {
+    public SurveyAdapter(SurveyStyle initialStyle, SurveyAdapterListener listener) {
         this.listener = listener;
+        this.style = initialStyle;
```
AndriiHorishniiMOC, line 61: I guess we don't need `public void setStyle(SurveyStyle style)` in the `SurveyAdapter` class.

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/survey/SurveyAdapter.java
andrews-moc: As far as I  remember we need it. Because configured style is set via setStyle() function. I guess the initial style added by Davit to the constructor of SurveyAdapter contains the default setup. Please correct me if I'm wrong, @DavDo.

DavDo: Yep, @andrews-moc rightly observed, I just added this to avoid null checks before using it.

AndriiHorishniiMOC: Got it.

DavDo: Agree! It is bit confusing to have 2 methods for setting style(constructor and setStyle method), but in fact one of them is default empty style.

andrews-moc: @DavDo Could you please add a comment to the constructor explaining the default style setting to the code?

gugalo (approved):

DavDo:
!squash

DavDo:
!merge

PR #375, opened 2022-10-13
Add post release workflow

gersonnoboa:
Because our android-sdk-widgets release depends on Maven Central, then we cannot increase the android-sdk-widgets dependency in android_cortex_banking_app until Maven Central reflects our release. This step thus needs to be done manually. The tag to be sent is the same as the released version.

I also had to re-add the spaces that were deleted in a previous PR because this absolutely screws up the formatting inside Bitrise and makes all steps with scripts to fail because commands show up overlapped.

MOB-1589

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -1,0 +1,16 @@
+name: Post Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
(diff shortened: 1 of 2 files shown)


DavDo (approved):

andrews-moc (approved):

gersonnoboa:
!merge

PR #377, opened 2022-10-17
Allow version updater to receive update type

gersonnoboa:
In order to make it easier to update the project version from Bitrise, the version updater, instead of receiving the specific version name and code, receives a `type` parameter. This way, we can make a GitHub Action to which you can send `patch`, `minor`, or `major`.


gugalo (approved):

gersonnoboa:
!merge

PR #379, opened 2022-10-18
Increment Core SDK version to 0.26.0

sm-deployer:


gersonnoboa (approved):

andrews-moc:
@gersonnoboa This PR is only for SDK version incrementation, do I understand correctly?
I'm asking because in case of release preparation we need to increment `versionCode` and `versionName`.

gersonnoboa:
@andrews-moc correct, only for Core SDK version increment so that it is synced with the latest release. No android-sdk-widgets is planned at the moment.

andrews-moc (approved):

gersonnoboa:
!merge

PR #382, opened 2022-10-20
Increment project version to 1.10.0

sm-deployer:


andrews-moc (approved):

gersonnoboa:
I will close this an manually open one with 1.9.1 because we decided to do patch releases instead

PR #384, opened 2022-10-20
Add GitHub Action for manual version increment

gersonnoboa:
Sometimes we will need to increment the minor version. However, right now the `_increment_project_version` will increment with type set as `patch` by default. This workflow lets you choose on a dropdown what type you want to use for the version increment. Bitrise then executes an authenticated version of incrementing project version. This is needed because _increment_project_version on its own cannot work, as it needs the SSH key.

### DIFF
--- a/.github/workflows/increment-project-version.yml
+++ b/.github/workflows/increment-project-version.yml
@@ -1,0 +1,21 @@
+name: Increment project version
+
+on:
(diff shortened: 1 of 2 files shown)


andrews-moc (approved):

gersonnoboa:
!merge

PR #390, opened 2022-11-07
Fix permission requests on upgrading to the one-way video

AndriiHorishniiMOC:
Prevent permission requests when upgrading to the one-way video

[MOB-1459](https://glia.atlassian.net/browse/MOB-1459)

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
@@ -143,12 +143,14 @@ public class CallActivity extends AppCompatActivity {
 
     private void startCallWithPermissions() {
         List<String> missingPermissions = new ArrayList<>();
-        if (configuration.getMediaType() == Engagement.MediaType.VIDEO && missingPermission(Manifest.permission.CAMERA)) {
-            missingPermissions.add(Manifest.permission.CAMERA);
-        }
-        if ((configuration.getMediaType() == Engagement.MediaType.VIDEO || configuration.getMediaType() == Engagement.MediaType.AUDIO)
-                && missingPermission(Manifest.permission.RECORD_AUDIO)) {
-            missingPermissions.add(Manifest.permission.RECORD_AUDIO);
+        if (!configuration.getIsUpgradeToCall()) { // In upgrading to call, permissions handle on the Core SDK side.
+            if (configuration.getMediaType() == Engagement.MediaType.VIDEO && missingPermission(Manifest.permission.CAMERA)) {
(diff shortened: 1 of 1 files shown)


DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
```diff
@@ -143,12 +143,14 @@ public void onRequestPermissionsResult(int requestCode, @NonNull String[] permis
...
-                && missingPermission(Manifest.permission.RECORD_AUDIO)) {
-            missingPermissions.add(Manifest.permission.RECORD_AUDIO);
+        if (!configuration.getIsUpgradeToCall()) {
+            if (configuration.getMediaType() == Engagement.MediaType.VIDEO && missingPermission(Manifest.permission.CAMERA)) {
+                missingPermissions.add(Manifest.permission.CAMERA);
+            }
+            if ((configuration.getMediaType() == Engagement.MediaType.VIDEO || configuration.getMediaType() == Engagement.MediaType.AUDIO)
+                    && missingPermission(Manifest.permission.RECORD_AUDIO)) {
+                missingPermissions.add(Manifest.permission.RECORD_AUDIO);
+            }
```
andrews-moc, line 153: Looking at this code, I would expect that:
 `if need to upgrade to call` => then `need to ask permissions`.

But it's strange to me that it's vice-a-versa: `!configuration.getIsUpgradeToCall()`. Could you please explain the logic?

AndriiHorishniiMOC: Core SDK can request permissions by itself. So upgrading to call handles on the Core SDK side.
As far as I understand, those permission requests from here are a way to prepare all needed for the call screen before making a deal with the Core SDK.

andrews-moc: Ok, maybe it makes sense to add some comment here.

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
gugalo: > As far as I understand, those permission requests from here are a way to prepare all needed for the call screen before making a deal with the Core SDK.  

Yes. There was some scenario in Widget SDK that caused an expected behavior. But I don't remember why and can't find the bug ticket

AndriiHorishniiMOC:
!squash

AndriiHorishniiMOC:
!merge

PR #391, opened 2022-11-09
Add push notifications support

gugalo:
Add push notifications support

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -38,25 +38,35 @@ android {
         ext["GLIA_API_KEY_SECRET"] = System.getenv("GLIA_API_KEY_SECRET") as String
         ext["GLIA_API_KEY_ID"] = System.getenv("GLIA_API_KEY_ID") as String
         ext["GLIA_SITE_ID"] = System.getenv("GLIA_SITE_ID") as String
+        ext["FIREBASE_PROJECT_ID"] = System.getenv("FIREBASE_PROJECT_ID") as String
+        ext["FIREBASE_API_KEY"] = System.getenv("FIREBASE_API_KEY") as String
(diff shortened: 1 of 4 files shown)


DavDo (approved):

andrews-moc (approved):

app/google-services.json
```diff
@@ -0,0 +1,187 @@
+{
```
AndriiHorishniiMOC, line 1: Is it okay to publish this file in the public repo?

gugalo: Hmm... Good question.
Will need to investigate it

gugalo: Though I already made it public by committing this changes.... 🤔

gugalo: Have checked a few questions in StackOverflow regarding this. And the general idea is that it is not a real security breach. But does expose some endpoints. So it is not recommended to have it fully public. One of the replies from Google support:
> The general answer is yes, the google-services.json is safe to check into your repo and is something that should be shared among engineers on your team.  The JSON file does not contain any super-sensitive information (like a server API key).  It does contain some information like your database URL, Android API key, and storage bucket.  These are not secrets, but if your security rules are not set up correctly an attacker could use them against you.  However since these values are also compiled into your APK as resources, the theoretical attacker would not need your JSON file to get these values anyway.

gugalo: I will bring this topic tomorrow on standup

gugalo: Change the way the Firebase is initialised

gugalo:
Made changes to firebase initialisation

AndriiHorishniiMOC (approved):

DavDo (approved):

gugalo:
!merge

PR #402, opened 2022-11-27
Call Screen. Implement Remote configuration for remaining widgets

DavDo:
- Implement Remote `Configuration` for remaining widgets on `Call` screen
- Update `MessageBalloonRemoteConfig.kt`
- Add Custom view to ease theme management for different states
- Add `CallState.isOnlyTimeChanged` parameter to avoid entire view re-rendering on timer update

MOB-1649

**Jira issue:**
https://glia.atlassian.net/browse/MOB-1649

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
@@ -22,6 +22,8 @@ class CallState {
     public final Engagement.MediaType requestedMediaType;
     public final boolean isSpeakerOn;
     public final boolean isOnHold;
+    //    Need this to not update all views when only time is changed.
+    public final boolean isOnlyTimeChanged;
 
     @NonNull
     @Override
@@ -56,14 +58,15 @@ class CallState {
                 Objects.equals(companyName, callState.companyName) &&
                 Objects.equals(requestedMediaType, callState.requestedMediaType) &&
                 isSpeakerOn == callState.isSpeakerOn &&
-                isOnHold == callState.isOnHold;
+                isOnHold == callState.isOnHold &&
+                isOnlyTimeChanged == callState.isOnlyTimeChanged;
     }
 
     @Override
     public int hashCode() {
         return Objects.hash(integratorCallStarted, isVisible, messagesNotSeen,
                 callStatus, landscapeLayoutControlsVisible, isMuted, hasVideo,
-                companyName, requestedMediaType, isSpeakerOn, isOnHold);
+                companyName, requestedMediaType, isSpeakerOn, isOnHold, isOnlyTimeChanged);
     }
 
     public boolean showOperatorStatusView() {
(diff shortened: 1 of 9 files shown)


widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
```diff
@@ -22,6 +22,8 @@ class CallState {
     public final Engagement.MediaType requestedMediaType;
     public final boolean isSpeakerOn;
     public final boolean isOnHold;
+    //    Need this to not update all views when only time is changed.
+    public final boolean isOnlyTimeChanged;
```
gugalo, line 26: It might be just me but I didn't understand this nither from the comment or the code implementation.

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
AndriiHorishniiMOC: I guess it is for the count-up timer.

widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
```diff
@@ -395,6 +402,8 @@ public static class Builder {
         private Engagement.MediaType requestedMediaType;
         private boolean isSpeakerOn;
         private boolean isOnHold;
+        //My be helpful when converting to Kotlin, as ide makes field nullable.
```
andrews-moc, line 405: I think there are some typos here:
My => May
ide => not sure what does it mean.

AndriiHorishniiMOC (approved):

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
DavDo: Every screen in our `Widgets SDK` has its own `state`(for `call screen` it is `CallState`), and based on that state screen is rendering widgets(`CallView.emitState(CallState)`). On `Call Screen`, we have a call timer and to update time on it, we are updating CallState every second, which is causing an entire screen update instead of updating only the timer. So I added this field which is indicating that only time has changed, and no need to update the remaining widgets' state.
If you have a suggestion on how to write a more clear comment I will change it

widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
DavDo: fixed https://github.com/salemove/android-sdk-widgets/pull/402/commits/23f80355fe595bd0b69fa31ece8b37f85c3486f3

DavDo:
!squash

DavDo:
!squash

DavDo:
!merge

DavDo:
!merge

widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
andrews-moc: @DavDo Ah, ide is Android Studio. 🙂 
If you wrote with capital letters (IDE) I would guess it. 🙂

PR #403, opened 2022-12-01
Fix incorrect changelog in release process

gersonnoboa:
The changelog was incorrectly generated before the new tag was created, so the changelog showed old commits. Two new steps have been added: one generates a new tag according to the value saved in $NEW_VERSION on workflow trigger, and another after it fetches all tags. This makes the changelog generation step aware of the new tag and generates it correctly.

MOB-1680


andrews-moc (approved):

gersonnoboa:
!merge

PR #404, opened 2022-12-01
Mob 1663 alerts implement remote configuration for alerts

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-1663

**Additional info:**
- Implement Remote Configuration for Alerts
- Moved applying the style from ChatHeadView.kt to ChatHeadService.kt, to be sure that style will work for all views inflated in Service
- Create BadgeTextView.kt custom class for Badges with default params and exact round shaped background
- Minor code cleanup

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -80,6 +80,7 @@ android {
                 'WrongViewIdFormat',
                 'HardcodedText'
     }
+    namespace 'com.glia.exampleapp'
 }
 
 dependencies {
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.glia.exampleapp">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
 
--- a/app/src/main/res/raw/remote_config.json
+++ b/app/src/main/res/raw/remote_config.json
@@ -1,5 +1,5 @@
 {
-  "alertScreen": {
+  "alert": {
     "backgroundColor": {
       "type": "fill",
       "value": [
@@ -141,7 +141,7 @@
       }
     },
     "title": {
-      "alignment": "leading",
+      "alignment": "center",
       "background": {
         "type": "fill",
         "value": [
@@ -162,7 +162,7 @@
     "titleImageColor": {
       "type": "fill",
       "value": [
-        "#FF0000"
+        "#04728c"
       ]
     }
   },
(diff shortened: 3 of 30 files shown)


widgetssdk/src/main/java/com/glia/widgets/core/chathead/ChatHeadService.kt
```diff
@@ -101,17 +111,14 @@ class ChatHeadService : Service() {
...
+        val display = displaySize
+        params.x = chatHeadPosition.first ?: getDefaultXPosition(display.width)
+        params.y = chatHeadPosition.second ?: getDefaultYPosition(display.height)
     }
 
-    private fun notNullOrDefault(item: Int?, defaultItem: Int): Int = item ?: defaultItem
-
     private fun getDefaultXPosition(screenWidth: Int): Int = screenWidth - chatHeadSize - chatHeadMargin
 
+    //🤔 Some indescribable magic here 🤦‍
```
andrews-moc, line 121: Let me double-check because it's a public repository. Have you added this comment intentionally? I don't mind but just want to make sure.

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/theme/alert/AlertTheme.kt
```diff
@@ -8,9 +8,9 @@ internal data class AlertTheme(
     val title: TextTheme?,
     val titleImageColor: ColorTheme?,
     val message: TextTheme?,
-    val backgroundColor: ColorTheme?,
+    val backgroundColor: ColorTheme?,//
     val closeButtonColor: ColorTheme?,
     val positiveButton: ButtonTheme?,
     val negativeButton: ButtonTheme?,
-    val isVerticalAxis: Boolean?
+    val isVerticalAxis: Boolean?//
```
andrews-moc, line 15: I guess, `//` can be removed here, right?

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/core/chathead/ChatHeadService.kt
DavDo: Yes, because I haven't able to understand or even guess what is it, why `10*8`.
But if it's not ok I can remove.

andrews-moc: I think it's not 10*8 but 8/10 of the entire height. Or, if simplify, it's 4/5 of the entire height. But why exactly 4/5 - don't know. Maybe just a strict position.

DavDo: It seems it's better to leave this method without any comment)

DavDo:
!squash

AndriiHorishniiMOC (approved):

PR #407, opened 2022-12-06
Added Message Center screen

AndriiHorishniiMOC:
Added new screen for secure messaging.

[MOB-1679](https://glia.atlassian.net/browse/MOB-1679)


DavDo (approved):
Good work! I've nothing to add 👍

app/src/main/res/values/strings.xml
```diff
@@ -71,6 +71,7 @@
     <string name="main_launch_fragment">Launch Chat in Fragment</string>
     <string name="main_start_in_integrator_app">Start Chat in integrator app</string>
     <string name="main_start_chat_flow">Start new chat flow</string>
+    <string name="main_start_message_center_flow">Start new chat flow</string>
```
andrews-moc, line 74: To my mind, this name may be confusing with the current chat flow.
Maybe "Start new messaging flow" or any other name?

widgetssdk/build.gradle
```diff
@@ -1,12 +1,12 @@
...
 
 android {
-    compileSdkVersion 31
+    compileSdkVersion 33
     buildToolsVersion "30.0.3"
     defaultPublishConfig "debug"
     defaultConfig {
         minSdkVersion 24
-        targetSdkVersion 31
+        targetSdkVersion 33
```
andrews-moc, line 9: I remember one of our integrators faced with some issues when we upgraded `targetSdkVersion`. @gugalo was working on that request. But I don't remember the details.
@gugalo, could you please remind me what was the issue and if `targetSdkVersion` won't cause any issues for our integrators?

widgetssdk/src/main/java/com/glia/widgets/messagecenter/MessageCenterController.kt
```diff
@@ -0,0 +1,27 @@
...
+        this.view = view
+    }
+
+    override fun onCheckMessagesClicked() {
+        view?.navigateToChat()
+    }
+
+    override fun onSendMessageClicked() {
+        view?.navigateToChat()
+    }
```
andrews-moc, line 16: According to Figma design, it should navigate to chat, indeed.
However, App Bar title should be 'Messaging' (not 'Chat')

app/src/main/res/values/strings.xml
AndriiHorishniiMOC: done

widgetssdk/build.gradle
AndriiHorishniiMOC: It shouldn't be an issue.

widgetssdk/src/main/java/com/glia/widgets/messagecenter/MessageCenterController.kt
AndriiHorishniiMOC: fixed

andrews-moc (approved):

widgetssdk/build.gradle
gugalo: It is not an issue on its own however I still would do that especially since you have caused the same issue as last time the integrator complained about this.

If we just upgrade the `compileSdkVersion` it is a good thing and won't cause any issues to the integrators. But the problem comes because of easy-to-miss changes in dependencies versions. Android dependencies now can require a minimum `compileSdkVersion` using a `minCompileSdk` property. And you just introduced in your changes `androidx.core:core-ktx:1.9.0` that requires compileSdkVersion 33+.

For example, this is an error that the integrator would get if they use Widgets from your PR with compiled SDK of a lower version:
```
One or more issues found when checking AAR metadata values:

Dependency 'androidx.core:core:1.9.0' requires 'compileSdkVersion' to be set to 33 or higher.
Compilation target for module ':app' is 'android-31'

Dependency 'androidx.core:core-ktx:1.9.0' requires 'compileSdkVersion' to be set to 33 or higher.
Compilation target for module ':app' is 'android-31'

Dependency 'androidx.annotation:annotation-experimental:1.3.0' requires 'compileSdkVersion' to be set to 33 or higher.
Compilation target for module ':app' is 'android-31'
```

You could argue that integrators can manually force override dependencies version in their build.gradle. And that is exactly what I shared with the client last time. But that requires additional effort and I don't remember why they had problems doing that even after I shared clear interactions with them. 

For reference here is the link to the old issue 
https://glia.atlassian.net/browse/MOB-1235
Slack discussion: https://salemove.slack.com/archives/CHN6UB7UH/p1650279001120289

widgetssdk/build.gradle
```diff
@@ -37,6 +37,10 @@ android {
                 'WrongViewIdFormat',
                 'HardcodedText'
     }
+
+    buildFeatures {
+        viewBinding true
+    }
```
gugalo, line 43: Would this require the integrators to also add this to their project?

widgetssdk/build.gradle
gugalo: We might want to discuss and see how we handle this in the future. @AndriiHorishniiMOC @andrews-moc @DavDo @mikktammeglia.

gugalo: For now, though the simplest solution would be to downgrade the `androidx.core` to 1.8.0 which has `minCompileSdk=31`.
Same for @DavDo PR

AndriiHorishniiMOC: @gugalo Thank you for the explanation. According to the Glia Android Core SDK, we should decrease the value to 32.
https://github.com/salemove/android-sdk/blob/master/androidSdk/build.gradle#L11

gugalo: That's actually strange because we should not have a higher compile SDK the target SDK requirements for Google Play Store upload which currently is API 31 until November 1, 2023 when it becomes API 33 (skipping the 32 version).

But that's ideally not sure if any of this is a real issue so I think we might just make some agreement that we as a team thinks is reasonable and just follow whatever we agreed upon.

gugalo: Might also incode someone related to the client engeneering in the discussion.

gugalo: @mikktammeglia can I ask you to organize the meeting since I am not sure who are the parties that should be invited to the discussion

widgetssdk/build.gradle
DavDo: If I'm not mistaken no, it generates binding classes for views at compile time

widgetssdk/build.gradle
mikktammeglia: Hei,

Discussed it a bit and we can have a short round of discussion after start-up tomorrow. If I understand things correctly:
1. It is easier for our clients if we use the minimum. (31 for now).
2. We can change this policy if there are specific gains we can say from having a higher version. 
3. If it makes sense then tomorrow in the engineering initiatives discussion it could be added as one more initiative to make it work with the highest one. (i understand we could add some checks to it)

But if I misunderstood anything I am happy to be shown the value of going with highest in as precise way as possible. :)

widgetssdk/build.gradle
AndriiHorishniiMOC: I did an investigation. It doesn't require the integrators to use the view binding in their projects.

widgetssdk/build.gradle
AndriiHorishniiMOC: @gugalo Sorry for the misunderstanding. The Core SDK has `targetSdkVersion` 32, not `minSdkVersion`.

gugalo (approved):

DavDo (approved):

AndriiHorishniiMOC:
!merge

PR #411, opened 2022-12-08
Resolve breaking changes related to Core SDK update

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MUIC-1612

**Additional info:**
These changes depend on still unreleased changes in Core SDK. Before merging need to wait for the Core SDK release.
Core SDK has breaking changes related to local chat message storage. There are slight changes to the behavior of 2 Core SDK methods:
- Chat.sendResponse(...)
- Glia.getChatHistory(...)

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
@@ -52,6 +52,7 @@ import com.glia.widgets.chat.model.history.OperatorAttachmentItem;
 import com.glia.widgets.chat.model.history.OperatorChatItem;
 import com.glia.widgets.chat.model.history.OperatorMessageItem;
 import com.glia.widgets.chat.model.history.OperatorStatusItem;
+import com.glia.widgets.chat.model.history.LinkedChatItem;
 import com.glia.widgets.chat.model.history.VisitorAttachmentItem;
 import com.glia.widgets.chat.model.history.VisitorMessageItem;
 import com.glia.widgets.core.dialog.DialogController;
@@ -93,6 +94,7 @@ import com.glia.widgets.view.MinimizeHandler;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Observable;
@@ -459,6 +461,10 @@ public class ChatController implements
 
     private void onMessage(@NonNull ChatMessageInternal messageInternal) {
         ChatMessage message = messageInternal.getChatMessage();
+        if (!isNewMessage(chatState.chatItems, message)) {
+            return;
+        }
+
         boolean isUnsentMessage = !chatState.unsentMessages.isEmpty() && chatState.unsentMessages.get(0).getMessage().equals(message.getContent());
         Logger.d(TAG, "onMessage: " + message.getContent() + ", id: " + message.getId() + ", isUnsentMessage: " + isUnsentMessage);
         if (isUnsentMessage) {
@@ -469,7 +475,7 @@ public class ChatController implements
             List<ChatItem> currentChatItems = new ArrayList<>(chatState.chatItems);
             int currentMessageIndex = currentChatItems.indexOf(currentMessage);
             currentChatItems.remove(currentMessage);
-            currentChatItems.add(currentMessageIndex, new VisitorMessageItem(message.getId(), false, message.getContent()));
+            currentChatItems.add(currentMessageIndex, VisitorMessageItem.asNewMessage(message));
 
             // emitting state because no need to change recyclerview items here
             emitViewState(chatState
@@ -518,7 +524,7 @@ public class ChatController implements
     private void appendUnsentMessage(String message) {
         Logger.d(TAG, "appendUnsentMessage: " + message);
         List<VisitorMessageItem> unsentMessages = new ArrayList<>(chatState.unsentMessages);
-        VisitorMessageItem unsentItem = new VisitorMessageItem(VisitorMessageItem.UNSENT_MESSAGE_ID, false, message);
+        VisitorMessageItem unsentItem = VisitorMessageItem.asUnsentItem(message);
         unsentMessages.add(unsentItem);
 
         List<ChatItem> currentChatItems = new ArrayList<>(chatState.chatItems);
@@ -890,7 +896,7 @@ public class ChatController implements
         }
 
         if (message.getContent() != null && !message.getContent().isEmpty()) {
-            currentChatItems.add(new VisitorMessageItem(VisitorMessageItem.HISTORY_ID, false, message.getContent()));
+            currentChatItems.add(VisitorMessageItem.asHistoryItem(message));
         }
     }
 
@@ -937,14 +943,8 @@ public class ChatController implements
             FilesAttachment filesAttachment = (FilesAttachment) attachment;
             AttachmentFile[] files = filesAttachment.getFiles();
             for (AttachmentFile file : files) {
-                int type;
-                if (file.getContentType().startsWith("image")) {
-                    type = ChatAdapter.VISITOR_IMAGE_VIEW_TYPE;
-                } else {
-                    type = ChatAdapter.VISITOR_FILE_VIEW_TYPE;
-                }
                 currentChatItems.add(
-                        new VisitorAttachmentItem(chatMessage.getId(), type, file, false, false, false)
+                        VisitorAttachmentItem.fromAttachmentFile(chatMessage.getId(), chatMessage.getTimestamp(), file)
                 );
             }
         }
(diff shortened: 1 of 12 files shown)


DavDo (approved):

build.gradle
```diff
@@ -20,6 +20,8 @@ allprojects {
         flatDir {
             dirs '../libs'
         }
+        // Used in case when testing local Core SDK snapshot releases
+//        maven { url '/Users/aleksandr.chatsky/Projects/glia-android-sdk/build/maven_repo' }
```
andrews-moc, line 24: I think we shouldn't release it to the public repository because it's your local path and it won't work for others, isn't it?

widgetssdk/src/main/java/com/glia/widgets/di/GliaCoreImpl.java
```diff
@@ -144,7 +144,7 @@ public void unsubscribeFromQueueUpdates(Consumer<GliaException> onError,
     @Override
     public void subscribeToQueueTicketUpdates(@NonNull String ticketId,
                                               RequestCallback<QueueTicket> callback) {
-        Glia.subscribeToQueueTicketUpdates(ticketId, callback);
+//        Glia.subscribeToQueueTicketUpdates(ticketId, callback);
```
andrews-moc, line 147: Could you please explain the reason for disabling this?

widgetssdk/src/main/java/com/glia/widgets/core/queue/GliaQueueRepository.java
```diff
@@ -150,9 +150,9 @@ public Completable observeQueueTicketStateChangeToUnstaffed() {
     @NonNull
     public RequestCallback<QueueTicket> emitOnQueueTicketChangeToUnstaffed(CompletableEmitter emitter) {
         return (queueTicket, e) -> {
-            if (queueTicket.getState() == QueueTicket.State.UNSTAFFED) {
-                emitter.onComplete();
-            }
+//            if (queueTicket.getState() == QueueTicket.State.UNSTAFFED) {
+//                emitter.onComplete();
+//            }
```
andrews-moc, line 155: Could you please explain the reason for disabling this?

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
```diff
@@ -1476,6 +1509,9 @@ public void notificationsDialogDismissed() {
 
     public void queueForEngagementStarted() {
         Logger.d(TAG, "queueForEngagementStarted");
+        if (chatState.isOperatorOnline()) {
+            return;
+        }
```
andrews-moc, line 1514: Could you please explain why we should not execute the queuing related code if an operator is online? Cannot get the point here.

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
```diff
@@ -826,15 +827,15 @@ private void operatorConnected(String operatorName, String profileImgUrl) {
...
                     )
             );
         } else {
             items.add(
                     OperatorStatusItem.OperatorFoundStatusItem(
-                            chatState.companyName,
+                            "operatorConnected2",
                             Utils.formatOperatorName(operatorName),
                             profileImgUrl
                     )
```
andrews-moc, line 841: Let me double-check, do you really want to replace `companyName` with constant values `operatorConnected1` and `operatorConnected2`?

widgetssdk/src/main/java/com/glia/widgets/chat/model/history/OperatorChatItem.java
```diff
@@ -2,14 +2,14 @@
 
 import java.util.Objects;
 
-public abstract class OperatorChatItem extends ChatItem {
+public abstract class OperatorChatItem extends ParticipantMessageChatItem {
```
andrews-moc, line 5: This solution is good. However, I have two concerns here. Let me describe them and there is a chance that we could mitigate them.
1. `ParticipantMessageChatItem` name itself doesn't correspond to the underlying idea - "object with message id". Seems like the word "participant"  has nothing to do with message identifier. Maybe we can find a better name like `ChatItemWithId` or `IdentifiedChatItem`, `LinkedChatItem`, `ChatItemWithPayload`, etc.
2. What do you think regarding adding message id to all types of chat items directly (without introducing a new superclass)? What benefit can this idea bring - at first, some ChattItems already contain message id and secondly avoiding new nesting level without a strong may reduce complexity.

But I am not insisting on these suggestions. Just think if they make sense from your point of view.

build.gradle
gugalo: Fixed in cf59e63

widgetssdk/src/main/java/com/glia/widgets/di/GliaCoreImpl.java
gugalo: Left over from debuging. Fixed ccca28d

widgetssdk/src/test/java/com/glia/widgets/core/queue/GliaQueueRepositoryTest.java
```diff
@@ -1,133 +0,0 @@
...
-import com.glia.widgets.core.queue.model.GliaQueueingState;
-import com.glia.widgets.di.GliaCore;
-
-import org.junit.After;
-import org.junit.Test;
-
-import io.reactivex.CompletableEmitter;
-import io.reactivex.disposables.CompositeDisposable;
-
-public class GliaQueueRepositoryTest {
```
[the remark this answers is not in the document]
gugalo: Left over from debuging. Fixed https://github.com/salemove/android-sdk-widgets/commit/ccca28d0c62f73021106e9e0bd8a900495ccfcb3

widgetssdk/src/main/java/com/glia/widgets/core/queue/GliaQueueRepository.java
gugalo: https://github.com/salemove/android-sdk-widgets/commit/ccca28d0c62f73021106e9e0bd8a900495ccfcb3

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
gugalo: Because if the operator is online then that means that engagement has been already established and so no need to enqueue visitor for a new engagement.

gugalo: Unless I am missing something

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
gugalo: Left over from debuging. Fixed https://github.com/salemove/android-sdk-widgets/commit/ccca28d0c62f73021106e9e0bd8a900495ccfcb3

widgetssdk/src/main/java/com/glia/widgets/chat/model/history/OperatorChatItem.java
gugalo: 1. Went with LinkedChatItem
2. Because not all times of ChatItems need the ID. For example OperatorStatusItem or the MediaUpgradeStartedTimerItem. And ChatItem id is not the same as the message ID that comes from the Glia Core SDK. ChatItem ID is an internal implementation for Widget SDK and is sometimes linked to Glia Core message ID and sometimes not. Ideally, it should be converted to 2 separate fields but my PR is already pretty big as is

gugalo:
Found one more bug related to message order so had to add a timestamp to some ChatItems so that I could sort based on timestamp

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
andrews-moc: Hm, I don't know about this.
I would try to use `GliaEngagementStateRepository.hasOngoingEngagement()` function for this instead. However, if you are sure, then let's leave your approach.

DavDo (approved):

andrews-moc (approved):

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
```diff
@@ -1378,22 +1384,45 @@ private void loadChatHistory() {
...
     }
 
+    private List<ChatMessageInternal> removeDuplicates(List<ChatItem> oldHistory, List<ChatMessageInternal> newHistory) {
+        if (newHistory == null || newHistory.isEmpty()) {
+            return newHistory;
+        }
+        if (oldHistory == null || oldHistory.isEmpty()) {
+            return newHistory;
+        }
+        return newHistory.stream().filter(newMessage -> oldHistory
```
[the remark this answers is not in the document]
DavDo: Sorry, didn’t notice it.

gugalo:
!merge

PR #423, opened 2023-01-05
Response card improvements

DavDo:
MOB-1566

**Jira issue:**
https://glia.atlassian.net/browse/MOB-1566

**Additional info:**

Response cards' behavior on mobile will be unified with the behavior on the web as follows:
- Upon clicking on the card, the card disappears and is replaced with the visitor's response as a text 
- Response cards are not shown in the chat history

I also fixed the issue - when the chat history's last item is the `response card`, input is not disabled and improved `isNewMessage(newMessage: ChatMessage)` method.


AndriiHorishniiMOC (approved):
Wow! Now we have Chat Controller on Kotlin.

gugalo (approved):

DavDo:
!merge

PR #429, opened 2023-01-16
Fix nullable operator image issue

gugalo:
After migration to Kotlin, the support for the operator picture being nullable was lost in one place. Because of that in case of the operator not having an image (as in acceptance tests) the chat state does not get updated on the engagement start


DavDo (approved):

AndriiHorishniiMOC (approved):

gugalo:
!merge

PR #431, opened 2023-01-18
Improve Custom Response Cards detection

gugalo:
In case a response card with an empty JSON in 'metadata' (e.g: { "metadata": {} } ) is sent by the operator it would be classified as a Custom Response Card even though it has no data in it. And as result would not display any 'options' only the message text and raw metadata as empty brackets

**Jira issue:**
No ticket.

**Additional info:**

**Screenshots:**
Was:

image

Became:

image


DavDo (approved):

gugalo:
why my images do not load?

AndriiHorishniiMOC (approved):

gugalo:
!squash

gugalo:
!merge

PR #436, opened 2023-01-31
Fix unknown property error when syncing Gradle

gugalo:
When opening a project for the first time Android Gradle might create a default 'local.properties' file. Gradle will attempt to read multiple properties once it detects that this file exists but will fail to find them later.

**P.S:** I have failed to extract this VERY repetitive logic to separate methods but it does its thing. If you know how to do that feel free to add to this solution.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -35,23 +35,24 @@ android {
         Properties p = new Properties()
         new FileInputStream(propsFile).withCloseable { is -> p.load(is) }
         p.each { name, value -> ext[name] = value }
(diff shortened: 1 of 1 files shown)


DavDo (approved):

gugalo:
Nice!!
Thank you @andrews-moc! 👍 🙇

andrews-moc (approved):

gugalo:
squash!

gugalo:
!merge

PR #437, opened 2023-02-01
Send secure attachments from the chat transcript screen

AndriiHorishniiMOC:
Send secure attachments from the chat transcript screen

[MOB-1743](https://glia.atlassian.net/browse/MOB-1743)

[MOB-1743]: https://glia.atlassian.net/browse/MOB-1743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
```diff
@@ -361,6 +365,12 @@ public void initChat(String companyName,
         createNewTimerCallback();
         callTimer.addFormattedValueListener(timerStatusListener);
         updateAllowFileSendState();
+
+        // TODO: enable chat panel with send message support
+        // will implement in the next task
+        if (isSecureEngagementUseCase.invoke()) {
+            emitViewState(chatState.enableChatPanel());
+        }
```
andrews-moc, line 373: To double-check, is this TODO still relevant? I'm asking because it looks like `enableChatPanel()` does what TODO states.

widgetssdk/src/main/java/com/glia/widgets/di/RepositoryFactory.java
```diff
@@ -99,7 +99,9 @@ public GliaQueueRepository getGliaQueueRepository() {
 
     public FileAttachmentRepository getGliaFileAttachmentRepository() {
         if (fileAttachmentRepository == null) {
-            fileAttachmentRepository = new FileAttachmentRepository(gliaCore);
+            fileAttachmentRepository = new FileAttachmentRepository(gliaCore,
+                    getGliaEngagementTypeRepository(),
+                    () -> new GliaCoreImpl().getSecureConversations());
```
DavDo, line 104: `new GliaCoreImpl().getSecureConversations()` can be changed to `gliaCore.getSecureConversations()`, and also I don't understand why was lambda used instead of passing direct instance.

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/messagecenter/MessageCenterController.kt
```diff
@@ -76,6 +77,7 @@ class MessageCenterController(
     @VisibleForTesting
     fun handleSendMessageResult(gliaException: GliaException?) {
         if (gliaException == null) {
+            setSecureEngagementUseCase(true)
```
andrews-moc, line 80: Sorry, I don't understand the logic here. Each time any message-sending result is handled and gliaException is null, we set `SecureEngagementUseCase` to `true`?

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
```diff
@@ -361,6 +365,12 @@ public void initChat(String companyName,
         createNewTimerCallback();
         callTimer.addFormattedValueListener(timerStatusListener);
         updateAllowFileSendState();
+
+        // TODO: enable chat panel with send message support
+        // will implement in the next task
+        if (isSecureEngagementUseCase.invoke()) {
```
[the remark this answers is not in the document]
andrews-moc: I also thought about it and as far as I understand, it is only possible for Kotlin classes. So, need to migrate ChatController to Kotlin. 🙂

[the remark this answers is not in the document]
DavDo: Yeah( My mistake, I didn't pay attention that this class is still in Java, this is the most dangerous file to convert to Kotlin

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.java
AndriiHorishniiMOC: Yes, it just show add attachment button. Needs to change.

widgetssdk/src/main/java/com/glia/widgets/di/RepositoryFactory.java
AndriiHorishniiMOC: The lambda uses because we can't get secure messaging before the SDK is not initialized.

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/messagecenter/MessageCenterController.kt
AndriiHorishniiMOC: Why each time? Only when we success sends the message from the Welcome to Message Center screen.

andrews-moc: Ah, I forgot that `MessageCenterController` is related only to secure messaging. I mistakenly thought that `MessageCenterController` refers to all messages and that's why misunderstood the logic.

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/di/RepositoryFactory.java
DavDo: What do you think? maybe it makes sense to move the initialization check from `getSecureConversations()` to methods inside `SecureConversations`, I myself removed a couple of that checks because we had that check when getting instances, but It seems this check is not mandatory at this step and making obstacles for integrators
cc: @andrews-moc @gugalo

AndriiHorishniiMOC:
!squash

widgetssdk/src/main/java/com/glia/widgets/di/RepositoryFactory.java
andrews-moc: Honestly, I don't understand why and how can `getSecureConversations()` be called before SDK is initialized. Maybe I am missing anything, could you please explain?

P.S. Andrii has just merged the PR.

DavDo: That's because Dependencies in Widgets SDK initializing in the `onAppCreate()` method, before Glia.initialize() call, and we need to pass SecureConversations instance to Repository constructor

andrews-moc: Your suggestion makes sense to me then.

DavDo: Okay, I will rise this question in the channel once again.

PR #446, opened 2023-02-13
Fixed the bug with duplicate messages

AndriiHorishniiMOC:
Fixed the bug with duplicate messages.

There was a bug in the chat screen, after the engagement is started the history messages were duplicated. It happened because the filter had worked incorrectly.

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
@@ -1,6 +1,7 @@
 package com.glia.widgets.chat.controller
 
 import android.net.Uri
+import androidx.annotation.VisibleForTesting
 import androidx.recyclerview.widget.RecyclerView
 import com.glia.androidsdk.GliaException
 import com.glia.androidsdk.Operator
@@ -276,7 +277,7 @@ class ChatController(
 
     private fun onMessage(messageInternal: ChatMessageInternal) {
         val message = messageInternal.chatMessage
-        if (!isNewMessage(message)) {
+        if (!isNewMessage(chatState.chatItems, message)) {
             return
         }
         val isUnsentMessage =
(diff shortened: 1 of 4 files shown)


andrews-moc:
@AndriiHorishniiMOC I think it might be good to have unit tests for this logic because it's rather complicated. Does it make sense to you?

Pelkar:
> @AndriiHorishniiMOC I think it might be good to have unit tests for this logic because it's rather complicated. Does it make sense to you?

One way to simplify the logic would be to use a LinkedHashSet. It would automatically remove duplicates and keep the order. However that would require a bit of refactoring

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -1262,13 +1262,17 @@ class ChatController(
...
-        } else newHistory.filter { isNewMessage(it.chatMessage) }
+        } else newHistory.filter { isNewMessage(oldHistory, it.chatMessage) }
     }
 
-    private val oldMessageIdsHash: MutableSet<String> by lazy { mutableSetOf() }
-
-    private fun isNewMessage(newMessage: ChatMessage): Boolean =
-        oldMessageIdsHash.add(newMessage.id)
+    private fun isNewMessage(oldHistory: List<ChatItem>?, newMessage: ChatMessage): Boolean =
+        oldHistory
```
Pelkar, line 1269: I would do the following:
* start with a let to get rid of the nullable, if oldHistory is null, then return true
* remove the type declarations, they are not needed, I think you can remove the map line entirely
* instead of filter, use any with both conditions with &&


Final code would be:
`oldHistory?.let { old ->
            old.filterIsInstance<LinkedChatItem>().any { oldMessage ->
                oldMessage.messageId != null && oldMessage.messageId == newMessage.id
            }
        } ?: true`

Pelkar (commented):
The current code can be condensed, but the logic should be solid

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -1262,13 +1262,17 @@ class ChatController(
...
             newHistory
-        } else newHistory.filter { isNewMessage(it.chatMessage) }
+        } else newHistory.filter { isNewMessage(oldHistory, it.chatMessage) }
     }
 
-    private val oldMessageIdsHash: MutableSet<String> by lazy { mutableSetOf() }
-
-    private fun isNewMessage(newMessage: ChatMessage): Boolean =
-        oldMessageIdsHash.add(newMessage.id)
+    private fun isNewMessage(oldHistory: List<ChatItem>?, newMessage: ChatMessage): Boolean =
```
DavDo, line 1268: Just interesting, what was the cause of the message duplication with the previous solution?

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
AndriiHorishniiMOC: In the method `removeDuplicates` we had:
```
        return if (newHistory.isNullOrEmpty() || oldHistory.isNullOrEmpty()) {
            newHistory
        } else
```
So, in the beginning, it added all messages without adding their ids to the `oldMessageIdsHash`.

DavDo: May it be productive to add all ids to hash at this point instead of iterating through all old messages for every new one?

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
AndriiHorishniiMOC: You lost to call `.not()`. But yeah, this can be simplified. But I prefer not to use `?.let` in the beggining.

AndriiHorishniiMOC:
I cover the remove duplication functionality with unit tests.

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
AndriiHorishniiMOC: I believe it does not make a lot of sense if we talking about not huge list. Instead, we will not have inconsistency between two collections.

andrews-moc (approved):

Pelkar (approved):
Looks good to me

gugalo (approved):

AndriiHorishniiMOC:
!squash

AndriiHorishniiMOC:
!merge

PR #455, opened 2023-02-17
Request screen sharing permission in Call Visualizer flow

andrews-moc:
[MOB-1859](https://glia.atlassian.net/browse/MOB-1859)

Not so many changes, but a long description for understanding them. 🙂 

**HOW SCREEN SHARING FLOW WORKS NOW**

1. Operator requests screen sharing
2. Widgets SDK asks the visitor to accept the request via `AlertDialog`
3. If visitor accepts the request, Widgets SDK calls Core SDK [accept()](https://github.com/salemove/android-sdk-widgets/blob/master/widgetssdk/src/main/java/com/glia/widgets/core/screensharing/data/GliaScreenSharingRepository.java#L58-L64) function
4. Core SDK inside it's `medra` module calls [startActivityForResult()](https://github.com/salemove/android-sdk/blob/master/medra/src/main/java/com/glia/medra/internal/webrtc/Permissions.java#L39) to [acquire a media projection token](https://developer.android.com/guide/topics/large-screens/media-projection#legacy_approach). Visitor has to accept system permission for exposing sensitive info during casting/recording
5. Widgets SDK overrides the `onActivityResult()` function in [CallActivity](https://github.com/salemove/android-sdk-widgets/blob/master/widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java#L130) as well as in [ChatActivity](https://github.com/salemove/android-sdk-widgets/blob/master/widgetssdk/src/main/java/com/glia/widgets/chat/ChatActivity.java#L102) and passes result to `GliaWidgets.onActivityResult()` function

**PROBLEM**
It was working well for regular engagements where screen sharing requests could happen only on `CallActivity` or `ChatActivity`. However, for Call Visualizer engagements, screen sharing request should be shown on top of any Activity of an integrator's app (not only `CallActivity` or `ChatActivity`). To make the current approach work, we would have had to ask the integrator to override `onActivityResult()` for each Activity that exists in integrator's app. I think that's not a good way to go.

**SOLUTION**
The idea is to use the [recommended approach](https://developer.android.com/guide/topics/large-screens/media-projection#recommended_approach) for acquiring a media projection token **for all cases, including `CallActivity` and `ChatActivity`**. This approach doesn't need to expect the result in `onActivityResult()` callback. Instead, it registers it's own callback on Activity creation. It's possible to register our own callback for each Activity using the [ActivityWatcher](https://github.com/salemove/android-sdk-widgets/blob/master/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForDialogs.kt) implemented earlier for CallVisualizer. However, we need to keep in mind that this approach works only for `Activity` that is extended from `ComponentActivity`. For other activities (I suppose it's rather rare case) we can offer our legacy approach - integrator will need to override the `onActivityResult()` function and pass result to `GliaWidgets.onActivityResult(requestCode, resultCode, data);` interface.

**NEW LOGIC**

- If permission is requested from `ComponentActivity`=> Core SDK should not call `startActivityForResult()`. It will be done by Widgets SDK. This way we will not show permissions dialog twice for a visitor.
- If permission is requested not from `ComponentActivity`=> `medra` will use Android legacy approach to ask for permission and integrator will need to add `GliaWidgets.onActivityResult(requestCode, resultCode, data);`  into each Activity's `onActivityResult()` function.

**NEW FLOW**

1. Operator requests screen sharing
2. Widgets SDK asks the visitor to accept the request via AlertDialog
3. If visitor accepts the request, Widgets SDK will use the [recommended approach](https://developer.android.com/guide/topics/large-screens/media-projection#recommended_approach) for acquiring a media projection token. So, Widgets SDK will register an Activity result callback and launch Media Projection intent.
4. Visitor has to accept system permission for exposing sensitive info during casting/recording
5. Core SDK checks if permission is asked from ComponentActivity. If so, then `medra` module will not call the legacy `startActivityForResult()`. It will just wait for result in the same place as before.

[MOB-1859]: https://glia.atlassian.net/browse/MOB-1859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherForDialogsTest.kt
+++ b/widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherForDialogsTest.kt
@@ -1,8 +1,14 @@
 package com.glia.widgets.callvisualizer
 
 import android.app.Activity
+import android.media.projection.MediaProjectionManager
+import androidx.activity.ComponentActivity
+import androidx.activity.result.contract.ActivityResultContract
+import androidx.fragment.app.FragmentActivity
+import com.glia.widgets.call.CallActivity
 import com.glia.widgets.callvisualizer.controller.CallVisualizerController
 import com.glia.widgets.callvisualizer.domain.IsCallOrChatScreenActiveUseCase
+import com.glia.widgets.chat.ChatActivity
 import com.glia.widgets.core.dialog.Dialog
 import com.glia.widgets.core.dialog.DialogController
 import com.glia.widgets.core.dialog.model.DialogState
@@ -11,7 +17,8 @@ import junit.framework.TestCase.assertNull
 import org.junit.Assert.assertNotNull
 import org.junit.Before
 import org.junit.Test
-import org.mockito.Mockito.mock
+import org.mockito.Mockito.*
+import org.mockito.kotlin.verify
 import java.lang.ref.WeakReference
 
 internal class ActivityWatcherForDialogsTest {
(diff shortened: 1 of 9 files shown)


DavDo (approved):

AndriiHorishniiMOC (approved):

widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherForDialogsTest.kt
```diff
@@ -93,4 +100,54 @@ internal class ActivityWatcherForDialogsTest {
...
+
+        verify(activity, never()).getSystemService(MediaProjectionManager::class.java)
+    }
+
+    @Test
+    fun mediaProjectionObjects_creation_whenComponentActivity() {
+        val activity = mock(ComponentActivity::class.java)
+        activityWatcherForDialogs.registerForMediaProjectionPermissionResult(activity)
+
+        verify(activity, times(1)).getSystemService(MediaProjectionManager::class.java)
```
Pelkar, line 135: NIT you don't need the .times(1) as it is implicit. If it is called twice, then the test will fail and if it is not called, it will fail. times(1) is the default value

Pelkar (approved):
Looks really good, especially the migration to service part. Added one nitpick, but approved either way

andrews-moc:
Closing in favor of [this](https://github.com/salemove/android-sdk-widgets/pull/461) approach.

PR #456, opened 2023-02-20
Feature/mob 1825 end screen sharing functionality

Pelkar:
Added logic to the button in endScreenSharingView to end sharing


DavDo (approved):

app/src/main/res/values/strings.xml
```diff
@@ -73,6 +73,7 @@
     <string name="main_start_chat_flow">Start new chat flow</string>
     <string name="main_start_message_center_flow">Start new messaging flow</string>
     <string name="main_start_audio_call">Start new audio call</string>
+    <string name="main_start_screen_share_test_view">Open screen share test view</string>
```
andrews-moc, line 76: It seems to me that this string is not needed, is it?

andrews-moc (approved):

PR #459, opened 2023-02-23
Mark messages in chat transcript as read after six seconds

DavDo:
Retrieving chat transcripts doesn’t mark messages as read for some reason. This has to be done manually, using the Core SDK method created in the related task linked below. From the moment you fetch a chat transcript, after six seconds, the method should be called. Calling this should clear the “new messages” indicator created in the task linked below.

**Jira issue:**
https://glia.atlassian.net/browse/MOB-1737


AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -1288,6 +1289,20 @@ internal class ChatController(
         )
     }
 
+    private fun markMessagesReadWithDelay() {
```
Pelkar, line 1292: Are we not clearing the disposable?

Pelkar (approved):
Looks good!

Added one comment and please add new lines to get rid of github showing error signs in code review stage

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
DavDo: But why do you think we need to do it here?

PR #461, opened 2023-02-23
MOB-1859 Accept screen sharing with a special request code when there is no need to ask screen sharing permission from core SDK

andrews-moc:
Alternative solution to [this PR](https://github.com/salemove/android-sdk-widgets/pull/455)


AndriiHorishniiMOC (approved):

andrews-moc:
!merge

PR #463, opened 2023-02-24
Show attachment button based on site configuration

AndriiHorishniiMOC:
Show attachment button on Welcome and Transcript screens based on site configuration.

[MOB-1877](https://glia.atlassian.net/browse/MOB-1877)

[MOB-1877]: https://glia.atlassian.net/browse/MOB-1877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/messagecenter/MessageCenterController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/messagecenter/MessageCenterController.kt
@@ -6,7 +6,9 @@ import com.glia.androidsdk.GliaException
 import com.glia.androidsdk.RequestCallback
 import com.glia.androidsdk.chat.VisitorMessage
 import com.glia.androidsdk.engagement.EngagementFile
+import com.glia.androidsdk.site.SiteInfo
 import com.glia.widgets.chat.domain.IsAuthenticatedUseCase
+import com.glia.widgets.chat.domain.SiteInfoUseCase
 import com.glia.widgets.core.dialog.DialogController
 import com.glia.widgets.core.fileupload.domain.AddFileToAttachmentAndUploadUseCase
 import com.glia.widgets.core.fileupload.model.FileAttachment
@@ -23,6 +25,7 @@ internal class MessageCenterController(
     private val removeFileAttachmentObserverUseCase: RemoveSecureFileAttachmentObserverUseCase,
     private val removeFileAttachmentUseCase: RemoveSecureFileAttachmentUseCase,
     private val isAuthenticatedUseCase: IsAuthenticatedUseCase,
+    private val siteInfoUseCase: SiteInfoUseCase,
     private val dialogController: DialogController
 ) : MessageCenterContract.Controller {
     private var view: MessageCenterContract.View? = null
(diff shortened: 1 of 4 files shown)


Pelkar (approved):
LGTM 👍

widgetssdk/src/main/java/com/glia/widgets/messagecenter/MessageCenterController.kt
```diff
@@ -50,6 +53,19 @@ internal class MessageCenterController(
...
+    }
+
+    private fun updateAllowFileSendState() {
+        siteInfoUseCase.execute { siteInfo: SiteInfo?, _ ->
+            onSiteInfoReceived(siteInfo)
+        }
+    }
+
+    private fun onSiteInfoReceived(siteInfo: SiteInfo?) {
+        val attachmentAllowed = siteInfo == null || siteInfo.allowedFileSenders.isVisitorAllowed
```
DavDo, line 67: NIT:
```suggestion
        val attachmentAllowed = siteInfo?.allowedFileSenders?.isVisitorAllowed ?: true
```
But maybe it would be better to hide attachment button when `siteInfo` is `null`

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/messagecenter/MessageCenterController.kt
andrews-moc: `val attachmentAllowed = siteInfo?.allowedFileSenders?.isVisitorAllowed ?: false`
So, if `siteInfo` is `null` then `attachmentAllowed` should be `false`, instead of `true`, isn't it?

widgetssdk/src/main/java/com/glia/widgets/messagecenter/MessageView.kt
```diff
@@ -220,7 +220,12 @@ class MessageView(
         updateSendButtonState(state.sendMessageButtonState)
         updateSendMessageError(state.showMessageLimitError)
         updateMessageEditText(state.messageEditTextEnabled, state.showMessageLimitError)
-        addAttachmentButton.isEnabled = state.addAttachmentButtonEnabled
+        if (state.addAttachmentButtonVisible) {
+            addAttachmentButton.visibility = VISIBLE
+            addAttachmentButton.isEnabled = state.addAttachmentButtonEnabled
+        } else {
+            addAttachmentButton.visibility = GONE
+        }
```
andrews-moc, line 228: Do we need to execute `addAttachmentButton.isEnabled = state.addAttachmentButtonEnabled` for the `else` case as well?

AndriiHorishniiMOC: Just optimization. The button is invisible, so it doesn't matter if is it enable or not.

widgetssdk/src/main/java/com/glia/widgets/messagecenter/MessageCenterController.kt
```diff
@@ -50,6 +53,19 @@ internal class MessageCenterController(
...
+    }
+
+    private fun updateAllowFileSendState() {
+        siteInfoUseCase.execute { siteInfo: SiteInfo?, _ ->
+            onSiteInfoReceived(siteInfo)
+        }
+    }
+
+    private fun onSiteInfoReceived(siteInfo: SiteInfo?) {
+        val attachmentAllowed = siteInfo?.allowedFileSenders?.isVisitorAllowed ?: true
```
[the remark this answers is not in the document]
AndriiHorishniiMOC: Yes, the same behavior as for the chat screen.

PR #465, opened 2023-02-28
Added hide audio buttons when callvisualizer session is active

Pelkar:
MOB-1817

Implemented Call visualizer engagement disables CallView audio controls

https://salemove.slack.com/archives/C04BTHLDL9Y/p1677679060595919 decision to not show disabled buttons.

no controls

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
@@ -24,6 +24,7 @@ class CallState {
     public final boolean isOnHold;
     //    Need this to not update all views when only time is changed.
     public final boolean isOnlyTimeChanged;
+    public final boolean isCallVisualizer;
 
     @NonNull
     @Override
@@ -40,6 +41,7 @@ class CallState {
                 ", requestedMediaType: " + requestedMediaType +
                 ", isSpeakerOn: " + isSpeakerOn +
                 ", isOnHold: " + isOnHold +
+                ", isCallVisualizer: " + isCallVisualizer +
                 '}';
     }
 
@@ -59,14 +61,15 @@ class CallState {
                 Objects.equals(requestedMediaType, callState.requestedMediaType) &&
                 isSpeakerOn == callState.isSpeakerOn &&
                 isOnHold == callState.isOnHold &&
-                isOnlyTimeChanged == callState.isOnlyTimeChanged;
+                isOnlyTimeChanged == callState.isOnlyTimeChanged &&
+                isCallVisualizer == callState.isCallVisualizer;
     }
 
     @Override
     public int hashCode() {
         return Objects.hash(integratorCallStarted, isVisible, messagesNotSeen,
                 callStatus, landscapeLayoutControlsVisible, isMuted, hasVideo,
-                companyName, requestedMediaType, isSpeakerOn, isOnHold, isOnlyTimeChanged);
+                companyName, requestedMediaType, isSpeakerOn, isOnHold, isOnlyTimeChanged, isCallVisualizer);
     }
 
     public boolean showOperatorStatusView() {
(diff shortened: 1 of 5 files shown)


widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
```diff
@@ -998,6 +992,27 @@ internal class CallView(
         }
     }
 
+    private fun setVisibilityBasedOnViewState(view1: View, view2: View, state: ViewState) {
```
andrews-moc, line 995: I would suggest to avoid names like `view1`, `view2`. Here, for example, we can name them `button` and `label`, I think.

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
```diff
@@ -998,6 +992,27 @@ internal class CallView(
...
+                view1.isEnabled = false
+                view2.isEnabled = false
+            }
+            ViewState.HIDE -> {
+                view1.visibility = View.GONE
+                view2.visibility = View.GONE
+            }
+        }
+    }
+
```
DavDo, line 1015: NIT: 
```suggestion
    private fun applyViewState(state: ViewState, vararg views: View) {
        val isVisible = state != ViewState.HIDE
        val isEnabled = state == ViewState.SHOW
        views.forEach {
            it.isVisible = isVisible
            it.isEnabled = isEnabled
        }
    }
```

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
Pelkar: I added Davids more elegant solution

widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
```diff
@@ -328,16 +331,28 @@ public CallState setTransferring() {
                 .createCallState();
     }
 
-    public boolean isMuteButtonEnabled() {
-        return (isAudioCall() || isVideoCall()) && !showOnHold();
+    public ViewState getGetMuteButtonViewState() {
```
andrews-moc, line 334: Looks like "Get" is unnecessary here, isn't it?
`getGetMuteButtonViewState()` => `getMuteButtonViewState()`

widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
```diff
@@ -328,16 +331,28 @@ public CallState setTransferring() {
...
+        }
     }
 
     public boolean isVideoButtonEnabled() {
         return is2WayVideoCall() && !showOnHold();
     }
 
-    public boolean isSpeakerButtonEnabled() {
-        return isAudioCall() || isVideoCall();
+    public ViewState getGetSpeakerButtonViewState() {
```
andrews-moc, line 348: Looks like "Get" is unnecessary here, isn't it?
`getGetSpeakerButtonViewState()` => `getSpeakerButtonViewState()`

widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
andrews-moc: Could you also please explain, where is this new function used? Cannot find the usage in the current PR. 🤔

widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
```diff
@@ -328,16 +331,28 @@ public CallState setTransferring() {
                 .createCallState();
     }
 
-    public boolean isMuteButtonEnabled() {
-        return (isAudioCall() || isVideoCall()) && !showOnHold();
+    public ViewState getGetMuteButtonViewState() {
+        if (isCallVisualizer) {
+            return ViewState.HIDE;
```
gugalo, line 336: I might be mistaken about something but in design mocks these buttons are actually disabled (grayed out) in the case of Call Visualizer and not hidden. But based on logic in `applyViewState` if I understood it correctly it actually set them as not visible.

gugalo (commented):
Can you please add a screenshot of the UI to the PR

widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
gugalo: Same for the 'speaker' button

gugalo: Also, this is my mistake for not having a clear description of the ticket. But the same logic should also apply to the 'chat' button as well.

Pelkar: Made changes based on this thread:

https://salemove.slack.com/archives/C04BTHLDL9Y/p1677679060595919

widgetssdk/src/main/java/com/glia/widgets/call/CallState.java
Pelkar: It is called from CallView LN:968

gugalo (approved):

andrews-moc (approved):

PR #478, opened 2023-03-06
Added Visitor Context adding access point

Pelkar:
MOB-1832

Added a way to add visitor context and have visitor code view use it

* Made minor adjustment to reading data from prefs to avoid empty data being passed on. 
* Added new access point in CallVisualizer for adding context
* Made accept call visualizer engagement use the visitor context value

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/MainFragment.java
+++ b/app/src/main/java/com/glia/exampleapp/MainFragment.java
@@ -36,6 +36,7 @@ import com.glia.widgets.call.CallActivity;
 import com.glia.widgets.call.Configuration;
 import com.glia.widgets.callvisualizer.EndScreenSharingActivity;
 import com.glia.widgets.chat.ChatActivity;
+import com.glia.widgets.core.callvisualizer.domain.CallVisualizer;
 import com.glia.widgets.core.configuration.GliaSdkConfiguration;
 import com.glia.widgets.messagecenter.MessageCenterActivity;
 import com.glia.widgets.view.VisitorCodeView;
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/glia/exampleapp/Utils.java
```diff
@@ -17,7 +17,11 @@
 class Utils {
 
     public static String getStringFromPrefs(@StringRes int keyValue, String defaultValue, SharedPreferences sharedPreferences, Resources resources) {
-        return sharedPreferences.getString(resources.getString(keyValue), defaultValue);
+        String value = sharedPreferences.getString(resources.getString(keyValue), defaultValue);
+        if (value.trim().isEmpty()) {
+            return null;
+        }
+        return value;
```
andrews-moc, line 24: I doubt if it won't break the code which already expects that an empty string for default value returns an empty string instead of `null`.
Unless you are 100% sure it won't, I would rather change the validation in `MainFragment` to `if (visitorContext is not empty) `.

andrews-moc:
Other than my single point, looks good. 👍

app/src/main/java/com/glia/exampleapp/MainFragment.java
```diff
@@ -377,7 +378,13 @@ private void clearSession() {
     }
 
     private void showVisitorCode() {
```
gugalo, line 380: As far as I understand there is currently no default value for the Visitor Context and I can't see anything being set for it in the example app. Or did I miss it somehow?

gugalo (approved):

app/src/main/java/com/glia/exampleapp/Utils.java
Pelkar: E/CallVisualizerRepository: Error during accepting engagement request, reasonvisitorContextAssetId must be UUID

Empty value is not null and can cause unexpected error while testing. That is why the change

app/src/main/java/com/glia/exampleapp/MainFragment.java
Pelkar: Default value is null as it was before

app/src/main/java/com/glia/exampleapp/Utils.java
andrews-moc: Yes, I understand this. 

And what I suggested is to avoid changing the `getStringFromPrefs()` common function behavior. Instead, change `MainActivity:384` from `if (visitorContext != null)` to `if (visitorContext != null && visitorContext.trim().isEmpty())`.

Does it make sense to you?

andrews-moc: The main reason for this suggestion is that I am not sure that if you change the behavior of the function (replace empty string with `null`) which is used in many other places, it will not break anything for current usages. Are you sure in this?

andrews-moc (approved):

Pelkar:
!squash

PR #480, opened 2023-03-07
Improved ADA compatibility

Pelkar:
MOB-1819

This was the best I was able to do, minor imperfections keep happening, but I think the current version is good enough.

TalkBack will talk to finish the sentence and depending on the loading time might skip some elements. For this reason I disabled contentDescription for close and glia image because those kept coming up in undesired ways. Close button description is added back in a way that won't be read out automatically

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/view/VisitorCodeView.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/view/VisitorCodeView.kt
@@ -5,6 +5,7 @@ import android.content.res.ColorStateList
 import android.content.res.TypedArray
 import android.os.CountDownTimer
 import android.view.View
+import android.view.accessibility.AccessibilityEvent
 import android.widget.FrameLayout
 import android.widget.ProgressBar
 import android.widget.TextView
@@ -22,11 +23,12 @@ import com.glia.widgets.di.Dependencies
 import com.glia.widgets.helper.Logger
 import com.glia.widgets.helper.Utils
 import com.glia.widgets.view.button.GliaPositiveButton
+import com.google.android.material.theme.overlay.MaterialThemeOverlay
+import com.glia.widgets.view.unifiedui.exstensions.*
 import com.glia.widgets.view.unifiedui.exstensions.applyButtonTheme
 import com.glia.widgets.view.unifiedui.exstensions.applyLayerTheme
 import com.glia.widgets.view.unifiedui.exstensions.applyTextTheme
 import com.glia.widgets.view.unifiedui.exstensions.layoutInflater
-import com.google.android.material.theme.overlay.MaterialThemeOverlay
 import com.glia.widgets.view.unifiedui.exstensions.applyImageColorTheme
 
 /**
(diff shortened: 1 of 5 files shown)


widgetssdk/src/main/java/com/glia/widgets/view/VisitorCodeView.kt
```diff
@@ -100,6 +99,9 @@ class VisitorCodeView internal constructor(
         runOnUi {
             showProgressBar(true)
             showSuccess()
+            successTitle.contentDescription = context.getString(R.string.glia_visitor_code_loading)
+            successTitle.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_ACCESSIBILITY_FOCUSED)
```
gugalo, line 103: Can you please add some comments inside the code for this line. Even checking the [property doc](https://developer.android.com/reference/android/view/accessibility/AccessibilityEvent#TYPE_VIEW_ACCESSIBILITY_FOCUSED) I don't understand why it is needed.

widgetssdk/src/main/java/com/glia/widgets/view/VisitorCodeView.kt
```diff
@@ -100,6 +99,9 @@ class VisitorCodeView internal constructor(
...
+            closeButton.contentDescription = context.getString(R.string.glia_chat_alert_dialog_close_content_description)
```
gugalo, line 104: Is there some specific reason why the close button content description is done programmatically? I guess it is related to the:
>For this reason I disabled contentDescription for close and glia image because those kept coming up in undesired ways.

Could be useful to have an in-code comment so that someone during refactoring does not move it back to XML

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/exstensions/ResourceExtensions.kt
```diff
@@ -16,3 +16,7 @@ fun Context.getDimenRes(@DimenRes dimenId: Int): Float {
 fun Context.getDimenResPx(@DimenRes dimenId: Int): Int {
     return this.resources.getDimensionPixelSize(dimenId)
 }
+
+fun String.separateStringWithSymbol(symbol: String): String {
+    return this.split("").joinToString(symbol)
+}
```
gugalo, line 22: NIT: You can ignore this but I think it might be better to extract it to some String dedicated utils class or a dedicated String extensions class.

gugalo (commented):
I think that Henric wanted to see how this works before it is merged to the master. Can you please make some video and post it in #prj-mobile-call-visualizer Slack channel?

andrews-moc:
By the way, Bitrise build has failed. Maybe it has something to do with conflicting changes. Could you please resolve them?

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/exstensions/ResourceExtensions.kt
Pelkar: Made a PrimitiveExtensions.kt class, might rename it later because technically String is not a primitive, but I would like to have String, Int, Long etc extensions in one place

widgetssdk/src/main/java/com/glia/widgets/view/VisitorCodeView.kt
Pelkar: Added comment

Pelkar:
!squash

PR #495, opened 2023-03-13
Fix a merge mechanism doesn't support Lists and maps

DavDo:
Need to change merge mechanisms to fit logic: if data class → deep merge otherwise → take value from the `right` if not null otherwise -> take the `left` value

MOB-1951

**Jira issue:**
https://glia.atlassian.net/browse/MOB-1951


andrews-moc (approved):

Pelkar (approved):
LGTM 👍

DavDo:
!merge

PR #497, opened 2023-03-14
Rename ActivityWatcherForDialogs to ActivityWatcherForCallVisualizer

andrews-moc:
This is a small PR aimed to rename 
`ActivityWatcherForDialogs` to `ActivityWatcherForCallVisualizer`.

The main idea is that `ActivityWatcherForDialogs` is currently not only for dialogs but for bubble also. That's why I think the new name can reflect what's class purpuse better.


DavDo (approved):

gugalo (approved):
Maybe just ActivityWatcher then?
I think later on it might have even more uses, e.g for Live Observation. Though we can rename it later as well if the need arises.

AndriiHorishniiMOC (approved):

andrews-moc:
> Maybe just ActivityWatcher then? I think later on it might have even more uses, e.g for Live Observation. Though we can rename it later as well if the need arises.

I think it's better to split responsibility as much as possible.

For example, I would go with a separate ActivityWatcher for a separate feature like LiveObservation, if possible.

If not  - we can rename the current ActivityWatcher later if the need arises.

andrews-moc:
!merge

PR #502, opened 2023-03-16
Fix acceptance tests

gugalo:
**Additional info:**
Acceptance tests started failing after this PR: https://github.com/salemove/android-sdk-widgets/pull/487
Because SDK configurations set up using deep-link where missing `companyName` property while the same property has been removed from the example app method that launched Widget SDK call/chat Activity because they have been deprecated.

**P.S: It is easier to review if you ignore changes made in the first commit which is just Android Studio's automated conversion of class to Kotlin. All related to conversion issues and the acceptance tests fix are resolved in the next commit separately**


andrews-moc (approved):

app/src/main/java/com/glia/exampleapp/GliaWidgetsConfigManager.kt
```diff
@@ -74,29 +78,8 @@ object GliaWidgetsConfigManager {
...
-            .setContext(applicationContext)
-            .build()
-    }
-
     @JvmOverloads
-    fun createDefaultConfig(
-        applicationContext: Context,
-        uiJsonRemoteConfig: String? = null
-    ): GliaWidgetsConfig {
+    fun createDefaultConfig(applicationContext: Context, uiJsonRemoteConfig: String? = null): GliaWidgetsConfig {
```
DavDo, line 82: `@JvmStatic` annotation will help to use this method directly/without `INSTANCE` in the java code

DavDo (approved):

app/src/main/java/com/glia/exampleapp/GliaWidgetsConfigManager.kt
gugalo: Thanks. That much nicer then INSTANCE option 👍

gugalo: Fixed in bdb1e06

gugalo:
!squash

PR #505, opened 2023-03-16
Fix the Screen Overlay permission request

AndriiHorishniiMOC:
Prevent to show the Screen Overlay permission on the Welcome to Message Center screen.

The crash cause: In case when the Welcome screen opens then the Chat screen opens:
- Chat controller during initialization asks to display the Screen Overlay dialog;
- Chat view is not ready for the handling of the dialogs;
- But the Welcome view is still ready to handle dialogues. However, it can't handle the Screen Overlay dialog. 

[MOB-1965](https://glia.atlassian.net/browse/MOB-1965)

[MOB-1965]: https://glia.atlassian.net/browse/MOB-1965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
@@ -184,8 +184,6 @@ internal class ChatController(
         val queueIds = if (queueId != null) arrayOf(queueId) else emptyArray()
         engagementConfigUseCase(chatType, queueIds)
 
-        if (isShowOverlayPermissionRequestDialogUseCase.execute()) dialogController.showOverlayPermissionsDialog()
-
         ensureSecureMessagingAvailable()
 
         if (chatState.integratorChatStarted || dialogController.isShowingChatEnderDialog) {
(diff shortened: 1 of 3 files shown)


andrews-moc:
I don't have any comments.
Though, I would ask the following:

1. Could you please add the steps to reproduce the issue with the expected and actual result?
2. How is `ChatController` related to the Message Center screen? My expectation was that `ChatController` should not affect the flow that is being fixed and shouldn't appear in changelog.

andrews-moc (approved):

DavDo (approved):

DavDo:
> I don't have any comments.
> 
> Though, I would like to ask the following:
> 
> 1. Could you please add the steps to reproduce the issue with the expected and actual result?
> 2. How is `ChatController` related to the Message Center screen? My expectation was that `ChatController` should not affect the flow that is being fixed and shouldn't appear in the changelog.

If I'm not mistaken this behavior is related to `DialogController` which is a `singleton`, we have the same `DialogController` instance for both `MessageCenter` and `Chat` views, and sometimes it con cause collisions.

PR #508, opened 2023-03-17
Parse default theme from attributes for ChatHeadView (bubble)

andrews-moc:
This is something that I was not able to do before.
Currently I've found another way to parse theme from attributes.
Looks like it's working fine - bubble gets default theme on the init stage as well.


DavDo (approved):

andrews-moc:
!merge

PR #511, opened 2023-03-20
Refactored ActivityWatcherForDialogs to support conf changes

Pelkar:
MOB-1944

Added dismiss dialog on pause and recreate it on resume
Refactored ActivityWatcherForDialogs to have it's own controller

Some things need to be discussed and improved on based on feedback, please review
@VisibleForTesting annotations need to be either removed or used in tests form previous implementation
Watcher tests need to be either deleted or updated


widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizer.kt
```diff
@@ -1,79 +1,60 @@
...
-import com.glia.widgets.view.head.controller.ServiceChatHeadController
-import com.google.android.material.theme.overlay.MaterialThemeOverlay
+import com.glia.widgets.view.unifiedui.exstensions.prepareContextWithStyle
 import java.lang.ref.WeakReference
 
 class ActivityWatcherForCallVisualizer(
-    private val callVisualizerController: CallVisualizerController,
-    private val screenSharingController: ScreenSharingController,
     private val dialogController: DialogController,
-    private var serviceChatHeadController: ServiceChatHeadController,
```
andrews-moc, line 49: @Pelkar Could you please explain the idea on the following topic? What is the reason why you left `DialogController` in `ActivityWatcherForCallVisualizer` and moved the other three controllers to `ActivityWatcherController`?

widgetssdk/src/main/java/com/glia/widgets/di/Dependencies.java
```diff
@@ -63,11 +63,7 @@ public static void onAppCreate(Application application) {
                 new ApplicationLifecycleManager(),
                 controllerFactory.getChatHeadController()
         );
-        ActivityWatcherForCallVisualizer activityWatcherForCallVisualizer = new ActivityWatcherForCallVisualizer(
-                controllerFactory.getCallVisualizerController(),
-                controllerFactory.getScreenSharingController(),
-                controllerFactory.getDialogController(),
-                controllerFactory.getChatHeadController());
+        ActivityWatcherForCallVisualizer activityWatcherForCallVisualizer = new ActivityWatcherForCallVisualizer(getControllerFactory().getDialogController(), getControllerFactory().getActivityWatcherController());
```
andrews-moc, line 66: NIT. I would suggest moving `getControllerFactory().getDialogController()` to a separate line.

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/exstensions/ViewExstensions.kt
```diff
@@ -320,3 +323,12 @@ internal fun GliaSurveyOptionButton.applyOptionButtonTheme(theme: OptionButtonTh
...
 }
+
+internal fun Activity.prepareContextWithStyle() : Context {
+    return MaterialThemeOverlay.wrap(
+        this,
+        null,
+        R.attr.gliaChatStyle,
+        R.style.Application_Glia_Chat
+    )
+}
```
andrews-moc, line 334: If we are extracting this to an extension function, I think it makes sense to use it in other places where `MaterialThemeOverlay.wrap()` is called. What do you think?

widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherControllerTest.kt
```diff
@@ -0,0 +1,243 @@
...
+        val view = mock(View::class.java)
+        whenever(activity.window).thenReturn(window)
+        whenever(window.decorView).thenReturn(view)
+        controller.setWatcher(watcher)
+        cleanup()
+        reset(watcher, callVisualizerController, callVisualizerRepository, serviceChatHeadController, screenSharingController)
+    }
+
+    @Test
+    fun `onActivityResumed call or chat is active`() {
```
andrews-moc, line 48: We had an [agreement](https://salemove.slack.com/archives/C01F9K88SQZ/p1677586771381839?thread_ts=1677153711.779559&cid=C01F9K88SQZ) recently that test name should have the following structure:

- First is the method that is being tested (e.g: "onActivityResumed")
- Second is the expected behavior of the test case (e.g: "completes with 0")
- Last is when that test case is applicable (e.g: "when error is returned")

It seems to me that we are missing the second part. That's why it's difficult to understand the test, at least for me.

Could you please make sure all tests in this PR have agreed name structure?

widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherControllerTest.kt
andrews-moc, line 1: Thank you for adding new test cases 👍

widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
```diff
@@ -308,4 +311,11 @@ public EndScreenSharingContract.Controller getEndScreenSharingController() {
     public VisitorCodeContract.Controller getVisitorCodeController() {
         return new VisitorCodeController(dialogController, repositoryFactory.getVisitorCodeRepository(), repositoryFactory.getGliaEngagementRepository());
     }
+
+    public ActivityWatcherContract.Controller getActivityWatcherController() {
+        if (activityWatcherController == null) {
+            activityWatcherController = new ActivityWatcherController(getCallVisualizerController(), getScreenSharingController(), serviceChatHeadController);
```
andrews-moc, line 317: NIT. I would suggest moving each parameter to a separate line.

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizer.kt
```diff
@@ -128,326 +94,245 @@ class ActivityWatcherForCallVisualizer(
...
-                        showVisitorCodeDialog(resumedActivity)
-                    }
-                else -> {
-                    Logger.d(TAG, "Unexpected dialog mode received")
-                }
-            }
+    override fun dismissAlertDialog(manualDismiss: Boolean) {
+        Logger.d(TAG, "Dismiss alert dialog")
+        try {
+            // dialogController.dismissCurrentDialog() // TODO Double dismiss?
```
andrews-moc, line 148: Have you found the reason for the double dismiss or this TODO is still in progress?

widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherControllerTest.kt
Pelkar: Thank you, I missed the final conclusion of that thread

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherController.kt
```diff
@@ -0,0 +1,210 @@
...
+    private val screenSharingController: ScreenSharingController,
+    private var serviceChatHeadController: ServiceChatHeadController
+) : ActivityWatcherContract.Controller {
+
+    companion object {
+        private val TAG = ActivityWatcherController::class.java.simpleName
+    }
+
+    @Mode
+    private var dialogType: Int = MODE_NONE
```
gugalo, line 37: NIT: Maybe can be renamed to `currentDialogType`

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizer.kt
```diff
@@ -128,326 +94,245 @@ class ActivityWatcherForCallVisualizer(
                 TAG, "Activity does not support ActivityResultRegistry APIs, " +
                         "legacy onActivityResult() will be used to acquire a media projection token"
             )
-            startMediaProjectionLaunchers.remove(activity::class.simpleName)
+            controller.removeMediaProjectionLaunchers(activity::class.java.simpleName)
```
DavDo, line 97: this might be confusing, because the method takes String as a key, maybe it would be better to pass activity instead?
also I see different classNames are used in different places `activity::class.java.simpleName`, `activity::class.simpleName`

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/exstensions/ViewExstensions.kt
DavDo: could be better to rename to `wrapWithMaterialThemeOverlay` to indicate what is this function doing, also I think we should have parameters for this method with default values to be able to use from all views.

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizer.kt
Pelkar: My idea was to remove as many UI references from the controller as possible with the least amount of back and fourth between the watcher and the controller. Dialog controller uses a lot of view references and that is why I decided to keep that in the watcher side, because if we wanted to do this through the controller, then a lot of back and fourth is required to keep context out of the controller. That is why I extracted the dialog state handling to the controller, but the direct view handling was left in the watcher

A fair point can be made about screen sharing controller being similar in this regard and I agree. That is why all suggestions are welcome on this topic.

widgetssdk/src/main/java/com/glia/widgets/di/Dependencies.java
Pelkar: done

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/exstensions/ViewExstensions.kt
Pelkar: done

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizer.kt
Pelkar: Still in progress

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherController.kt
Pelkar: done

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherController.kt
```diff
@@ -0,0 +1,210 @@
...
+
+    private lateinit var watcher: ActivityWatcherContract.Watcher
+
+    override fun setWatcher(watcher: ActivityWatcherContract.Watcher) {
+        this.watcher = watcher
+    }
+
+    @VisibleForTesting
+    internal var screenSharingViewCallback: ScreenSharingController.ViewCallback? = null
```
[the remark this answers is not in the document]
Pelkar: done

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizer.kt
```diff
@@ -128,326 +94,245 @@ class ActivityWatcherForCallVisualizer(
...
-                dialogController.dismissCurrentDialog()
-                screenSharingController.onScreenSharingDeclined()
-            },
-            cancelListener = {
-                it.dismiss()
-                dialogController.dismissCurrentDialog()
-                screenSharingController.onScreenSharingDeclined()
-            }
-        )
+        activity.runOnUiThread {
```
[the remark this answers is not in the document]
Pelkar: This was a precaution for refactoring stage. Now I tested what actually needs it and have it only in the two places

widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
Pelkar: done

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherController.kt
```diff
@@ -0,0 +1,210 @@
...
+    fun WeakReference<Activity?>.getGliaViewOrRootView(): View? {
+        return this.get()?.let {
+            return it.findViewById(R.id.call_view)
+                ?: it.findViewById<FilePreviewView>(R.id.file_preview_view)
+                ?: it.findViewById<ChatView>(R.id.chat_view)
+                ?: it.findViewById<EndScreenSharingView>(R.id.screen_sharing_screen_view)
+                ?: it.findViewById(android.R.id.content)
+                ?: it.window.decorView.findViewById(android.R.id.content)
+        }
+    }
```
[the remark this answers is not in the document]
Pelkar: made slight refactor and it is now in Watcher

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizer.kt
Pelkar: activity::class.java.simpleName and activity::class.simpleName return the same value, so might as well use the shorter one. I prefer keeping the view references in controller to the minimum, thats why the change. This way we are passing on exactly what we need and nothin more

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/exstensions/ViewExstensions.kt
Pelkar: And most places we are using this from is in constructor, so I don't think we can universally use it

DavDo: Sorry for the late addition, maybe it makes sense, in this case, to create this method on Context instance like - `Context.wrapWithMaterialThemeOverlay(...)`

widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherControllerTest.kt
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    @Test
+    fun `onActivityDestroyed bubble is destroyed`() {
+        controller.onActivityDestroyed()
+        verify(serviceChatHeadController).onDestroy()
+    }
+
+    @Test
+    fun isCallOrChatActive() {
```
andrews-moc, line 93: I think the name of this test doesn't follow the naming pattern.

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/exstensions/ViewExstensions.kt
Pelkar: done

widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherControllerTest.kt
Pelkar: fixed

PR #516, opened 2023-03-22
Update the core SDK dependency

AndriiHorishniiMOC:
Update the core SDK dependency to provide the secure conversations functionality.

[MOB-1046](https://glia.atlassian.net/browse/MOB-1046)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,8 @@ allprojects {
         flatDir {
             dirs '../libs'
         }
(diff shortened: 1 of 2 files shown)


Pelkar (approved):
We really should have a way of publishing snapshot builds for cases just like this, but I guess this is ok with our current capabilities

DavDo (approved):

PR #519, opened 2023-03-23
Bugfix/mob 1980 call visualizer video call not opening

Pelkar:
The issue came from ActivityWatcherForDialogs not having access to onActivityResult and could not get the camera permission result.

For AppCompatActivity this was resolved with an ActivityResultLauncher, but for Activity views, a transparent CallVisualizerSupportActivity is opened to handle everything with the same type of ActivityResultLauncher but in our transparent view that is closed upon completion.

Currently for testing purposes there is a button in MainFragment that opens a android.app.Activity type of view with what we can test this functionality. Before merger that button and view will be deleted.

### DIFF
--- a/widgetssdk/src/main/AndroidManifest.xml
+++ b/widgetssdk/src/main/AndroidManifest.xml
@@ -38,6 +38,10 @@
             android:launchMode="singleTask"
             android:theme="@style/Application.Glia.Call.Activity"
             android:windowSoftInputMode="adjustResize"/>
+        <activity android:name=".callvisualizer.CallVisualizerSupportActivity"
+            android:launchMode="singleTask"
+            android:noHistory="true"
+            android:theme="@style/Application.Glia.Translucent.Activity"/>
 
         <service
             android:name=".core.chathead.ChatHeadService"
--- a/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerSupportActivity.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerSupportActivity.kt
@@ -1,0 +1,5 @@
+package com.glia.widgets.callvisualizer
+
+import androidx.appcompat.app.AppCompatActivity
+
+class CallVisualizerSupportActivity : AppCompatActivity()
--- a/widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherControllerTest.kt
+++ b/widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherControllerTest.kt
@@ -5,6 +5,7 @@ import android.content.Intent
 import android.view.View
 import android.view.Window
 import androidx.activity.result.ActivityResultLauncher
+import androidx.appcompat.app.AppCompatActivity
 import com.glia.androidsdk.GliaException
 import com.glia.androidsdk.comms.MediaUpgradeOffer
 import com.glia.widgets.callvisualizer.controller.CallVisualizerController
@@ -12,14 +13,13 @@ import com.glia.widgets.callvisualizer.domain.IsCallOrChatScreenActiveUseCase
 import com.glia.widgets.core.dialog.Dialog.*
 import com.glia.widgets.core.dialog.model.DialogState
 import com.glia.widgets.core.screensharing.ScreenSharingController
-import com.glia.widgets.view.head.controller.ServiceChatHeadController
 import junit.framework.TestCase.assertTrue
+import junit.framework.TestCase.fail
 import org.junit.After
 import org.junit.Before
 import org.junit.Test
 import org.mockito.Mockito.mock
 import org.mockito.kotlin.*
-import java.lang.ref.WeakReference
 
 internal class ActivityWatcherControllerTest {
 
@@ -29,9 +29,10 @@ internal class ActivityWatcherControllerTest {
     private val isCallOrChatActiveUseCase = mock(IsCallOrChatScreenActiveUseCase::class.java)
     private val watcher = mock(ActivityWatcherContract.Watcher::class.java)
     private val controller = ActivityWatcherController(callVisualizerController, screenSharingController)
-    private val activity = mock(Activity::class.java)
-    private val activityReference: WeakReference<Activity?> = WeakReference(activity)
+    private val activity = mock(AppCompatActivity::class.java)
+    private val supportActivity = mock(CallVisualizerSupportActivity::class.java)
     private val view = mock(View::class.java)
+    private val mediaUpgradeOffer = mock(MediaUpgradeOffer::class.java)
 
     @Before
     fun setup() {
@@ -48,21 +49,31 @@ internal class ActivityWatcherControllerTest {
     fun `onActivityResumed screen sharing skipped when call or chat is active`() {
         whenever(callVisualizerController.isCallOrChatScreenActiveUseCase).thenReturn(isCallOrChatActiveUseCase)
         whenever(isCallOrChatActiveUseCase(activity)).thenReturn(true)
-        whenever(watcher.fetchGliaOrRootView()).thenReturn(view)
         controller.onActivityResumed(activity)
         verify(callVisualizerController, times(2)).isCallOrChatScreenActiveUseCase
     }
 
     @Test
+    fun `onActivityResumed bubble is resumed and no camera permissions are called when onScreenSharingStarted with activity`() {
+        mockOnActivityResumeAndEnsureCallbacksSet(activity)
+        controller.screenSharingViewCallback?.onScreenSharingStarted()
+    }
+
+    @Test
+    fun `onActivityResumed bubble is resumed and camera permissions are called when onScreenSharingStarted with CallVisualizerSupportActivity`() {
+        mockOnActivityResumeAndEnsureCallbacksSet(activity)
+        controller.screenSharingViewCallback?.onScreenSharingStarted()
+    }
+
+    @Test
     fun `onActivityResumed error is shown when onScreenSharingError`() {
-        `onActivityResumed callbacks are set when call or chat are not active`()
+        mockOnActivityResumeAndEnsureCallbacksSet(supportActivity)
         controller.screenSharingViewCallback?.onScreenSharingRequestError(GliaException("message", GliaException.Cause.INTERNAL_ERROR))
         verify(watcher).showToast("message")
     }
 
     @Test
     fun `onActivityPaused cleanup of callbacks`() {
-        whenever(watcher.fetchGliaOrRootView()).thenReturn(view)
         controller.onActivityPaused()
         verify(watcher).dismissAlertDialog(false)
         verify(watcher).removeDialogCallback()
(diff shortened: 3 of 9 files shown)


widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherControllerTest.kt
```diff
@@ -223,6 +226,120 @@ internal class ActivityWatcherControllerTest {
...
+        controller.onMediaUpgradeDecline(GliaException("message", GliaException.Cause.INTERNAL_ERROR))
+        verify(watcher).destroySupportActivity()
+    }
+
+    @Test
+    fun `onMediaUpgradeDecline does not destroy support activity when error occurs`() {
+        controller.onMediaUpgradeReceived(mediaUpgradeOffer, true)
+        resetMocks()
+        controller.onMediaUpgradeDecline(GliaException("message", GliaException.Cause.INTERNAL_ERROR))
+    }
```
andrews-moc, line 341: Maybe I'm missing something, but why doesn't this test contain the "verify" check? Besides, I'm not sure I understand how is it different from the previous test.

Pelkar: A few tests are incomplete due to Mockito not being able to mock direct variable access. The problem comes from:

`if (mediaUpgradeOffer.video != null && mediaUpgradeOffer.video != MediaDirection.NONE) {
                onPositiveDialogButtonClicked()
            } else {`
            
mediaUpgradeOffer has direct access to variables, no setter and no constructor. So currently we can't mock the value or create a value ourselves. The only way would be to hide the value access behind an extenal file that I might actually do, but in general I am against making code changes due to testing requirements

widgetssdk/src/main/AndroidManifest.xml
```diff
@@ -38,6 +38,9 @@
             android:launchMode="singleTask"
             android:theme="@style/Application.Glia.Call.Activity"
             android:windowSoftInputMode="adjustResize"/>
+        <activity android:name=".callvisualizer.CallVisualizerSupportActivity"
+            android:launchMode="singleTask"
```
DavDo, line 42: I'd also suggest adding  `android:noHistory="true"`

widgetssdk/src/main/res/layout/call_visualizer_support_activity.xml
DavDo, line 1: I think no need for this layout

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerSupportActivity.kt
```diff
@@ -0,0 +1,17 @@
...
+class CallVisualizerSupportActivity : AppCompatActivity() {
+
+    private lateinit var binding: CallVisualizerSupportActivityBinding
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        binding = CallVisualizerSupportActivityBinding.inflate(layoutInflater)
+        setContentView(binding.root)
+    }
+}
```
DavDo, line 17: I think it is possible to have just empty activity without setContent etc
```suggestion
import androidx.appcompat.app.AppCompatActivity

class CallVisualizerSupportActivity : AppCompatActivity()
```

DavDo (changes requested):
Couple of suggestions I didn't find the right place to write.
I didn't find permission check anywhere for camera, but we should check permissions and if it not granted then request.
`if (ContextCompat.checkSelfPermission(Activity.this, Manifest.permission.CAMERA) == PackageManager.PERMISSION_GRANTED) ...`

I also don’t like the approach with the `boolean flag` for permission requests, It would be better to start `SupportActivity` for results, and make `SupportActivity` request `permissions` and handle results itself, and then return the result to `ActivityWatcher`, this way we will reduce complexity and, will have two possible options - if current activity is `ComponentActivity` request permission from `ActivityWatcher` otherwise `startSupportActivityforresult`

widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherControllerTest.kt
andrews-moc: Is this test "working" then? I mean, is it actually testing something? If so, what is the mechanics?

andrews-moc: In other words, how does it make sure that "does not destroy support activity" is true, could you please explain just to understand?

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/exstensions/ViewExstensions.kt
```diff
@@ -320,3 +324,16 @@ internal fun GliaSurveyOptionButton.applyOptionButtonTheme(theme: OptionButtonTh
         )
     )
 }
+
+internal fun Activity.wrapWithMaterialThemeOverlay(
```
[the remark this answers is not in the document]
Pelkar: This was already done but apparently merge went a little wrong

widgetssdk/src/main/res/layout/call_visualizer_support_activity.xml
Pelkar: true, I created it at first and I thought I would need it, ended up not using it

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerSupportActivity.kt
Pelkar: true, see previous comment

Pelkar:
The reason why I went for the boolean implementation was that by using ActivityResultLauncher activity the overriden onActivityResult was skipped and I could not call finish from there. That is why the solution was made as it was. But I will try to make a more elegant solution

DavDo:
> The reason why I went for the boolean implementation was that by using ActivityResultLauncher activity the overriden onActivityResult was skipped and I could not call finish from there. That is why the solution was made as it was. But I will try to make a more elegant solution

As I understood, you've tried to use `onActivityResult` with `activityResultLauncher`, which is impossible to work. 
There are two possible solutions, both inside SupportActivity
1. request permissions the old way and get results inside `onActivityResult`
2. request permissions with `resultLauncher` and receive results inside the callback

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/exstensions/ViewExstensions.kt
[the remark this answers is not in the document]
Pelkar: fixed

widgetssdk/src/main/AndroidManifest.xml
Pelkar: added

widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherControllerTest.kt
Pelkar: I made some name changes and test functionality changes, but still, some tests test until some blocker and tests that all things that can be tested will be tested and otherwise tests that nothing that is not expected is called

andrews-moc:
@Pelkar Regarding the [It seems that the package name was wrong somehow](https://github.com/salemove/android-sdk-widgets/pull/519/commits/1ffa8828da82ede77aa856ba13ce54242e078d10) commit message.
Just letting you know, we have [agreements](https://glia.atlassian.net/wiki/spaces/ENG/pages/15728856/Style+Guide+Conventions#Git-Commit-Messages) regarding the commit message.

So, ideally, commit message should answer the question "What value will this commit bring to the project?". For example: "Add support for large screen devices".

.idea/androidTestResultsUserPreferences.xml
DavDo, line 1: Do we need this file in the repository?

widgetssdk/src/test/java/com/glia/widgets/callvisualizer/ActivityWatcherControllerTest.kt
```diff
@@ -203,22 +216,102 @@ internal class ActivityWatcherControllerTest {
...
+
     private fun prepareMediaUpgradeApplicationState() {
         val offer = mock(MediaUpgradeOffer::class.java)
         val state = DialogState.MediaUpgrade(offer, "name", MODE_MEDIA_UPGRADE)
         controller.onDialogControllerCallback(state)
         verify(watcher).showUpgradeDialog(state)
     }
 
-    private fun `onActivityResumed callbacks are set when call or chat are not active`() {
+    private fun `onActivityResumed callbacks are set when call or chat are not active`(activity: Activity) {
```
DavDo, line 304: this function looks confusing to me, especially when it is called inside another function, I think we should rename it and also make it look like an ordinary function smth like 
`fun mockOnActivityResultAndEnsureCallbacksSet()`

DavDo (changes requested):

app/src/main/res/layout/legacy_activity.xml
```diff
@@ -0,0 +1,17 @@
...
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:background="@color/glia_base_light_color">
+
+    <TextView
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="EXAMPLE LEGACY APPLICATION"
```
andrews-moc, line 11: Taking into account that it's an example app in our official public repo, it seems to me that we can consider a more informative text. For example, "Example with onActivityResult() API".

app/src/main/res/values/strings.xml
```diff
@@ -80,6 +80,7 @@
     <string name="main_deauthenticate">De-authenticate</string>
     <string name="main_clear_visitor_session">Clear Session</string>
     <string name="main_show_visitor_code">Show Visitor Code</string>
+    <string name="main_legacy_activity">Open legacy activity</string>
```
andrews-moc, line 83: Taking into account that it's an example app in our official public repo, it seems to me that we can consider a more informative text. For example, "Open non ComponentActivity" or "Open activity that doesn't support Results API".

app/src/main/res/layout/legacy_activity.xml
Pelkar: Comment in the PR header:

_Currently for testing purposes there is a button in MainFragment that opens a android.app.Activity type of view with what we can test this functionality. Before merger that button and view will be deleted._

app/src/main/res/values/strings.xml
Pelkar: Comment in the PR header:

_Currently for testing purposes there is a button in MainFragment that opens a android.app.Activity type of view with what we can test this functionality. Before merger that button and view will be deleted._

app/src/main/res/layout/legacy_activity.xml
andrews-moc: Ah, sorry, I forgot about it.

andrews-moc (approved):
I think there is no need to add `.idea/androidTestResultsUserPreferences.xml` file. Please remove it. Other than that, I don't have any comments.

DavDo (approved):

PR #524, opened 2023-03-28
Added a safeguard for seting only positive duration to timer with a f…

Pelkar:
…allback

MOB-1962

* Added a preventative check for duration to only be a positive value
* fixed tests
* destroy timer on close button clicked

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/controller/VisitorCodeController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/controller/VisitorCodeController.kt
@@ -1,6 +1,7 @@
 package com.glia.widgets.callvisualizer.controller
 
 import com.glia.androidsdk.Engagement
+import com.glia.androidsdk.omnibrowse.VisitorCode
 import com.glia.widgets.callvisualizer.VisitorCodeContract
 import com.glia.widgets.core.callvisualizer.domain.VisitorCodeRepository
 import com.glia.widgets.core.dialog.DialogController
@@ -20,6 +21,8 @@ class VisitorCodeController(
         dialogController.dismissVisitorCodeDialog()
     }
 
+    private val defaultRefreshDurationMilliseconds: Long = 3 * 60 * 1000L
+
     private lateinit var view: VisitorCodeContract.View
 
     override fun setView(view: VisitorCodeContract.View) {
@@ -30,6 +33,7 @@ class VisitorCodeController(
 
     override fun onCloseButtonClicked() {
         dialogController.dismissCurrentDialog()
+        view.destroyTimer()
     }
 
     override fun onLoadVisitorCode() {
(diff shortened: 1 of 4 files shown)


widgetssdk/src/main/java/com/glia/widgets/callvisualizer/controller/VisitorCodeController.kt
```diff
@@ -20,6 +21,8 @@ class VisitorCodeController(
         dialogController.dismissVisitorCodeDialog()
     }
 
+    private val defaultRefreshDuration: Long = 180000L
```
andrews-moc, line 24: I would like to suggest specifying the unit.
For example:
`defaultRefreshDurationMilliseconds`
or add a comment `// X minutes` or whatever
Because currently it will take some time to figure out what 180000 actually is.

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/controller/VisitorCodeController.kt
```diff
@@ -40,14 +44,24 @@ class VisitorCodeController(
...
     }
 
+    private fun failGuardDuration(visitorCode: VisitorCode) : Long {
+        return visitorCode.duration.run {
+            if (this <= 0) {
+                defaultRefreshDuration
+            } else {
+                this
+            }
+        }
```
DavDo, line 62: I really like the approach suggested by @yurii-glia .
what do you think about throwing an `exception` if the duration is negative

gugalo: This is a very siriouse question )
@mikktammeglia

Pelkar: in some ways I agree, but this is for now just to get reoccurring web accesses to a minimum. Lets discuss this separately. I will make a task to the backlog

andrews-moc: I don't see the benefits of throwing an exception because:
- in this case, an integrator would need to handle this exception
- it means extra work for an integrator, additional risk for missing the handling
- anyway, integration will need to show some error message, however, why cannot we show it on our own, or just handle it silently as this PR supposes?

Maybe I'm missing anything, but as of now, I would not throw an esception.

DavDo (approved):

Pelkar:
!squash

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/controller/VisitorCodeController.kt
AndriiHorishniiMOC: I prefer to write the number as multipliers:
`3 * 60 * 1000`

PR #529, opened 2023-03-29
Migrate notifications related classes to kotlin

DavDo:
MOB 2025


widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
gugalo, line 1: Is all these files just the automated migration and follow-up fixes or is there something else?

DavDo: I did some optimizations after automated migration like changing `execute() to operator fun invoke()`, turning some static functions into extensions, etc.

widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
```diff
@@ -159,7 +159,7 @@ public CallController(
                 .setIsSpeakerOn(false)
                 .setIsMuted(false)
                 .setHasVideo(false)
-                .setIsCallVisualizer(isCallVisualizerUseCase.execute())
+                .setIsCallVisualizer(isCallVisualizerUseCase.invoke())
```
AndriiHorishniiMOC, line 162: Call controller is still in java :(

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
DavDo: Unfortunately.
That's a bit risky)

gugalo (approved):

DavDo:
!merge

PR #534, opened 2023-04-03
Created the way for overlay permissions to be handled in call visualizer

Pelkar:
MOB-1927

Added a way for overlay permissions to be handled to CallVisualizer flow and improved test coverage.


gugalo (approved):

gugalo:
@Pelkar add more people from the mobile team for review.
I usually add 3-4 people. Anyone who has time reviews. Once you have 2 reviews you don't have to wait for others.

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherController.kt
```diff
@@ -172,18 +183,34 @@ internal class ActivityWatcherController(
...
+            }
+        }
+    }
+
+    @VisibleForTesting
+    internal fun onShowOverlayDialog(show: Boolean) {
+        if (show) {
+            this.permissionType = OVERLAY
         }
     }
```
andrews-moc, line 208: This function looks strange to me. Why?
1. It's not clear what it does, what is the purpose of it.
2. Name states that overlay dialog is showing, however, it accepts param that can be `false`. So... `onShowOverlayDialog` is called when dialog is not showing?
3. It's difficult to understand, what is the purpose of setting `OVERLAY` to `permissionType`.

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherController.kt
```diff
@@ -88,13 +98,13 @@ internal class ActivityWatcherController(
         when (state.mode) {
             MODE_NONE -> watcher.dismissAlertDialog(true)
             MODE_MEDIA_UPGRADE -> watcher.showUpgradeDialog(state as MediaUpgrade)
-            MODE_OVERLAY_PERMISSION -> watcher.showOverlayPermissionsDialog()
-            MODE_START_SCREEN_SHARING -> watcher.showScreenSharingDialog()
+            MODE_OVERLAY_PERMISSION -> Logger.d(TAG, "should not be here")
```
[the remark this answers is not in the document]
Pelkar: This actually requires some investigation I forgot to do, The overlay permission dialog is somehow handled differently and the DialogState is not emitted. I will try to make the changes so it would work the same way, but this will require some investigation first. If this starts to work properly, we might even get rid of the permissionType variable from the PR

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherController.kt
Pelkar: Removed with new logic

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherController.kt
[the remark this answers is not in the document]
Pelkar: Completely revamped the dialog logic to be in sync with previous implementation

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherController.kt
gugalo: > 1. It's not clear what it does, what is the purpose of it.
I think it was added for unit tests

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
```diff
@@ -82,9 +82,13 @@ internal class CallView(
...
         }
 
-        override fun onScreenSharingStarted() {
+        override fun onScreenSharingRequestSuccess() {
             Handler(Looper.getMainLooper()).post { appBar.showEndScreenSharingButton() }
         }
+
+        override fun onScreenSharingStarted() {
+            // no-op
+        }
```
DavDo, line 91: to avoid such kinds of `empty` functions I'd suggest making these functions inside the `interface ViewCallback` `default`, the such way it is possible to override only the functions we need.

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherController.kt
```diff
@@ -32,6 +34,7 @@ internal class ActivityWatcherController(
 
     @Mode
     private var currentDialogMode: Int = MODE_NONE
+    //private var permissionType: PermissionType? = null
```
DavDo, line 37: I assume this is leftover

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizer.kt
```diff
@@ -95,7 +97,23 @@ internal class ActivityWatcherForCallVisualizer(
...
+        if (resumedActivity.get() is ComponentActivity) {
+            overlayPermissionLauncher?.run { this.launch(Settings.ACTION_MANAGE_OVERLAY_PERMISSION) }
+        }
+    }
+
+    override fun openOverlayPermissionView() {
+        val activity = resumedActivity.get() ?: return
+        val overlayIntent = Intent(
+            Settings.ACTION_MANAGE_OVERLAY_PERMISSION,
+            Uri.parse("package:" + activity.packageName)
```
DavDo, line 111:
```suggestion
            Uri.parse("package:${activity.packageName}")
```

PR #535, opened 2023-04-03
Mob 2029 customize welcome screen input and progress button

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2029

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/messagecenter/StatefulEditText.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/messagecenter/StatefulEditText.kt
@@ -1,0 +1,134 @@
+package com.glia.widgets.messagecenter
+
+import android.content.Context
+import android.util.AttributeSet
+import androidx.annotation.ColorInt
+import androidx.appcompat.widget.AppCompatEditText
+import com.glia.widgets.R
+import com.glia.widgets.view.unifiedui.extensions.*
+import com.glia.widgets.view.unifiedui.theme.base.ColorTheme
+import com.glia.widgets.view.unifiedui.theme.base.LayerTheme
+import com.glia.widgets.view.unifiedui.theme.base.TextInputTheme
+import com.glia.widgets.view.unifiedui.theme.base.TextTheme
+
+class StatefulEditText @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null) :
+    AppCompatEditText(context, attrs) {
+    private val adapterCallback: StatefulWidgetAdapterCallback<State, TextInputTheme> =
+        StatefulWidgetAdapterCallback { updateTheme(it) }
+
+    private val statefulWidgetAdapter: StatefulWidgetAdapter<State, TextInputTheme> by lazy {
+        SimpleStatefulWidgetAdapter(State.ENABLED, adapterCallback)
+    }
+
+    @get:ColorInt
+    private val gliaBaseShadeColor: Int by lazy {
+        getColorCompat(getAttr(R.attr.gliaBaseShadeColor, R.color.glia_base_shade_color))
+    }
+
+    @get:ColorInt
+    private val gliaBaseDarkColor: Int by lazy {
+        getColorCompat(getAttr(R.attr.gliaBaseDarkColor, R.color.glia_base_dark_color))
+    }
+
+    @get:ColorInt
+    private val gliaBrandPrimaryColor: Int by lazy {
+        getColorCompat(getAttr(R.attr.gliaBrandPrimaryColor, R.color.glia_brand_primary_color))
+    }
+
+    @get:ColorInt
+    private val gliaDisabledColor: Int by lazy { getColorCompat(R.color.glia_disable_button_bg) }
+
+    @get:ColorInt
+    private val gliaNegativeColor: Int by lazy {
+        getColorCompat(getAttr(R.attr.gliaSystemNegativeColor, R.color.glia_system_negative_color))
+    }
+
+    enum class State {
+        ENABLED,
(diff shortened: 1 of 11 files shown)


widgetssdk/src/main/java/com/glia/widgets/messagecenter/StateFulEditText.kt
```diff
@@ -0,0 +1,134 @@
...
+import androidx.annotation.ColorInt
+import androidx.appcompat.widget.AppCompatEditText
+import com.glia.widgets.R
+import com.glia.widgets.view.unifiedui.extensions.*
+import com.glia.widgets.view.unifiedui.theme.base.ColorTheme
+import com.glia.widgets.view.unifiedui.theme.base.LayerTheme
+import com.glia.widgets.view.unifiedui.theme.base.TextInputTheme
+import com.glia.widgets.view.unifiedui.theme.base.TextTheme
+
+class StateFulEditText @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null) :
```
AndriiHorishniiMOC, line 14:
```suggestion
class StatefulEditText @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null) :
```

AndriiHorishniiMOC (approved):
LGTM

widgetssdk/src/main/java/com/glia/widgets/messagecenter/StateFulEditText.kt
DavDo: fixed [786bde7](https://github.com/salemove/android-sdk-widgets/pull/535/commits/786bde75e6784474c7ea3a0ae18bd9848097aebe)

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/extensions/Merge.kt
```diff
@@ -28,15 +28,15 @@ import kotlin.reflect.full.primaryConstructor
...
+    return unsafeMerge(other)
 }
 
 // Added to make method above compatible with Java
 internal fun deepMerge(old: UiTheme?, other: UiTheme?): UiTheme? {
     return old deepMerge other
 }
 
-private fun <T : Any?> T.merge(other: T): T {
+internal fun <T : Any?> T.unsafeMerge(other: T): T {
```
Pelkar, line 39: Why unsafeMerge? the name suggests that this should not be used?

widgetssdk/src/main/res/layout/progress_button.xml
```diff
@@ -0,0 +1,50 @@
...
+    <androidx.appcompat.widget.LinearLayoutCompat
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
+        android:gravity="center_vertical"
+        android:importantForAccessibility="no"
+        android:minHeight="48dp"
+        android:orientation="horizontal">
+
+        <com.google.android.material.progressindicator.CircularProgressIndicator
```
Pelkar, line 18: We should possibly be consistent with the loaders we use in app. Visitor code view uses progressBar. We can use CircularProgressIndicator, but then please change the visitorCodeView one as well

widgetssdk/src/main/java/com/glia/widgets/messagecenter/StatefulEditText.kt
```diff
@@ -0,0 +1,134 @@
...
+import androidx.annotation.ColorInt
+import androidx.appcompat.widget.AppCompatEditText
+import com.glia.widgets.R
+import com.glia.widgets.view.unifiedui.extensions.*
+import com.glia.widgets.view.unifiedui.theme.base.ColorTheme
+import com.glia.widgets.view.unifiedui.theme.base.LayerTheme
+import com.glia.widgets.view.unifiedui.theme.base.TextInputTheme
+import com.glia.widgets.view.unifiedui.theme.base.TextTheme
+
+class StatefulEditText @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null) :
```
Pelkar, line 14: NIT personally I prefer dumb views that are being controlled by controller. I think we should have a team discussion on the subject and decide on an approach.

Pelkar (approved):
I might have a lack of context, but it looks largely ok. Left minor comments, but nothing mayor.

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/extensions/Merge.kt
DavDo:
```kotlin
internal inline infix fun <reified T : Any?> T.deepMerge(other: T): T {
    if (!T::class.isData) throw UnsupportedOperationException("Merge supports only data classes")
    return unsafeMerge(other)
}
``` 
deepMerge will throw an exception if it is called on a `non-data class` type(we're using the primary constructor with reflection and the data classes had a primary constructor with at least one argument) and it will prevent unwanted behavior so why it is `safe`, but it is impossible to use inside generic classes/functions if it is not `inline`, so we had to use `unsafe` one, which will not check for data classes.

widgetssdk/src/main/java/com/glia/widgets/messagecenter/StatefulEditText.kt
DavDo: What's the benefit of `controller` for this type of `View`? It only does the `View` specific work(state + theme) and there is no business logic.

gugalo (approved):

DavDo:
!squash

DavDo:
!merge

PR #546, opened 2023-04-12
Added CV specific exception to the permission

Pelkar:
MOB-2060

CallVisualizer does not want record audio permission, but with CallActivity logic it was requested, added an exception to CallVisualizer

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
@@ -1,6 +1,5 @@
 package com.glia.widgets.call;
 
-import android.Manifest;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageManager;
@@ -14,7 +13,6 @@ import androidx.appcompat.app.AppCompatActivity;
 import androidx.core.content.ContextCompat;
 import androidx.core.util.Pair;
 
-import com.glia.androidsdk.Engagement;
 import com.glia.androidsdk.engagement.Survey;
 import com.glia.widgets.GliaWidgets;
 import com.glia.widgets.R;
@@ -24,7 +22,6 @@ import com.glia.widgets.helper.Logger;
 import com.glia.widgets.helper.Utils;
 import com.glia.widgets.survey.SurveyActivity;
 
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.IntStream;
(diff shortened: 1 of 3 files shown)


widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
```diff
@@ -145,9 +146,11 @@ private void startCallWithPermissions() {
...
                 missingPermissions.add(Manifest.permission.CAMERA);
             }
-            if ((configuration.getMediaType() == Engagement.MediaType.VIDEO || configuration.getMediaType() == Engagement.MediaType.AUDIO)
-                    && missingPermission(Manifest.permission.RECORD_AUDIO)) {
-                missingPermissions.add(Manifest.permission.RECORD_AUDIO);
+            if (!Dependencies.getUseCaseFactory().createIsCallVisualizerUseCase().invoke()) {
+                if ((configuration.getMediaType() == Engagement.MediaType.VIDEO || configuration.getMediaType() == Engagement.MediaType.AUDIO)
+                        && missingPermission(Manifest.permission.RECORD_AUDIO)) {
+                    missingPermissions.add(Manifest.permission.RECORD_AUDIO);
+                }
```
AndriiHorishniiMOC, line 153: I prefer to move this to the CallController.
Also, I don't think it is a good idea to get the UseCase there.

andrews-moc: @Pelkar Agree with Andrii. 
et me also ask you, have you understood why the Call screen is opened without permission for the first time but asks permission the second time?

Pelkar: > I prefer to move this to the CallController. Also, I don't think it is a good idea to get the UseCase there.

Good suggestion, will implement the changes

Pelkar: > @Pelkar Agree with Andrii. Let me also ask you, have you understood why the Call screen is opened without permission for the first time but asks permission the second time?

It seems to me that the reason is pretty simple. Upgrade permissions are handled on the core side and core only requests camera with no audio(as it should). The resume that triggers the second dialog however wants to have all permissions a regular call engagement would need, thus audio dialog is asked separately when view resumes outside of upgrade

gugalo (approved):
No additional comments from me

widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
Pelkar: I migrated the missing permission logic to controller. However, because there is no concrete way of communication from view structure to controller and back, the current solution is not as elegant as it could be...

.idea/androidTestResultsUserPreferences.xml
andrews-moc, line 1: I think this file shouldn't be committed.

Pelkar: sorry, will remove. Why aren't all .idea/ files in gitignore?

andrews-moc (approved):

PR #547, opened 2023-04-12
Update secure maven url

DavDo:
Update secure maven url for `core_sdk` and fix `new core_sdk` related issues.

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
@@ -804,7 +804,7 @@ internal class ChatController(
         currentChatItems: MutableList<ChatItem>,
         message: ChatMessage
     ) {
-        if (!message.content.isNullOrEmpty()) {
+        if (message.content.isNotEmpty()) {
             currentChatItems.add(VisitorMessageItem.asHistoryItem(message))
         }
     }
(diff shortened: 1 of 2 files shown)


widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -804,7 +804,7 @@ internal class ChatController(
         currentChatItems: MutableList<ChatItem>,
         message: ChatMessage
     ) {
-        if (!message.content.isNullOrEmpty()) {
+        if (message.content.isNotEmpty()) {
```
AndriiHorishniiMOC, line 807: Are these significant changes? (Making fields nullable or not nullable)

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
DavDo: I think not, because before that they have no strict nullability in Java. Fields mainly turned into the safer/non-nullable type so it is possible just to use them as they are before.

andrews-moc: Hm, I'm also not sure if the change from `isNullOrEmpty()` to just `isNotEmpty()` won't affect anything.
It's okay if you are completely sure.

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
andrews-moc: @DavDo Cool, thank you for the explanation. I understand now and agree with the change. 👍

gugalo (approved):

AndriiHorishniiMOC:
!merge

PR #548, opened 2023-04-12
Added switch for switching between dialog and integrated view visitor…

Pelkar:
… code

MOB-2072

Added a way to switch between a dialog and integrated view for Visitor Code
switch

### DIFF
--- a/app/src/main/res/layout/main_fragment.xml
+++ b/app/src/main/res/layout/main_fragment.xml
@@ -184,6 +184,45 @@
                 app:layout_constraintStart_toStartOf="@+id/start_guideline"
                 app:layout_constraintTop_toBottomOf="@+id/clear_session_button" />
 
+            <androidx.appcompat.widget.SwitchCompat
+                android:id="@+id/visitor_code_switch"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_marginTop="-10dp"
+                app:layout_constraintTop_toBottomOf="@id/visitor_code_button"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintEnd_toEndOf="parent" />
+
+            <TextView
+                android:id="@+id/visitor_code_dialog_label"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                style="@style/Application.Glia.Subtitle2"
+                android:textColor="@color/glia_black_color"
+                android:text="Inside dialog"
+                android:layout_marginEnd="@dimen/glia_medium"
+                app:layout_constraintTop_toTopOf="@id/visitor_code_switch"
+                app:layout_constraintBottom_toBottomOf="@id/visitor_code_switch"
+                app:layout_constraintEnd_toStartOf="@+id/visitor_code_switch" />
+
+            <TextView
+                android:id="@+id/visitor_code_view_label"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/main_fragment.xml
```diff
@@ -184,6 +184,46 @@
                 app:layout_constraintStart_toStartOf="@+id/start_guideline"
                 app:layout_constraintTop_toBottomOf="@+id/clear_session_button" />
 
+            <androidx.constraintlayout.widget.ConstraintLayout
+                android:id="@+id/visitor_code_switch_container"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                app:layout_constraintTop_toBottomOf="@id/visitor_code_button"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintEnd_toEndOf="parent" >
```
gugalo, line 193: NIT: would try to move this container a bit closer to the Visitor Code button. Maybe using negative margin or something else

gugalo (approved):

app/src/main/res/layout/main_fragment.xml
DavDo, line 193: NIT: `ConstraintLayout` provides all necessary mechanisms to not nest `ConstraintLayout`'s to each other. So this container is redundant. You can constraint views inside parent `ConstraintLayout` and use the `Group` widget to work with view groups of views e.g. change visibility etc.

DavDo (approved):

andrews-moc:
I don't have any comments, the implementation looks good.
However, my concern is that the Main screen might become overloaded soon.
@Pelkar , @gugalo , @DavDo , @AndriiHorishniiMOC What do you think regarding moving this setting to the Settings screen?

DavDo:
> I don't have any comments, the implementation looks good. However, my concern is that the Main screen might become overloaded soon. @Pelkar , @gugalo , @DavDo , @AndriiHorishniiMOC What do you think regarding moving this setting to the Settings screen?

Agree.

Pelkar:
> I don't have any comments, the implementation looks good. However, my concern is that the Main screen might become overloaded soon. @Pelkar , @gugalo , @DavDo , @AndriiHorishniiMOC What do you think regarding moving this setting to the Settings screen?

To me, it would look out of place, especially with the current implementation. What I mean by that is that Settings screen is a place for permanent changes, this however is not stored anywhere and will reset with relaunch of the application.

So either we would need to make the selection persist or keep it in the main fragment.

Essentially it is the teams choice

DavDo:
> > I don't have any comments, the implementation looks good. However, my concern is that the Main screen might become overloaded soon. @Pelkar , @gugalo , @DavDo , @AndriiHorishniiMOC What do you think regarding moving this setting to the Settings screen?
> 
> To me, it would look out of place, especially with the current implementation. What I mean by that is that Settings screen is a place for permanent changes, this however is not stored anywhere and will reset with relaunch of the application.
> 
> So either we would need to make the selection persist or keep it in the main fragment.
> 
> Essentially it is the teams choice

Sounds reasonable to me. Maybe in this case it makes sense to let it as it is and later, try to redesign/refactor this screen(e.g move the end engagement button to appBar, place similar items next to each other, move some rarely used items(clear session) to menu). It will also lead to changes in instrumentation tests.

Pelkar:
If @andrews-moc is ok with it, I would go ahead with the merge. We can do the redesign at some point in the future where we can rethink the whole view from the ground up if need be.

andrews-moc:
Okay, agree ✅

PR #551, opened 2023-04-13
Finish Chat screen when navigating back to Call screen

andrews-moc:
[MOB-2055](https://glia.atlassian.net/browse/MOB-2055)

Steps:

1. start audio engagement (Call screen is opened)
2. Tap "Chat" icon (Chat screen is opened)
3. Tap "back" icon from application (Call screen is opened)
4. Tap "back" icon from application

**Expected result:** Call screen is closed, Main screen is opened
**Actual result:** Chat screen is opened again (loops back several times)

PR fixes this.

Besides, there was (and still is) a difference between application "back" and device "back" behavior on the 3rd step.
Application "back" closes Chat screen and navigates to Call screen. I would say it's expected.
Device "back" closes Chat screen and navigates to Main screen. 
I haven't found a way to align the behaviors, unfortunately.
If you have an idea - I would appreciate it.
Anyway, right now behavior is improved (primary goal of the ticket is achieved).

[MOB-2055]: https://glia.atlassian.net/browse/MOB-2055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/ChatFragment.java
+++ b/app/src/main/java/com/glia/exampleapp/ChatFragment.java
@@ -21,17 +21,7 @@ public class ChatFragment extends Fragment {
 
     private NavController navController;
     private ChatView chatView;
-    private ChatView.OnBackClickedListener onBackClickedListener = () -> {
-        chatView.backPressed();
-        navController.popBackStack();
-    };
     private ChatView.OnEndListener onEndListener = () -> navController.popBackStack();
-    private final NavController.OnDestinationChangedListener onDestinationChangedListener =
-            (controller, destination, arguments) -> {
-                if (destination.getId() == R.id.main_fragment && chatView != null) {
-                    chatView.backPressed();
-                }
-            };
 
     @Nullable
     @Override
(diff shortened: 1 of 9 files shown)


DavDo (approved):

Pelkar:
Correct me if I am wrong, but ChatActivity:

```
@Override
    public void onBackPressed() { if (chatView.backPressed()) super.onBackPressed(); }
```

And

```
private ChatView.OnBackClickedListener onBackClickedListener = () -> {
        if (chatView.backPressed()) finish();
    };
```
Shouldn't these be the same if we want same behaviour?

andrews-moc:
Back press logic on Chat screen is different from Call screen.

Chat screen has also different logic for device "back" and application "back" button, not sure if there were requirements for that.

DavDo:
Maybe I'm mistaken but can't we handle the system back(return false to block system back) and call the same method we call when the app back button is called?
Also as I remember `public void onBackPressed()` should be deprecated and replaced with `onBackPressedDispatcher`

AndriiHorishniiMOC (approved):

Pelkar (approved):
It will work, but still, I do believe at some point we will need to unify the system back and manual back on all views. Can you please create a task for that?

gugalo (approved):
No comments from me.
But I would check what David wrote )

andrews-moc:
> No comments from me. But I would check what David wrote )

@gugalo I'm working on this, no worries. 😉

andrews-moc:
@DavDo @Pelkar @gugalo @AndriiHorishniiMOC My attempt to handle the system back the same way as AppBar back in `ChatView` ends up to the following changes: a7a811cc90c06d38daccb7ab03786e8a2188d64b
Changes overview:
- rethought all "back handling" logic, came to conclusion that it can be simplified a lot
- renamed some functions to understand the logic better (can restore some of the names now though)
- moved all business logic to Controller
- collected all "back handling" logic in a single place (controller)

gugalo (approved):
Nice 👍

Pelkar (approved):
LGTM 👍

andrews-moc:
!squash

andrews-moc:
!merge

PR #564, opened 2023-04-24
Fixed persisting dialog

Pelkar:
MOB-2089

Created a new separation between dismissing dialog and dismissing dialog in backstack

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizer.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizer.kt
@@ -195,10 +195,14 @@ internal class ActivityWatcherForCallVisualizer(
         }
     }
 
-    override fun dismissAlertDialog(manualDismiss: Boolean) {
+    override fun dismissAlertDialog() {
         Logger.d(TAG, "Dismiss alert dialog")
         alertDialog?.dismiss()
         alertDialog = null
+    }
+
+    override fun removeDialogFromStack() {
+        dialogController.dismissCurrentDialog()
     }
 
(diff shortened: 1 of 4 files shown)


widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizer.kt
```diff
@@ -195,12 +195,16 @@ internal class ActivityWatcherForCallVisualizer(
...
-    override fun dismissAlertDialog(manualDismiss: Boolean) {
+    override fun dismissAlertDialog() {
         Logger.d(TAG, "Dismiss alert dialog")
         alertDialog?.dismiss()
         alertDialog = null
     }
 
+    override fun removeDialogFromStack() {
+        dialogController.dismissCurrentDialog()
+    }
```
gugalo, line 206: It is very confusing to me when which method should be called. Could you please rename methods or write comments in the code?

But just for me to understand:
`dismissAlertDialog` - just removes the current on-screen dialog if has any, correct?
`removeDialogFromStack` - will in the end call the `dismissAlertDialog` but through a proper workflow `DialogController.dismissCurrentDialog()` -> `DialogManager.dismissCurrent()` -> skiped few steps -> this classes `dismissAlertDialog`.

Pelkar: dismissAlertDialog should not called from anywhere except onDialogControllerCallback.

For everyday use you would use removeDialogFromBackstack, because it will cause a dialog state change and call onDialogControllerCallback down the line

andrews-moc (approved):

DavDo (approved):

PR #572, opened 2023-04-25
Universal logging tag

gugalo:
TAG object is used for logging all type of information and at the moment every class that needs it needs to create its own companion object or property with such name

Additionally migrated the Logger class to Kotlin

There are still some TAG definitions left in Kotlin. I didn't replace them because they were different from the class name.

**Screenshots:**

Logs example before the update
image

Logs example after update:
image

P.S: Most logs in the example are not affected by this change but still there is basically no difference even for the effected logs


DavDo (approved):

andrews-moc (approved):

Pelkar (approved):
LGTM 👍

gugalo:
!merge

PR #573, opened 2023-04-25
Open the chat transcript screen by clicking on push notification

AndriiHorishniiMOC:
Open the chat transcript screen by clicking on push notification

MOB-1749


Pelkar (commented):
Do I understand correctly that the integrator must handle the push notification on their own? Does it not make sense to handle it ourselves?

DavDo (approved):

AndriiHorishniiMOC:
@Pelkar Yes, they should handle it on their own. But if you have an idea how we can handle it inside the SDK we can make a meeting and discuss it.

Pelkar (approved):

Pelkar:
> @Pelkar Yes, they should handle it on their own. But if you have an idea how we can handle it inside the SDK we can make a meeting and discuss it.

Thank you for the clarification

andrews-moc (approved):
I don't have any comments.

AndriiHorishniiMOC:
!merge

PR #576, opened 2023-04-27
Dismiss screen sharing notification if visitor rejects system permission

andrews-moc:
Changes overview:
- Dismiss screen sharing notification if visitor rejects system permission
- Fix broken mechanism of requesting permissions through CallVisualizerSupportActivity
- Expand this mechanism to be able to use for requesting screen-sharing permission for a non ComponentActivity

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerSupportActivity.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerSupportActivity.kt
@@ -1,5 +1,17 @@
 package com.glia.widgets.callvisualizer
 
+import android.os.Parcelable
 import androidx.appcompat.app.AppCompatActivity
+import com.glia.widgets.callvisualizer.controller.CallVisualizerController
+import kotlinx.parcelize.Parcelize
 
-class CallVisualizerSupportActivity : AppCompatActivity()
+class CallVisualizerSupportActivity : AppCompatActivity() {
+    companion object {
+        val PERMISSION_TYPE_TAG: String = CallVisualizerController::class.java.simpleName
+    }
+}
+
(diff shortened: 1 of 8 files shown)


widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerSupportActivity.kt
```diff
@@ -1,5 +1,17 @@
...
+import android.os.Parcelable
 import androidx.appcompat.app.AppCompatActivity
+import com.glia.widgets.callvisualizer.controller.CallVisualizerController
+import kotlinx.parcelize.Parcelize
 
-class CallVisualizerSupportActivity : AppCompatActivity()
\ No newline at end of file
+class CallVisualizerSupportActivity : AppCompatActivity() {
+    companion object {
+        val PERMISSION_TYPE_TAG: String = CallVisualizerController::class.java.simpleName
```
DavDo, line 10: NIT: this should be `const`. With `const` it is possible to use without pointing to `Companion`

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerSupportActivity.kt
andrews-moc: We figured out that it's not possible for this case:

Screenshot 2023-04-28 at 12 44 26

DavDo: My bad

AndriiHorishniiMOC (approved):
LGTM!

andrews-moc:
!merge

PR #578, opened 2023-05-02
The new message divider is placed at the wrong index

DavDo:
Previously the index was calculated a step earlier, without taking into account that one message could contain multiple attachments.
MOB 2139

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/domain/AddNewMessagesDividerUseCase.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/domain/AddNewMessagesDividerUseCase.kt
@@ -1,0 +1,19 @@
+package com.glia.widgets.chat.domain
+
+import com.glia.widgets.chat.model.history.ChatItem
+import com.glia.widgets.chat.model.history.NewMessagesItem
+
+internal class AddNewMessagesDividerUseCase(
+    private val findNewMessagesDividerIndexUseCase: FindNewMessagesDividerIndexUseCase
+) {
+    operator fun invoke(messages: MutableList<ChatItem>, unreadMessagesCount: Int): Boolean {
+        val index = findNewMessagesDividerIndexUseCase(messages, unreadMessagesCount)
+
+        if (index != NOT_PROVIDED) {
+            messages.add(index, NewMessagesItem)
+            return true
+        }
+
+        return false
+    }
+}
--- a/widgetssdk/src/main/java/com/glia/widgets/core/engagement/domain/model/ChatHistoryResponse.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/engagement/domain/model/ChatHistoryResponse.kt
@@ -1,6 +1,8 @@
 package com.glia.widgets.core.engagement.domain.model
 
+import com.glia.widgets.core.secureconversations.domain.NO_UNREAD_MESSAGES
+
 internal data class ChatHistoryResponse(
     val items: List<ChatMessageInternal>,
-    val newMessagesDividerIndex: Int = -1
+    val newMessagesCount: Int = NO_UNREAD_MESSAGES
 )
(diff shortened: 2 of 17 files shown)


AndriiHorishniiMOC (approved):
Wow, a lot of changes to just fix!

widgetssdk/src/main/java/com/glia/widgets/chat/domain/AddNewMessagesDividerUseCase.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.glia.widgets.chat.model.history.NewMessagesItem
+import com.glia.widgets.core.secureconversations.domain.NO_UNREAD_MESSAGES
+
+internal class AddNewMessagesDividerUseCase(
+    private val findNewMessagesDividerIndexUseCase: FindNewMessagesDividerIndexUseCase
+) {
+    operator fun invoke(messages: MutableList<ChatItem>, unreadMessagesCount: Int): Boolean {
+        val index = findNewMessagesDividerIndexUseCase(messages, unreadMessagesCount)
+
+        if (index != NO_UNREAD_MESSAGES) {
```
andrews-moc, line 13: I'd like to suggest improving the readability. I think there are too many negative statements here: !=, "No" and "Unread". There is an opinion that too many negative statements complicate human understanding. Besides, it looks a bit strange that the index is compared to a variable with a boolean-like name. I would like to suggest renaming `NO_UNREAD_MESSAGES` to `INDEX_OUT_OF_RANGE` or something like this.

widgetssdk/src/main/java/com/glia/widgets/core/engagement/domain/model/ChatHistoryResponse.kt
```diff
@@ -2,5 +2,5 @@ package com.glia.widgets.core.engagement.domain.model
 
 internal data class ChatHistoryResponse(
     val items: List<ChatMessageInternal>,
-    val newMessagesDividerIndex: Int = -1
+    val newMessagesCount: Int = -1
```
andrews-moc, line 5: To me, it looks a bit strange that the count is negative. I would expect to see 0 instead.

widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/GetUnreadMessagesCountWithTimeoutUseCase.kt
```diff
@@ -18,14 +18,16 @@ private const val TAG = "GetUnreadMessagesCountUseCase"
 @VisibleForTesting
 const val TIMEOUT_SEC = 3L
 
+internal const val NO_UNREAD_MESSAGES = -1
```
andrews-moc, line 21: It should look more logical to me if the value of `NO_UNREAD_MESSAGES` was `0` instead of `-1`. Is it possible?

widgetssdk/src/main/java/com/glia/widgets/chat/domain/AddNewMessagesDividerUseCase.kt
DavDo: Agree with you! I also don't like this name but didn't find any valuable.
There is a case when the index is not provided, and the name should reflect this also, so what do you think about `NOT_PROVIDED`?

widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/GetUnreadMessagesCountWithTimeoutUseCase.kt
DavDo: fixed [d9a95cf](https://github.com/salemove/android-sdk-widgets/pull/578/commits/d9a95cf17ddc59bf167997cda61337a5ead43dd6)

widgetssdk/src/main/java/com/glia/widgets/core/engagement/domain/model/ChatHistoryResponse.kt
DavDo: fixed [d9a95cf](https://github.com/salemove/android-sdk-widgets/pull/578/commits/d9a95cf17ddc59bf167997cda61337a5ead43dd6)

widgetssdk/src/main/java/com/glia/widgets/chat/domain/AddNewMessagesDividerUseCase.kt
DavDo: fixed [d9a95cf](https://github.com/salemove/android-sdk-widgets/pull/578/commits/d9a95cf17ddc59bf167997cda61337a5ead43dd6)

Pelkar (approved):
LGTM

DavDo:
!merge

PR #593, opened 2023-05-10
Fix Activity onPause/onResume breaking the permission dialog

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2171

**Additional info:**
The issue was caused by the DialogController addCallback() method, which always fired the current dialog state. 

CallActivity and ChatActivity always add/remove the callback on Activity onResume/onPause to prevent the same dialog from appearing on both Activities. But as a result, because of this, a duplicate second dialog was being with Activity having no callbacks attached to it.

Ideally, need the full refactoring of the Alert dialogs logic in Widgets SDK. Asked this refactoring request to be added to the Android tech depts.

**Screenshots:**


DavDo (approved):
LGTM

Pelkar (approved):
LGTM

gugalo:
!merge

PR #594, opened 2023-05-10
Refactoring

DavDo:
- GliaSdkConfiguration improvements
- Logger improvements
- Utils improvements
- Operator name formatter
- Elapsed time formatter
- Keyboard management
- Get Activity from the Context
- All extensions moved to CommonExtensions.kt file, except Extensions directly related to Unified Ui.
- Turn Utils.getFullHybridTheme to extension function
- Move File related functions to Files.kt

MOB 2178


Pelkar (approved):
LGTM, Thank you for removing wildcard imports where possible

widgetssdk/src/main/java/com/glia/widgets/helper/CommonExtensions.kt
andrews-moc, line 1: It seems to me that "CommonExtensions" is a pretty broad term.
I'm afraid that It will be challenging to keep in mind what functionality is available through extensions.

Maybe it makes sense to split into several standalone Extension classes? For instance, `ThemeExtensions`, `ViewExtensions`, `OperatorExtensions`?

If your idea is to keep everything in a single extension class then let's leave it as you suggest.

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/Merge.kt
```diff
@@ -31,11 +30,6 @@ internal inline infix fun <reified T : Any?> T.deepMerge(other: T): T {
     return unsafeMerge(other)
 }
 
-// Added to make method above compatible with Java
-internal fun deepMerge(old: UiTheme?, other: UiTheme?): UiTheme? {
```
andrews-moc, line 35: Just to double-check, `deepMerge()` is not needed anymore?

widgetssdk/src/main/java/com/glia/widgets/helper/CommonExtensions.kt
DavDo: > It seems to me that "CommonExtensions" is a pretty broad term. I'm afraid that It will be challenging to keep in mind what functionality is available through extensions.
> 
> Maybe it makes sense to split into several standalone Extension classes? For instance, `ThemeExtensions`, `ViewExtensions`, `OperatorExtensions`?
> 
> If your idea is to keep everything in a single extension class then let's leave it as you suggest.

We also had a discussion with @Pelkar about this.
I think the best way could be:
- Keep `widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/ViewExtensions.kt` (I left there only UnifiedUi-related extensions)
- Move all Context-Related extensions to `widgetssdk/src/main/java/com/glia/widgets/helper/ContextExtensions.kt`
- Move all View-Related extensions to `widgetssdk/src/main/java/com/glia/widgets/helper/ViewExtensions.kt`
- Keep remaining extensions in `widgetssdk/src/main/java/com/glia/widgets/helper/CommonExtensions.kt`

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/Merge.kt
DavDo: Yep, it was created for one only Java method, which I converted to kotlin

andrews-moc (approved):
@DavDo You've done a good deal of work 💪

widgetssdk/src/main/java/com/glia/widgets/helper/CommonExtensions.kt
DavDo: > Yes! And maybe create `UnifiedUiExtensions` and move UnifiedUi-related extensions there? 🤔 🙂

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/ViewExtensions.kt currently contains only UnifiedUi extensions, I thought the package name could also refer to unified UI, but we can rename file to `UnifiedUiExtensions`

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/helper/CommonExtensions.kt
andrews-moc: Yes, I would go with renaming it to `UnifiedUiExtensions`.

DavDo: Done ✅

PR #599, opened 2023-05-12
Show the welcome message after the operator joined header

AndriiHorishniiMOC:
Fixed to show the welcome message after the operator joined header.

Used unangagement messages only to change the visitor messages with attachments and to display system messages.
After engagement started:
- unsubscribe from the unangagement messages;
- load the history to get messages in correct order;
- start listening to new messages.

[MOB-2182](https://glia.atlassian.net/browse/MOB-2182)


DavDo (approved):
LGTM

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -409,7 +409,7 @@ internal class ChatController(
                 items.removeAll { (it as? VisitorAttachmentItem)?.messageId == message.id }
                 addVisitorAttachmentItemsToChatItems(items, message, currentMessageIndex + 1)
                 return@emitChatItems chatState.changeItems(items)
-            } else {
+            } else if (message.senderType == Chat.Participant.SYSTEM) {
```
andrews-moc, line 412: Two things seem unlogical here to me:
1. I would expect to see the `if (message.senderType == Chat.Participant.SYSTEM) ` check at the beginning of `onUnengagementMessage()` function. I expect this because the name of the function makes me think that everything that is inside the function should be related to UnengagementMessage. And `UnengagementMessage` is a message with `Chat.Participant.SYSTEM`, isn't it (maybe I'm mistaken in this statement)?
2. Do we still need an `else` case or no need to do anything if none of conditions are true?

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -1553,7 +1560,11 @@ internal class ChatController(
 
     override fun newEngagementLoaded(engagement: OmnicoreEngagement) {
         Logger.d(TAG, "newEngagementLoaded")
-        subscribeToMessages()
+
+        // Loading chat history again on engagement start in case it was an-authenticated visitor that restored ongoing engagement
```
andrews-moc, line 1564: `an-authenticated` => `un-authenticated` or `de-authenticated`
But it seems to me that you mean `re-authenticated`:
`... in case it the visitor was re-authenticated and ongoing engagement is being restored.`

Besides, I would add step by step description of this case because it's still not clear why we need something special in addition to standard load chat history flow.

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -1436,10 +1436,17 @@ internal class ChatController(
         viewCallback?.smoothScrollToBottom()
     }
 
-    private fun loadChatHistory() {
+    private fun loadChatHistory(subscribeToMessages: Boolean = false) {
```
andrews-moc, line 1439: NIT, renaming a bit `subscribeToMessages` to  `subscribeToNewMessages` might be better for understanding.
But, it's not critical, just my opinion.

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
AndriiHorishniiMOC: 1. Unengagement Messages are all messages that come before an engagement started. We need it to display system messages and visitor attachments. But sometimes the engagement started event comes after the welcome message. It this case the operator message will come in the Unengagement Messages callback. To avoid adding this message in wrong order the condition was added.
2. No, we don't need to do anything else.

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -1553,7 +1560,11 @@ internal class ChatController(
 
     override fun newEngagementLoaded(engagement: OmnicoreEngagement) {
         Logger.d(TAG, "newEngagementLoaded")
-        subscribeToMessages()
+
+        // Loading chat history again on engagement start in case it was an-authenticated visitor that restored ongoing engagement
+        // Currently there is no direct way to know if Visitor is authenticated.
```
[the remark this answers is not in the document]
AndriiHorishniiMOC: This comment comes from previous place. I'd better remove it, because now it doesn't have any sense.

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
andrews-moc: Okay, got it.
Looks like I have an idea for a better name for `UnengagementMessage` - `PreengagementMessage`. 🙂 
However, it's too late, I know.

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
AndriiHorishniiMOC: New comments were added.

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
AndriiHorishniiMOC: @andrews-moc You're a genious!
I will rename yem!

gugalo (approved):

Pelkar (approved):
LGTM

AndriiHorishniiMOC:
Closed because it was decided to fix this from backend.

PR #603, opened 2023-05-16
Attachment loader should become blue for "Ready to send" attachment

DavDo:
MOB 2191

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/adapter/UploadAttachmentAdapter.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/adapter/UploadAttachmentAdapter.kt
@@ -246,16 +246,19 @@ class ViewHolder(
                 progressValue = 0,
                 inDeterminate = true
             )
+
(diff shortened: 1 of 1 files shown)


andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/UploadAttachmentAdapter.kt
```diff
@@ -271,9 +274,9 @@ class ViewHolder(
         inDeterminate: Boolean = true
     ) {
         progressIndicator.apply {
+            isIndeterminate = inDeterminate
```
Pelkar, line 277: Does only changing the order of operations fix the issue?

DavDo: Yes, with the old order, the view didn't get `invalidated`, and as a result, the progress which was set before changing mode to `indeterminate=false` didn’t get applied to the view.

PR #605, opened 2023-05-17
Fix wrong image preview

AndriiHorishniiMOC:
Fixed wrong image preview.
Prevent opening an image unless it is ready for preview (same behavior as on iOS).
Also, refactored work with files.

[MOB-2190](https://glia.atlassian.net/browse/MOB-2190)

[MOB-2190]: https://glia.atlassian.net/browse/MOB-2190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
@@ -101,6 +101,7 @@ import com.glia.widgets.core.survey.OnSurveyListener
 import com.glia.widgets.core.survey.domain.GliaSurveyUseCase
 import com.glia.widgets.di.Dependencies
 import com.glia.widgets.filepreview.domain.usecase.DownloadFileUseCase
+import com.glia.widgets.filepreview.domain.usecase.IsFileReadyForPreviewUseCase
 import com.glia.widgets.helper.Logger
 import com.glia.widgets.helper.TAG
 import com.glia.widgets.helper.TimeCounter
@@ -167,7 +168,8 @@ internal class ChatController(
     private val setPendingSurveyUsedUseCase: SetPendingSurveyUsedUseCase,
     private val isCallVisualizerUseCase: IsCallVisualizerUseCase,
     private val preEngagementMessageUseCase: PreEngagementMessageUseCase,
-    private val addNewMessagesDividerUseCase: AddNewMessagesDividerUseCase
+    private val addNewMessagesDividerUseCase: AddNewMessagesDividerUseCase,
+    private val isFileReadyForPreviewUseCase: IsFileReadyForPreviewUseCase
 ) : GliaOnEngagementUseCase.Listener, GliaOnEngagementEndUseCase.Listener, OnSurveyListener {
     private var backClickedListener: ChatView.OnBackClickedListener? = null
     private var viewCallback: ChatViewCallback? = null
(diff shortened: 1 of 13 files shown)


Pelkar (approved):
LGTM

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -359,6 +361,12 @@ internal class ChatController(
         mediaUpgradeOfferRepositoryCallback?.let { removeMediaUpgradeCallbackUseCase(it) }
     }
 
+    fun onImageItemClick(item: AttachmentFile) {
+        if (isFileReadyForPreviewUseCase(item)) {
```
DavDo, line 365: I'd suggest showing some `Toast` in a case when the file is not yet ready, to provide some feedback to the user.

widgetssdk/src/main/java/com/glia/widgets/filepreview/domain/usecase/IsFileReadyForPreviewUseCase.kt
```diff
@@ -0,0 +1,11 @@
+package com.glia.widgets.filepreview.domain.usecase
+
+import com.glia.androidsdk.chat.AttachmentFile
```
DavDo, line 3: can you please make this class internal

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/helper/Files.kt
```diff
@@ -81,40 +87,49 @@ internal fun fixCapturedPhotoRotation(uri: Uri, context: Context) {
...
-            downloadsContentUri,
-            projection,
-            selection,
-            selectionArgs,
-            sortOrder
-        )?.use {
-            if (it.count == 0) return@use Uri.EMPTY
-
+@RequiresApi(Build.VERSION_CODES.Q)
+private fun getContentUriAPI29(fileName: String, context: Context): Uri? {
```
andrews-moc, line 91: NIT, not critical at all, related to style guide.
I think the name should be `getContentUriApi29()`

AndriiHorishniiMOC: fixed c564b40cc7d5f30ec018a0225b08b5e33b5ef236

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
AndriiHorishniiMOC: fixed 45537a8b2ebe0acb61caa153d8e805ed7723169b

widgetssdk/src/main/java/com/glia/widgets/filepreview/domain/usecase/IsFileReadyForPreviewUseCase.kt
AndriiHorishniiMOC: fixed 45537a8b2ebe0acb61caa153d8e805ed7723169b

DavDo (approved):
LGTM

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
andrews-moc: Ideally, I think we need to show some placeholder until preview is available. I think so because the space is currently empty until preview is loaded. From a visitor point of view, it looks like the image hasn't been sent or disappeared.
If you agree with this point, I can raise this question in Slack and create a separate ticket for later (if we agree to make changes).

DavDo: I agree, I even would suggest having a `shimmer effect` as a placeholder, which will indicate that image loading is in progress. But I think it should be implemented for both platforms.

1_Tu5jffgXlt_aP-uVPKqKng

AndriiHorishniiMOC:
!squash

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
AndriiHorishniiMOC: Agree with you. I will rise the topic in slack.

PR #606, opened 2023-05-17
Clear survey and engagement state when clearing visitor session

andrews-moc:
[MOB-2186](https://glia.atlassian.net/browse/MOB-2186)

Changes overview:
- reset survey from `clearVisitorSession()`
- close bubble when `clearVisitorSession()` is called
- unsubscribe engagement end listeners for a bubble when `clearVisitorSession()` is called
- do not destroy bubble controllers on chat closing
- rename `engagementEnded()` to `callVisualizerEngagementEnded()` for Call Visualizer to have separate handling
- amend `onSurveyLoaded()` logic in `ChatController`

Check list for manual testing is [here](https://glia.atlassian.net/wiki/spaces/ENG/pages/3360489599/End+engagement+and+clear+visitor+session+flows).

[MOB-2186]: https://glia.atlassian.net/browse/MOB-2186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
@@ -339,7 +339,7 @@ internal class ChatController(
         synchronized(this) { viewCallback = null }
         backClickedListener = null
         if (!retain) {
-            disposable.dispose()
+            disposable.clear()
             mediaUpgradeOfferRepository.stopAll()
             mediaUpgradeOfferRepositoryCallback = null
             timerStatusListener = null
@@ -554,7 +554,8 @@ internal class ChatController(
             viewCallback?.backToCall()
         } else {
             backClickedListener?.onBackClicked()
-            Dependencies.getControllerFactory().destroyControllers()
+            Dependencies.getControllerFactory().destroyChatController()
+            Dependencies.getControllerFactory().destroyCallController()
         }
         updateFromCallScreenUseCase.updateFromCallScreen(false)
     }
@@ -1612,12 +1613,15 @@ internal class ChatController(
         Logger.d(TAG, "newSurveyLoaded")
         setPendingSurveyUsedUseCase.invoke()
         if (viewCallback != null && survey != null) {
+            // Show survey
             viewCallback!!.navigateToSurvey(survey)
             Dependencies.getControllerFactory().destroyControllers()
         } else if (shouldHandleEndedEngagement && !isVisitorEndEngagement) {
+            // Show "Engagement ended" pop-up
             shouldHandleEndedEngagement = false
             dialogController.showEngagementEndedDialog()
-        } else {
+        } else if (shouldHandleEndedEngagement) {
+            // Close chat screen
             Dependencies.getControllerFactory().destroyControllers()
         }
     }
--- a/widgetssdk/src/main/java/com/glia/widgets/di/UseCaseFactory.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/di/UseCaseFactory.java
@@ -778,4 +778,8 @@ public class UseCaseFactory {
     public ResetSurveyUseCase createResetSurveyUseCase() {
         return new ResetSurveyUseCase(surveyStateManager, repositoryFactory.getGliaSurveyRepository());
     }
+
+    public void resetState() {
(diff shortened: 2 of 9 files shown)


AndriiHorishniiMOC (approved):
LGTM!

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -1602,13 +1603,18 @@ internal class ChatController(
...
         } else if (shouldHandleEndedEngagement && !isVisitorEndEngagement) {
+            // Show "Engagement ended" pop-up
             shouldHandleEndedEngagement = false
             dialogController.showEngagementEndedDialog()
-        } else {
+        } else if (shouldHandleEndedEngagement) {
+            // Close chat screen
             Dependencies.getControllerFactory().destroyControllers()
+        } else {
+            // No need to do anything
```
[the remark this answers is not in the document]
andrews-moc: Removed 1cbf3ea4cfc20f14cdb9c1fa954816f9b2b90ab8

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -544,7 +544,8 @@ internal class ChatController(
             viewCallback?.backToCall()
         } else {
             backClickedListener?.onBackClicked()
-            Dependencies.getControllerFactory().destroyControllers()
+            Dependencies.getControllerFactory().destroyChatController()
+            Dependencies.getControllerFactory().destroyCallController()
```
gugalo, line 548: I am not 100% sure, but possibly you might want to also reset ActivityWatcherForCallVisualizer. Technically it is possible that it is displaying some Dialog in the integrator app when clearSession was called and it should then dismiss any pending dialogs.

widgetssdk/src/main/java/com/glia/widgets/di/UseCaseFactory.java
```diff
@@ -772,4 +772,8 @@ DecodeSampledBitmapFromInputStreamUseCase createDecodeSampledBitmapFromInputStre
     public ResetSurveyUseCase createResetSurveyUseCase() {
         return new ResetSurveyUseCase(surveyStateManager, repositoryFactory.getGliaSurveyRepository());
     }
+
+    public void reset() {
```
DavDo, line 776: NIT: would be better to have a more descriptive name, currently it is not clear what `usecaseFactory.reset()` should reset. Maybe `resetSurveyState`?

DavDo (approved):
LGTM, only one small NIT

widgetssdk/src/main/java/com/glia/widgets/di/UseCaseFactory.java
andrews-moc: @DavDo Thank you. The idea was to have a function where we can add more code that will reset something later on. That's why I used a generic name, not a specific one. Does it make sense to you?

I can still rename it to `resetSurveyState()` because, at this point, this function resets only Survey, you are right. Do you insist on renaming to `resetSurveyState()`?

DavDo: Not at all, but still I think we need to find some descriptive name(e.g. `resetState`), because from the current name I have an impression that it should reset the `UseCaseFactory`

andrews-moc: Agree. I also thought about `resetState()`
Renamed: 79d45217c2dc8c54d47889199e63cdc882b55042

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
andrews-moc: Honestly, I am not sure that resetting anything additionally won't brake something.
I would add it only in case we see the need (e.g. flow that needs it).
I haven't seen such cases so far.
So, if you don't mind, I will leave it as it is now.

andrews-moc:
!squash

andrews-moc:
!merge

PR #607, opened 2023-05-17
Renaming UnengagementMessage to PreEngagementMessage

AndriiHorishniiMOC:
Since we decided not to merge the changes for the order of the welcome message fix, I prepared the commit for the relevant renaming that we decided to do there.

`PreEngagementMessage` way better describes the situation there they use instead of `UnengagementMessage`. Thank @andrews-moc for that advice!


andrews-moc (approved):

gugalo (approved):

DavDo (approved):
LGTM

PR #610, opened 2023-05-22
Fix the file preview screen differs from the design

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2196

I have researched almost all photo viewer apps and they all have black backgrounds and it can be said that this is a common behavior for Android. Previously FilePreviewScreen had a toolbar with our default `brandPrimaryColor` and integrators were not able to configure it(change color). Our FilePreview screen will also look the same for all integrators. I also added an animation to make the transition to FilePreview smoother.

https://github.com/salemove/android-sdk-widgets/assets/19994238/4305f368-40be-4bae-8074-1b2e48268567


AndriiHorishniiMOC (approved):
Good job!

DavDo:
> I like these improvements. As well as migrating a bunch of classes to Kotlin. 👍 💪
> 
> Just have one concern. Taking into account that:
> 
> * we are approaching to the release
> * such kind of big changes add risk to breaking the functionality that has been properly tested,
> 
> may it be better to include these changes to the next release?

I totally agree with you!
But in this particular case, I think it's not so dangerous, because there are a few use cases and it is easy to test.

DavDo:
!squash

PR #611, opened 2023-05-22
Fixed VisitorCodeView not dismissed

Pelkar:
MOB-2211 Prevented opening a new visitor code view when app has an active CallVisualizerEngagement

Thread in slack
https://salemove.slack.com/archives/CHN6UB7UH/p1684759806975669

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/MainFragment.java
+++ b/app/src/main/java/com/glia/exampleapp/MainFragment.java
@@ -480,9 +480,11 @@ public class MainFragment extends Fragment {
     // For testing the integrated Visitor Code solution
     private void showVisitorCodeInADedicatedView() {
         VisitorCodeView visitorCodeView = GliaWidgets.getCallVisualizer().createVisitorCodeView(getContext());
-        CardView cv = containerView.findViewById(R.id.container);
-        cv.addView(visitorCodeView);
-        cv.setVisibility(View.VISIBLE);
+        if (visitorCodeView != null) {
+            CardView cv = containerView.findViewById(R.id.container);
+            cv.addView(visitorCodeView);
+            cv.setVisibility(View.VISIBLE);
+        }
     }
 
     private void showToast(String message) {
--- a/app/src/main/java/com/glia/exampleapp/MainFragment.java
+++ b/app/src/main/java/com/glia/exampleapp/MainFragment.java
@@ -41,6 +41,7 @@ import com.glia.widgets.chat.ChatActivity;
 import com.glia.widgets.chat.ChatType;
 import com.glia.widgets.core.callvisualizer.domain.CallVisualizer;
 import com.glia.widgets.core.configuration.GliaSdkConfiguration;
+import com.glia.widgets.helper.Logger;
 import com.glia.widgets.messagecenter.MessageCenterActivity;
 import com.glia.widgets.view.VisitorCodeView;
 
--- a/widgetssdk/src/main/java/com/glia/widgets/core/callvisualizer/CallVisualizerManager.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/callvisualizer/CallVisualizerManager.java
@@ -4,6 +4,7 @@ import android.content.Context;
 
 import com.glia.androidsdk.GliaException;
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 
 import com.glia.androidsdk.omnibrowse.OmnibrowseEngagement;
 import com.glia.widgets.callvisualizer.CallVisualizerRepository;
@@ -43,6 +44,7 @@ public class CallVisualizerManager implements CallVisualizer {
         }
     }
 
+    @Nullable
     @Override
     public VisitorCodeView createVisitorCodeView(Context context) {
         if (!engagementRepository.isCallVisualizerEngagement()) {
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/glia/exampleapp/MainFragment.java
```diff
@@ -480,9 +480,11 @@ private void showVisitorCode() {
...
     private void showVisitorCodeInADedicatedView() {
         VisitorCodeView visitorCodeView = GliaWidgets.getCallVisualizer().createVisitorCodeView(getContext());
-        CardView cv = containerView.findViewById(R.id.container);
-        cv.addView(visitorCodeView);
-        cv.setVisibility(View.VISIBLE);
+        if (visitorCodeView != null) {
+            CardView cv = containerView.findViewById(R.id.container);
+            cv.addView(visitorCodeView);
+            cv.setVisibility(View.VISIBLE);
+        }
```
gugalo, line 487: I think some info log could be useful here

widgetssdk/src/main/java/com/glia/widgets/core/callvisualizer/CallVisualizerManager.java
```diff
@@ -44,16 +44,22 @@ private void checkForProperInit() {
...
+            return null;
+        }
     }
 
     @Override
     public void showVisitorCodeDialog(Context context) {
-        checkForProperInit();
-        startListeningForEngagements();
-        Dependencies.getControllerFactory().getDialogController().showVisitorCodeDialog();
+        if (!engagementRepository.isCallVisualizerEngagement()) {
```
gugalo, line 58: I think some error/warning log could be useful here

widgetssdk/src/main/java/com/glia/widgets/core/callvisualizer/CallVisualizerManager.java
```diff
@@ -44,16 +44,22 @@ private void checkForProperInit() {
...
-        checkForProperInit();
-        startListeningForEngagements();
-        return buildVisitorCodeUseCase.invoke(context, false);
+        if (!engagementRepository.isCallVisualizerEngagement()) {
+            checkForProperInit();
+            startListeningForEngagements();
+            return buildVisitorCodeUseCase.invoke(context, false);
+        } else {
+            return null;
+        }
```
gugalo, line 53: This is a custom view. I don't think it still should continue working as before. In my opinion this is already the Integrator responsibility to handle case when there is already ongoing engagement

app/src/main/java/com/glia/exampleapp/MainFragment.java
Pelkar: Why? This is the sample application and will not be used in debugging?

widgetssdk/src/main/java/com/glia/widgets/core/callvisualizer/CallVisualizerManager.java
Pelkar: Yes, but we need to decide if we are allowing this approach or not. Returning a null here will avoid the integrator from using our sdk in a way that we do not approve.

widgetssdk/src/main/java/com/glia/widgets/core/callvisualizer/CallVisualizerManager.java
```diff
@@ -44,16 +45,25 @@ private void checkForProperInit() {
 
     @Override
     public VisitorCodeView createVisitorCodeView(Context context) {
```
andrews-moc, line 47: Suggest annotating the return type as `@Nullable`. As well as in the interface.

andrews-moc (commented):
Having looked at the changes, i admit that I would better leave the original case not handled rather than returning `null` instead of the visitor code view. It's solely my opinion. Does my thought make sense to you, @Pelkar, @gugalo, @DavDo, @AndriiHorishniiMOC ?

andrews-moc (commented):
Having looked at the changes, i admit that I would better leave the original case not handled rather than returning `null` instead of the visitor code view. It's solely my opinion. Does my thought make sense to you, @Pelkar, @gugalo, @DavDo, @AndriiHorishniiMOC ?

Pelkar:
> instead

To me the reasoning is the following:

We can either allow the user to init new visitor code or not, there should not be a middle ground. If we allow this, then we will be responsible for any and all repercussions from edge cases. The simplest and safest option is to just not allow edge case causing behaviour in the application. Returning a null when visitor code view can cause unwanted anomalies just seems like the safest solution to me.

andrews-moc:
Okay, I agree with your reasoning.

andrews-moc (approved):

AndriiHorishniiMOC (approved):

app/src/main/java/com/glia/exampleapp/MainFragment.java
gugalo: Yes. That makes sense 👍

widgetssdk/src/main/java/com/glia/widgets/core/callvisualizer/CallVisualizerManager.java
gugalo: I think that this makes things more complicated but does not really prevent the integrator from showing the visitor code during engagement since if he created the custom Visitor Code view before the engagement that view would still continue to get displayed and refreshed. So returning the null does not really add benefit but if you do that you will need to still add extra explanations to the integrator guides saying that this method might also return 'null' and why it does that.

But if you still think this way is better then I just ask you to update the integration guides.

gugalo:
Added my comment in [the thread](https://github.com/salemove/android-sdk-widgets/pull/611#discussion_r1205003414) by I disagree that 'null' solves the problem.

gugalo (approved):

Pelkar:
Closing PR for it is no longer required

PR #614, opened 2023-05-23
Fixed dialog being called from background thread

Pelkar:
MOB-2214

With no notification permission, the dialog was called on background thread, fixed it


gugalo (approved):

andrews-moc (approved):

DavDo (approved):
LGTM

PR #615, opened 2023-05-23
Apply SDK default configs consistently for Call Visualizer and Chat/Call screens

andrews-moc:
[Screen sharing bubble is not displayed until the app passes chat engagement flow](https://glia.atlassian.net/browse/MOB-2216)

What problems do these changes solve?
- If `GliaWidgetsConfig.setUseOverlay()` is not set on Glia init, bubble is not displayed for CallVisualizer screen sharing (wrong)  but is displayed for regular engagements (correct)
- If `GliaWidgetsConfig.setUseOverlay(false)` is set, bubble is not displayed for CallVisualizer screen sharing (correct) but is still displayed for regular engagements (wrong)

The reason for that was the inconsistent usage of default values.


andrews-moc:
!merge

PR #621, opened 2023-05-31
Remove temporary maven repo used for internal Core SDK build

andrews-moc:
We don't need temporary repo anymore because we are going to release


DavDo:
If I'm not mistaken we also need to change the core_sdk version to 1.0.0 manually.

PR #625, opened 2023-06-01
Fixed no survey chat engagement end

Pelkar:
During converting from if statements to when statements it seems that one close condition got modified a little that changed the logic in case of no surveys and manual end engagement on android side.

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
@@ -1638,9 +1638,10 @@ internal class ChatController(
                 dialogController.showEngagementEndedDialog()
             }
 
-            shouldHandleEndedEngagement -> {
+            else -> {
                 // Close chat screen
                 Dependencies.getControllerFactory().destroyControllers()
+                destroyView()
             }
         }
     }


DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -1638,9 +1638,10 @@ internal class ChatController(
                 dialogController.showEngagementEndedDialog()
             }
 
-            shouldHandleEndedEngagement -> {
+            else -> {
                 // Close chat screen
                 Dependencies.getControllerFactory().destroyControllers()
+                viewCallback?.destroyView()
```
andrews-moc, line 1644: It's better to use a dedicated function `destroyView()` I think.

Pelkar: Changed

andrews-moc:
> During converting from if statements to when statements it seems that one close condition got modified a little that changed the logic in case of no surveys and manual end engagement on android side.

@Pelkar I checked Git history and there was no `viewCallback?.destroyView()` inside the `onSurveyLoaded()` function, it seems. So I don't think the reason is identified correctly. That's why I'm concerned if we don't brake anything with this change. Could you elaborate on how you came up with this conclusion?

Pelkar:
> > During converting from if statements to when statements it seems that one close condition got modified a little that changed the logic in case of no surveys and manual end engagement on android side.
> 
> @Pelkar I checked Git history and there was no `viewCallback?.destroyView()` inside the `onSurveyLoaded()` function, it seems. So I don't think the reason is identified correctly. That's why I'm concerned if we don't brake anything with this change. Could you elaborate on how you came up with this conclusion?

There needs to be a case that works outside shouldHandleEndedEngagement, because it is true when we need to show engagement ended dialog. However when the user triggers end engagement there is no need to the dialog. In our case, the when block was entered, but nothing was done about the view being showed and thus an empty chatview remained behind.

The possible cases are:
Has survey -> open survey
Is not user triggered and needs to handle ended engagement and no survey -> show ended dialog
Is user triggered and no survey -> close view

andrews-moc (approved):

Pelkar:
!squash

PR #627, opened 2023-06-01
Fix Dialog Ttitle and Push Notification Permission

DavDo:
Fixed some minor bugs I found during work on big tasks. These are not critical, and not so big to make different PRs for each.

- Moved `android.permission.POST_NOTIFICATIONS` permission declaration from testing app `Manifest` to widget module `Manifest`. After this change, integrators will no longer need to manually declare this permission.

- Modified dialog title for some views to make it responsive. This fixes the bug when the title goes out of its paddings when a title text is long.

- Fixed UnifiedUiConfigurationLoader to make it not block the main thread when loading remote configurations.

**Additional info:**
Will merge after release, that’s the only reason I made it draft

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,8 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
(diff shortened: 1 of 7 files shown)


Pelkar (approved):
LGTM

andrews-moc (approved):

PR #629, opened 2023-06-05
Add correct commit author to Bitrise commits

gersonnoboa:
Bitrise does commits sometimes for us, like for example updating dependencies. Bitrise's documentation recommends to introduce an envvar with the values. These are the same values used in other projects.

MOB-1696

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -4,426 +4,432 @@ project_type: android
 workflows:
   _increment_project_version:
     steps:
-      - script@1:
-          inputs:
-            - content: |-
-                #!/usr/bin/env bash
-                # fail if any commands fails
-                set -e
-                # make pipelines' return status equal the last command to exit with a non-zero status, or zero if all commands exit successfully
-                set -o pipefail
-                # debug log
-                set -x
-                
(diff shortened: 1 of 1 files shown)


gersonnoboa:
https://github.com/salemove/android-sdk-widgets/pull/629/files#diff-61d1ee831cb1df242161a7d23171914a1d5d628c7659b52db957d80a9394bd68R423

DavDo (approved):

andrews-moc:
@gersonnoboa Let me clarify, if I see the actual changes right - `GIT_AUTHOR_NAME` and `GIT_AUTHOR_EMAIL` were added. All the rest is reformatting, right?

andrews-moc (commented):
@gersonnoboa Let me clarify, if I see the actual changes right - GIT_AUTHOR_NAME and GIT_AUTHOR_EMAIL were added. All the rest is reformatting, right?

gugalo (approved):

andrews-moc:
I think we should either change the target branch to `development` or leave `master` as a target but keep in mind to create backport commit to `development`.

gersonnoboa:
@andrews-moc that should be done as part of another PR, so that we can rollback easily in case something goes wrong.

gersonnoboa:
!merge

PR #633, opened 2023-06-12
Hide metadata from TestApp custom card rendering

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2228

**Additional info:**
The issue does not stream from the SDK but from custom chat message rendering that has been created as an example and for testing purposes.

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/ExampleCustomCardAdapter.java
+++ b/app/src/main/java/com/glia/exampleapp/ExampleCustomCardAdapter.java
@@ -23,18 +23,25 @@ import com.glia.widgets.chat.adapter.holder.CustomCardViewHolder;
 import com.glia.widgets.chat.adapter.holder.WebViewViewHolder;
 
 import org.json.JSONException;
+import org.json.JSONObject;
 
 public class ExampleCustomCardAdapter extends CustomCardAdapter {
-    private static final int WEB_VIEW_TYPE = 1;
-    private static final int NATIVE_VIEW_TYPE = 2;
+    private static final Integer WEB_VIEW_TYPE = 1;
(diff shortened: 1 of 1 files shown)


DavDo (approved):

app/src/main/java/com/glia/exampleapp/ExampleCustomCardAdapter.java
```diff
@@ -92,6 +99,16 @@ public NativeViewViewHolder(@NonNull View itemView, @NonNull UiTheme uiTheme) {
             this.okButton = itemView.findViewById(R.id.ok_button);
         }
 
+        public static boolean isNativeViewType(ChatMessage message) {
+            JSONObject metadata = message == null ? null : message.getMetadata();
+            if (metadata == null || metadata.length() == 0) {
+                return false;
+            } else {
+                return metadata.has(NativeViewViewHolder.SHOW_BUTTON_KEY)
```
AndriiHorishniiMOC, line 107: I am not sure that it is for these cases `NativeView` should be shown.
I'd better remove this custom view from our testing app.

AndriiHorishniiMOC (approved):

app/src/main/java/com/glia/exampleapp/ExampleCustomCardAdapter.java
gugalo: I think it is alright. But I can remove it as well if needed. @salemove/tm-mobile-android

AndriiHorishniiMOC: I mean that the `NativeView` is not necessary enough to leave it for such rather strange conditions.

gugalo:
!merge

PR #637, opened 2023-06-16
Use AudioSwitch to handle audio setting during the call

AndriiHorishniiMOC:
Use AudioSwitch to handle audio setting during the call

MOB-2082


DavDo (approved):
LGTM

andrews-moc (approved):

Pelkar (approved):

PR #640, opened 2023-06-22
Fix the survey fetching logic

AndriiHorishniiMOC:
Sometimes the survey was loaded multiple times because the onEngagementEnded called multiple times with delay. So when try to request an engagement again, the survey screen was launched.


andrews-moc (approved):

DavDo (approved):
LGTM

PR #659, opened 2023-07-11
Feature/mob 2451 android caching proof of concept

Pelkar:
The caching implementation should consist of two parts:

1. Shared Preferences that would contain a HashMap of locale codes and time in milliseconds of when the cache was made.
2. Local file cache to store the HashMap as a json in a file that has a the following naming pattern: {locale_code}.locale

In GliaWidgets there will be the locale download cycle, the proof of concept was implemented as part of the GliaInit procedure with the following steps being taken.

1. Sample locale is created serialised into JSON and stored in a cached file in the application sandbox(CacheLocaleUseCase).
2. When this storage succeeds, cached locale names are loaded from shared preferences(CachedFileNamesUseCase), new name is added with the current timestamp and stored back into shared preferences.
3. After storage, next step in the proof of concept is to load the stored data(ReadCachedLocaleUseCase).

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/core/locales/CachedFileNamesUseCase.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/locales/CachedFileNamesUseCase.kt
@@ -1,0 +1,31 @@
+package com.glia.widgets.core.locales
+
+import android.content.SharedPreferences
+import com.glia.widgets.helper.Logger
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
+import io.reactivex.Completable
+import io.reactivex.Single
+
+class CachedFileNamesUseCase(private val sharedPrefs: SharedPreferences, private val gson: Gson) {
+
+    fun store(locale: String): Completable {
+        return retrieve().map { map ->
+            map.remove(locale)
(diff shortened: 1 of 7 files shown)


widgetssdk/src/main/java/com/glia/widgets/core/locales/CachedFileNamesUseCase.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.google.gson.reflect.TypeToken
+import io.reactivex.Completable
+import io.reactivex.Single
+
+class CachedFileNamesUseCase(private val sharedPrefs: SharedPreferences, private val gson: Gson) {
+
+    fun store(locale: String): Completable {
+        return retrieve().map { map ->
+            map.remove(locale)
+            map[locale] = System.currentTimeMillis()
```
DavDo, line 15: The SharedPreferences is a Key-Value storage, and I think in this case it would be better to store locale-time directly in Preferences instead of wrapping it in Map than serializing/deserializing to JSON.
So in this case, the key could be `"com.glia.widgets.core.locales.$locale"`

DavDo (commented):
LGTM, added one suggestion.

PR #669, opened 2023-07-12
- Handle Broadcast events

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2389

Show Toast message — “This action is not currently supported on mobile'.

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
@@ -482,6 +482,10 @@ class ChatView(context: Context, attrs: AttributeSet?, defStyleAttr: Int, defSty
             override fun fileIsNotReadyForPreview() {
                 showToast(context.getString(R.string.glia_view_file_not_ready_for_preview))
             }
+
+            override fun showBroadcastNotSupportedToast() {
+                showToast(context.getString(R.string.gva_not_supported))
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


AndriiHorishniiMOC (approved):
LGTM

EgorovEI:
Project plan says:
On the GVA side, there is a feature flag to show/hide this type of button, and it is and should always be off. So this kind of buttons shouldn’t appear on mobile GVA. But to make sure that in case if they sneak into a mobile GVA, will not break the SDK, it should be rendered, but will **show an error message to the user (“This feature is not supported on mobile”) and log an error (level = info) when clicked.**

I guess you also need to add logging.

gugalo (commented):
No notes from me, but Egor mentioned some missing logging

DavDo:
> Project plan says: On the GVA side, there is a feature flag to show/hide this type of button, and it is and should always be off. So this kind of buttons shouldn’t appear on mobile GVA. But to make sure that in case if they sneak into a mobile GVA, will not break the SDK, it should be rendered, but will **show an error message to the user (“This feature is not supported on mobile”) and log an error (level = info) when clicked.**
> 
> I guess you also need to add logging.

Good point. I will remove the logging part from the project plan. 
When we started working on the project, there was an idea to just silently log the error when this type of button is clicked. 
But with Aksel, we figured out that this could confuse visitors, so we decided to replace the log with a toast message. I think there is no benefit to logging.

DavDo:
!squash

PR #670, opened 2023-07-13
Handle external or deep link when the button is clicked

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2381

### DIFF
--- a/widgetssdk/src/main/AndroidManifest.xml
+++ b/widgetssdk/src/main/AndroidManifest.xml
@@ -6,6 +6,11 @@
             <action android:name="android.intent.action.VIEW" />
             <data android:mimeType="*/*" />
         </intent>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <category android:name="android.intent.category.BROWSABLE" />
+            <data android:scheme="*" />
+        </intent>
     </queries>
 
     <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW" />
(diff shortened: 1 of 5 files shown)


widgetssdk/src/main/AndroidManifest.xml
```diff
@@ -6,6 +6,11 @@
             <action android:name="android.intent.action.VIEW" />
             <data android:mimeType="*/*" />
         </intent>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <category android:name="android.intent.category.BROWSABLE" />
+            <data android:scheme="*" />
+        </intent>
```
gugalo, line 13: To be honest I have little experience with deep links and don't understand how this works. In my experience, the manifest changes should be fully implemented by the integrator side and would look a bit different from these changes. 

Could you explain how this works or to dave time just send a link to related docs.

Pelkar (approved):
LGTM

widgetssdk/src/main/AndroidManifest.xml
DavDo: > If your app targets Android 11 or higher and needs to interact with apps other than the ones that are visible automatically, add the [<queries>](https://developer.android.com/guide/topics/manifest/queries-element) element in your app's manifest file. Within the <queries> element, specify the other apps [by package name](https://developer.android.com/training/package-visibility/declaring#package-name), [by intent signature](https://developer.android.com/training/package-visibility/declaring#intent-filter-signature), or [by provider authority](https://developer.android.com/training/package-visibility/declaring#provider-authority), as described in the following sections.

You can also find details [here](https://developer.android.com/training/package-visibility/declaring)

PR #683, opened 2023-08-01
Implemented new unified strings

Pelkar:
MOB-2346

Created and unified new string keys in a backwards compatible manner. The only exception was made for the plurals key, where the override should be for the individual values, not the plural itself.

In minor places, trimmed down reference links in string values, still some TODO-s left in the strings.xml file but vast majority of them is about ADA witch is yet to be unified and will be done in MOB-2280

Decided it is probably be for the best to go with a feature branch approach for this, because we will probably start breaking things otherwise until main functionality is there. Should not cause many conflicts by the end though.

### DIFF
--- a/widgetssdk/src/main/res/values/donottranslate.xml
+++ b/widgetssdk/src/main/res/values/donottranslate.xml
@@ -2,4 +2,16 @@
 <resources>
     <string name="glia_outlined_options_icon_content_description">Outlined options icon description</string>
     <string name="glia_file_preview_transition_name">image_preview_transition</string>
+
+    <!-- Survey -->
+    <string name="glia_survey_open_text_item_title" translatable="false">Additional thoughts?</string>
+    <string name="glia_survey_scale_item_title" translatable="false">How would you rate this interaction? (1 - Bad, 5 - Good)</string>
+    <string name="glia_survey_scale_item_1" translatable="false">1</string>
+    <string name="glia_survey_scale_item_2" translatable="false">2</string>
+    <string name="glia_survey_scale_item_3" translatable="false">3</string>
+    <string name="glia_survey_scale_item_4" translatable="false">4</string>
+    <string name="glia_survey_scale_item_5" translatable="false">5</string>
+    <string name="glia_survey_single_choice_item_title" translatable="false">If you hadn\'t chatted with us today, what would you have done?</string>
+    <string name="glia_survey_require_label" translatable="false"><![CDATA[%1$s\u00A0<font color=#%2$s>*</font>]]></string>
+
 </resources>
--- a/widgetssdk/src/main/res/values/donottranslate.xml
+++ b/widgetssdk/src/main/res/values/donottranslate.xml
@@ -4,14 +4,11 @@
     <string name="glia_file_preview_transition_name">image_preview_transition</string>
 
     <!-- Survey -->
-    <string name="glia_survey_open_text_item_title" translatable="false">Additional thoughts?</string>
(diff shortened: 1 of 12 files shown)


DavDo (approved):
LGTM

widgetssdk/src/main/res/values/donottranslate.xml
```diff
@@ -2,4 +2,16 @@
...
+
+    <!-- Survey -->
+    <string name="glia_survey_open_text_item_title" translatable="false">Additional thoughts?</string>
+    <string name="glia_survey_scale_item_title" translatable="false">How would you rate this interaction? (1 - Bad, 5 - Good)</string>
+    <string name="glia_survey_scale_item_1" translatable="false">1</string>
+    <string name="glia_survey_scale_item_2" translatable="false">2</string>
+    <string name="glia_survey_scale_item_3" translatable="false">3</string>
+    <string name="glia_survey_scale_item_4" translatable="false">4</string>
+    <string name="glia_survey_scale_item_5" translatable="false">5</string>
+    <string name="glia_survey_single_choice_item_title" translatable="false">If you hadn\'t chatted with us today, what would you have done?</string>
```
andrews-moc, line 14: I think that survey items can or should also be translated (except 1-5 numbers).

gugalo: I think that numbers actually might be also available for translations. Very long ago worked on an app that needed that. 
image
But probably we won't need that anytime soon so maybe it is not important )

gugalo (approved):

widgetssdk/src/main/res/values/donottranslate.xml
gugalo: Probably you can ignore my previous comment

Pelkar: These seem to be some temporary implementation. Waiting for final confirmation from Davit

DavDo: I'm sorry, I forgot to check(((
But at first glance, it looks like these strings are placeholders and should be replaced with text from the server. However, on IOS there are no such texts as far as I'm informed, @Pelkar can confirm.
As far as I remember, @andrews-moc implemented Survey, so he knows better than me.

Pelkar: Yes, iOS has nothing like this

AndriiHorishniiMOC (approved):

widgetssdk/src/main/res/values/donottranslate.xml
andrews-moc: Sorry for a late reply. I double-checked and indeed, looks like these survey strings should not be displayed. They are just like placeholders and there shouldn't be a case when they are displayed.

@AndriiHorishniiMOC Please confirm if you remember because if I'm not mistaken, you've added them initially.

Pelkar: @AndriiHorishniiMOC any input?

AndriiHorishniiMOC: Strings `glia_survey_scale_item_1` ... `glia_survey_scale_item_5` are used as labels for the buttons. In my view, they can be translated.

`glia_survey_open_text_item_title`, `glia_survey_scale_item_title`, `glia_survey_single_choice_item_title` - better to remove them, cause those values are replaced with text from the server.

Pelkar:
!squash

PR #703, opened 2023-08-10
Fixed overlay dialog being dismissed by ENGAGEMENT_ENDED at start of …

Pelkar:
…engagement

END_ENGAGEMENT being called at the beginning of the engagement dismisses overlay dialog and breaks acceptance tests

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
@@ -231,7 +231,7 @@ public class CallController implements
     public void engagementEnded() {
         Logger.d(TAG, "engagementEndedByOperator");
         stop();
-        if (!Glia.getCurrentEngagement().isPresent()) {
+        if (!Dependencies.getUseCaseFactory().createIsOngoingEngagementUseCase().invoke()) {
             dialogController.dismissDialogs();
         }
     }
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
@@ -231,7 +231,9 @@ public class CallController implements
     public void engagementEnded() {
         Logger.d(TAG, "engagementEndedByOperator");
         stop();
-        dialogController.dismissDialogs();
+        if (!Glia.getCurrentEngagement().isPresent()) {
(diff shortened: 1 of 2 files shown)


andrews-moc (approved):

DavDo (commented):
Can use `!isOngoingEngagementUseCase()` instead of calling `Glia.getCurrentEngagement().isPresent`

widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
```diff
@@ -231,7 +231,9 @@ public void onHoldChanged(boolean isOnHold) {
     public void engagementEnded() {
         Logger.d(TAG, "engagementEndedByOperator");
         stop();
-        dialogController.dismissDialogs();
+        if (!Glia.getCurrentEngagement().isPresent()) {
```
AndriiHorishniiMOC, line 234: But what if present? The call screen will be stuck?
Maybe better fix the GliaEngagementStateRepository instead of fixing behavior on its events.

Pelkar: There is a task for investigating anomaly behaviours with engagement states. This is to fix acceptance tests now

widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
```diff
@@ -231,7 +231,9 @@ public void onHoldChanged(boolean isOnHold) {
     public void engagementEnded() {
         Logger.d(TAG, "engagementEndedByOperator");
         stop();
-        dialogController.dismissDialogs();
+        if (!Dependencies.getUseCaseFactory().createIsOngoingEngagementUseCase().invoke()) {
```
DavDo, line 234: `IsOngoingEngagementUseCase` is present in both controllers, you can just call `if(!isOngoingEngagementUseCase())`

Pelkar:
> @Pelkar I looked into the comments and thought more.
> 
> If the "engagement ended" event is fired multiple times, there is a risk that we can close dialogs unpredictably, I think. Maybe we shall: a) try to find another solution b) investigate why the "engagement ended" event is fired multiple times, even when it shouldn't c) leave this bug as is until we find a reliable solution.
> 
> What do you think?

I commented out the solution with the bug fix until investigation is complete

PR #711, opened 2023-08-14
Fix Attachment button is hidden for sc with optimized chat flow

DavDo:
MOB 2530

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
@@ -219,12 +219,7 @@ internal class ChatController(
                 return
             }
 
-            var initChatState = chatState.initChat(companyName, queueId, visitorContextAssetId)
-            if (isSecureEngagement) {
-                initChatState = initChatState.setSecureMessagingState()
-            }
-            prepareChatComponents()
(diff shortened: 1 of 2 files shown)


widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -219,12 +219,7 @@ internal class ChatController(
                 return
             }
 
-            var initChatState = chatState.initChat(companyName, queueId, visitorContextAssetId)
-            if (isSecureEngagement) {
-                initChatState = initChatState.setSecureMessagingState()
-            }
-            prepareChatComponents()
-            emitViewState { initChatState }
```
andrews-moc, line 227: So, there is no need to execute these two lines now, correct?

DavDo: these two lines were moved to the next step (after the chat history loaded) of chat initialization.

AndriiHorishniiMOC (approved):

andrews-moc (approved):

Pelkar (approved):
Approving on the condition that Andrij-s comments answer is "yes"

gugalo (approved):

PR #712, opened 2023-08-14
Bugfix/mob 2523 investigate engagement state

Pelkar:
Initial problem was "End engagement" dialog could persist from one session to another.

The fix for this raised acceptance test issue where ENGAGEMENT_ENDED was called any time that Engagement was not ongoing.

Adding new type to engagement state NO_ENGAGEMENT to clear ambiguity and ENGAGEMENT_ENDED is only sent when engagement ends.

Original issue is https://glia.atlassian.net/browse/MOB-2503

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
@@ -231,12 +231,9 @@ public class CallController implements
     public void engagementEnded() {
         Logger.d(TAG, "engagementEndedByOperator");
         stop();
-        // TODO re-enable during MOB-2523
-        /*
         if (!isOngoingEngagementUseCase.invoke()) {
-            dialogController.dismissDialogs()
+            dialogController.dismissDialogs();
         }
-        */
     }
 
     @Override
--- a/widgetssdk/src/main/java/com/glia/widgets/core/engagement/GliaEngagementStateRepository.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/engagement/GliaEngagementStateRepository.java
@@ -1,6 +1,7 @@
 package com.glia.widgets.core.engagement;
 
 import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
 
 import com.glia.androidsdk.Engagement;
 import com.glia.androidsdk.Operator;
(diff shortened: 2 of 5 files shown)


Pelkar:
Will clear conflicts after initial feedback

DavDo (approved):
LGTM, It's truly a great discovery.

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
```diff
@@ -231,7 +231,9 @@ public void onHoldChanged(boolean isOnHold) {
     public void engagementEnded() {
         Logger.d(TAG, "engagementEndedByOperator");
         stop();
-        dialogController.dismissDialogs();
+        if (!isOngoingEngagementUseCase.invoke()) {
+            dialogController.dismissDialogs();
+        }
```
andrews-moc, line 236: I'm a little confused about git diff. I was expecting to see "TODO re-enable during MOB-2523" here, added by [this pull request](https://github.com/salemove/android-sdk-widgets/pull/703). Is there a chance that you checkout some outdated `development` state?

widgetssdk/src/main/java/com/glia/widgets/core/engagement/GliaEngagementStateRepository.java
```diff
@@ -99,12 +100,16 @@ private EngagementStateEvent mapToEngagementStateChangeEvent(
...
+        if (engagementState == null && isOngoingEngagement) {
+            isOngoingEngagement = false;
             return new EngagementStateEvent.EngagementEndedEvent();
+        } else if (engagementState == null) {
+            return new EngagementStateEvent.NoEngagementEvent();
         } else if (engagementState.getVisitorStatus() == EngagementState.VisitorStatus.TRANSFERRING) {
             return new EngagementStateEvent.EngagementTransferringEvent();
         } else {
             if (operator == null) {
+                isOngoingEngagement = true;
```
andrews-moc, line 112: I am not sure that `isOngoingEngagement = true;` should only be set `if (operator == null)`. Maybe it should be set at line number 111, for the whole else block?

andrews-moc:
I like the idea, overall. Looks reasonable to me.
I would suggest trying to add unit tests because the logic is quite fragile, it seems.

widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
Pelkar: That is interesting that these don't show here, I actually deleted the TODO-s during development

widgetssdk/src/main/java/com/glia/widgets/core/engagement/GliaEngagementStateRepository.java
Pelkar: This looks unreasonable, but `if (operator == null)` triggers `EngagementOperatorConnectedEvent` that seems to be the most logical place to mark engagement as started

Pelkar:
> I like the idea, overall. Looks reasonable to me. I would suggest trying to add unit tests because the logic is quite fragile, it seems.

I will add test coverage

widgetssdk/src/main/java/com/glia/widgets/call/CallController.java
andrews-moc: Git shows some conflicts for this branch. Maybe they are related to this question.

andrews-moc (approved):

PR #715, opened 2023-08-15
Improve chat transcript quick reply button behaviour

DavDo:
MOB 2535

**Additional info:**
We only show quick reply buttons for the latest item in the history. However, completed engagements render the button non-functional and could be outdated for future engagements. Therefore, we need to check if the engagement is ongoing before displaying quick reply buttons.


AndriiHorishniiMOC (approved):
LGTM!

andrews-moc (approved):

PR #720, opened 2023-08-18
Replace PagerSnapHelper to LinearSnapHelper for Gallery list

DavDo:
The purpose of `SnapHelper` is to snap gallery items to the center. `PagerSnapHelper` scrolls one by one, which is not desired. We replaced it with `LinearSnapHelper` for center snapping only.


gugalo (approved):
Can you add some explanation to the commit and the PR of why you replaced it?

AndriiHorishniiMOC (approved):

PR #722, opened 2023-08-18
Improve solution for opening dialogs on top of CallVisualizerSupportActivity instead of integrator's Activity

andrews-moc:
You can check ticket description to find the scenarios, which these changes are to solving:
[Integrators want to use Android Widgets SDK without a need to use Material Theme in their projects](https://glia.atlassian.net/browse/MOB-2515)

There are two scenarios that are not solved yet, but their reason is different and I'm working on them separately.

@DavDo FYI, I decided not to move the logic from `ActivityWatcherForCallVisualizer` to `CallVisualizerSupportActivity` because:
- I'm not completely sure that this logic is only needed for `CallVisualizerSupportActivity`
- I've spent a good deal of time already, on this solution and don't want to spend more
- this change is, actually, refactoring and it doesn't solve the initial bugs


DavDo (approved):
LGTM

AndriiHorishniiMOC (approved):

Pelkar (approved):

gugalo (approved):

PR #719, opened 2023-08-18
Fix 🤖Line breaks are not rendered in GVA content

DavDo:
MOB 2546

PR #723, opened 2023-08-21
Improve line break replacement logic

DavDo:
Improved according to [suggested](https://github.com/salemove/android-sdk-widgets/pull/719#discussion_r1299156114) by @AndriiHorishniiMOC regex.

Here is a detailed explanation of the characters:

- `\r\n` is a line break in Windows.
- `\n` is a line break in Unix (macOS, Linux, etc.).

The regular expression (`\r\n|\n`) is saying that it should match either a Windows line break `\r\n` OR a Unix line break `\n`. Therefore, this is a way of matching a line break in a text file regardless of the operating system where the file was created.


AndriiHorishniiMOC (approved):
LGTM

andrews-moc (approved):
NIT: maybe it makes sense to extend a comment and explain in human words why `\r\n|\n` instead of just `\n`. As far as I understand, this changes is needed to support different platforms, right?

DavDo:
Updated description

andrews-moc:
Thank you. I meant in Android project. 🙂 Maybe it makes sense to mention this in javadoc. Something like `Replaces \n to <br> regardless of the operating system where the file was created.`

andrews-moc:
Cool, thank you. 🤝

DavDo:
!squash

PR #728, opened 2023-08-25
Fix Survey survives GliaWidgets.endEngagement event

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2579

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
@@ -219,12 +219,13 @@ public class GliaWidgets {
      */
     public static void endEngagement() {
         Logger.d(TAG, "endEngagement");
+        Dependencies.getControllerFactory().destroyControllers();
+
         Dependencies.glia().getCurrentEngagement().ifPresent(engagement -> engagement.end(e -> {
             if (e != null) {
                 Logger.e(TAG, "Ending engagement error: " + e);
             }
         }));
-        Dependencies.getControllerFactory().destroyControllers();
     }
 
     /**
--- a/widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
@@ -212,6 +212,7 @@ public class ControllerFactory {
         destroyChatController();
         serviceChatHeadController.onDestroy();
         applicationChatHeadController.onDestroy();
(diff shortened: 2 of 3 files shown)


andrews-moc:
Could you please explain how is `MessageNotSeenHandler` connected with Survey - how was it influencing the Survey?

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -219,12 +219,13 @@ public static void clearVisitorSession() {
...
     public static void endEngagement() {
         Logger.d(TAG, "endEngagement");
+        Dependencies.getControllerFactory().destroyControllers();
+
         Dependencies.glia().getCurrentEngagement().ifPresent(engagement -> engagement.end(e -> {
             if (e != null) {
                 Logger.e(TAG, "Ending engagement error: " + e);
             }
         }));
-        Dependencies.getControllerFactory().destroyControllers();
```
andrews-moc, line 227: Does the sequence of calls matter here? I'm asking because there is a risk that changed sequence can cause other issues which we are not aware of.

Pelkar (commented):
Waiting for clarifications to Andriy-s questions, otherwise looks ok

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
DavDo: I think yes, it matters because destroying controllers before calling the `endEngagement` guarantees that all the `end engagement` subscribers will be unsubscribed before the `end-engagement` event is received, so we will be safe from leaking the survey to the next engagement and so on.
I want to remind you that integrators are using this method to end engagement outside of the Widgets SDK app, so we need to clean all the data(surveys included), because this is usually used when integrators log out users from the app.

DavDo: cc: @Pelkar @AndriiHorishniiMOC for visibility

AndriiHorishniiMOC (approved):
Ok, if iе solves the issue.

andrews-moc (approved):

DavDo:
> Could you please explain how is `MessageNotSeenHandler` connected with Survey - how was it influencing the Survey?

`MessageNotSeenHandler` registers a listener for `EngagementEndUseCase` and then doesn’t unregister them when the engagement ends. This way `EngagementEndUseCase` keeps listening for the engagement end event, and as a result fetches the survey. Which is causing this issue.

PR #732, opened 2023-08-28
Do not unsubscribe from engagement events for Call Visualizer

andrews-moc:
Some listeners subscribe only once in Call Visualizer, so we shouldn't unsubscribe

[SDK should receive and handle subsequent 2-way video and screen-sharing requests](https://glia.atlassian.net/browse/MOB-2556)


DavDo (approved):
LGTM

AndriiHorishniiMOC (approved):

PR #735, opened 2023-08-29
Response caching for authenticated visitors with chat history

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2572

**Description**
 Postback Buttons are Persistent Buttons, which send utterances to the chat when clicked or Quick Replies.
When an authenticated visitor enters the chat with history, which contains a postback button, and clicks one of those buttons, an engagement request should be sent and the button response (SingleChoiceAttachment) should be cached until engagement is established and then sent to the server one by one.
Currently, we have similar functionality for cases when an authenticated visitor with chat sends a message when there is no engagement.


AndriiHorishniiMOC (approved):
LGTM

widgetssdk/src/main/java/com/glia/widgets/chat/ChatManager.kt
```diff
@@ -128,7 +128,7 @@ internal class ChatManager constructor(
 
     @VisibleForTesting
     fun checkUnsentMessages(state: State) {
-        sendUnsentMessagesUseCase(state.unsentItems.firstOrNull()?.message ?: return) {}
+        sendUnsentMessagesUseCase(state.unsentItems.firstOrNull() ?: return) {}
     }
```
andrews-moc, line 132: The name of the functions makes me think that it's possible that several messages can be unsent.

However, as far as I understand, `state.unsentItems.firstOrNull()` will pass only one message to `sendUnsentMessagesUseCase`.

So, my concern is: won't we loose other unsent messages?

DavDo: When a response is successfully sent, the message is received through a socket and removed from the unsent messages list, and the method is called again until the list is empty. This is a kind of recursive method based on socket callback.

andrews-moc (approved):

PR #743, opened 2023-08-31
Improve chat flow to handle sent message responses

DavDo:
Previously, the chat flow only rendered visitor messages from the chat messages socket and ignored responses from the send message endpoint. However, these messages were not rendered during socket reconnection, leading to user confusion.

***Changes:***
- Add functions to handle send message response
- Remove redundant sealed classes from the Visitor message
- Optimize ChatAdapter for delivered status update

MOB 2603

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/ChatManager.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/ChatManager.kt
@@ -125,7 +125,9 @@ internal class ChatManager constructor(
 
     @VisibleForTesting
     fun checkUnsentMessages(state: State) {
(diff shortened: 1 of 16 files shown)


AndriiHorishniiMOC (approved):
LGTM

andrews-moc:
@DavDo Could you please describe the steps how the issue was reproduced? To be sure I clearly understand the scope of the changes.

DavDo:
With the previous implementation, when the user sends some message when the socket is still reconnecting, the sent messages will not be visible until the socket is not reconnected.
With these changes, the sent message will be also posted from the `sendMessage` response, so in the case of socket is reconnecting visitor will see their messages in chat.

Pelkar (approved):

PR #744, opened 2023-09-01
Feature/mob 2563 new strings for notifications

Pelkar:
Added remote string fetching logic to Android notifications.

As this is basically also the first time using the string fetching, if something stands out and should be accessed differently, please give feedback!


DavDo (approved):
LGTM

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/helper/ResourceProvider.java
```diff
@@ -78,6 +80,14 @@ public int convertDpToIntPixel(float dp) {
         return Math.round(convertDpToPixel(dp));
     }
 
+    @Override
+    public String getResourceKey(int stringKey) {
+        /*
+            getResourceName wil contain package and resourceName separated by "/", discarding the first package part
```
andrews-moc, line 86: I think several examples might be useful here. At least for me it's a bit difficult to imagine how it will actually look like because I'm not that deed in the project context.

widgetssdk/src/main/res/values/legacy_strings.xml
```diff
@@ -0,0 +1,195 @@
+<?xml version="1.0" encoding="utf-8"?>
```
andrews-moc, line 1: Is `legacy_strings.xml` file used anywhere? I cannot find any usages for some reason.

widgetssdk/src/main/java/com/glia/widgets/helper/ResourceProvider.java
```diff
@@ -78,6 +80,14 @@ public int convertDpToIntPixel(float dp) {
         return Math.round(convertDpToPixel(dp));
     }
 
+    @Override
+    public String getResourceKey(int stringKey) {
```
andrews-moc, line 84: Does it make sense to add a javadoc for the `getResourceKey(int stringKey)` function? With several examples, maybe.

widgetssdk/src/main/java/com/glia/widgets/di/Dependencies.java
```diff
@@ -42,9 +43,18 @@ public class Dependencies {
...
+    private static StringProvider stringProvider;
 
     public static void onAppCreate(Application application) {
-        notificationManager = new NotificationManager(application);
+        resourceProvider = new ResourceProvider(application.getBaseContext());
+        stringProvider = stringKey -> {
+            // TODO IMPLEMENT NEW METHOD FROM CORE
+            String fallback = resourceProvider.getString(stringKey);
+            String key = resourceProvider.getResourceKey(stringKey);
+            // TODO IMPLEMENT NEW METHOD FROM CORE
```
andrews-moc, line 54: I understand that this is a draft for TODO, just want to make sure my current understanding is correct.
What needs to be done here in non-technical words is:
1. Prepare a string fallback.
2. Prepare a string key for getting string from remote source.
3. Get string from remote source. This step is missing from the draft now, right?

widgetssdk/src/main/java/com/glia/widgets/helper/ResourceProvider.java
Pelkar: We can add it, but it is not a commonly used functionality

widgetssdk/src/main/res/values/legacy_strings.xml
Pelkar: Legacy strings are being referenced by the new strings to avoid problems with overridden strings, but I think we need to rename the files to achieve that safety. Thank you for making me think about this 😄

widgetssdk/src/main/java/com/glia/widgets/di/Dependencies.java
Pelkar: Yes, that is correct. When download task gets merged in SDK, then we should be able to make a temporary release and hook the two features together.

widgetssdk/src/main/java/com/glia/widgets/helper/ResourceProvider.java
andrews-moc: I think some short doc will be useful even for internal usage. Today you remember what function does, in a year you might need time to figure it out, you know. For a new team member it might not be obvious what resource key is and why do we need it in addition to keys from string res file.

andrews-moc (approved):
Comments are not critical, you will handle them based on your vision. I am approving PR.

PR #746, opened 2023-09-01
Fix an app crash when there is no application to open uri

DavDo:
MOB 2608

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
@@ -512,6 +512,7 @@ class ChatView(context: Context, attrs: AttributeSet?, defStyleAttr: Int, defSty
         if (intent.resolveActivity(context.packageManager) != null) {
             context.startActivity(intent)
         } else {
+            Logger.e(TAG, "No email client, uri - $uri")
             showToast(context.getString(R.string.glia_dialog_unexpected_error_title))
         }
     }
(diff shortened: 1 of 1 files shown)


widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
```diff
@@ -528,7 +528,11 @@ class ChatView(context: Context, attrs: AttributeSet?, defStyleAttr: Int, defSty
 
     private fun requestOpenUri(uri: Uri) {
         Intent(Intent.ACTION_VIEW, uri).addCategory(Intent.CATEGORY_BROWSABLE).also {
-            context.startActivity(it)
+            if (it.resolveActivity(context.packageManager) != null) {
+                context.startActivity(it)
+            } else {
+                showToast(context.getString(R.string.glia_dialog_unexpected_error_title))
```
andrews-moc, line 534: I would suggest to ask @alina-kauts to help with composing a specific error message that can give a visitor and integrator a clear understanding what's wrong. Something like "Application that can handle action not found".

DavDo: I can add Logs for developers, but I don't think we need to expose such information to the visitors, that's why I thought to use this common message, but it is also okay to not show anything for visitors in this case.

Pelkar (approved):

DavDo:
!squash

PR #760, opened 2023-09-12
Feature/mob 2462 remote strings chat

Pelkar:
Implemented remote strings for chat, fixed multi argument string fetching


widgetssdk/src/main/java/com/glia/widgets/StringProviderImpl.kt
```diff
@@ -6,8 +6,8 @@ class StringProviderImpl(private val resourceProvider: ResourceProvider): String
 
     override fun getRemoteString(stringKey: Int, vararg values: String): String {
         // TODO IMPLEMENT NEW METHOD FROM CORE
-        val fallback = resourceProvider.getString(stringKey, values)
         val key = resourceProvider.getResourceKey(stringKey)
+        val fallback = resourceProvider.getString(stringKey, *values)
```
AndriiHorishniiMOC, line 10: I don't get the idea of the fallback.

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/StringProviderImpl.kt
Pelkar: of fallback or of the change in the fallback?

Fallback is the native string value that will be used ONLY if no value is present in the remote strings

andrews-moc (approved):

widgetssdk/src/test/snapshots/images/com.glia.widgets.chat_ChatViewQuickReplySnapshotTest_withWithUnifiedTheme.png
AndriiHorishniiMOC, line 1: Initially, the gradient for the app bar title was vertical. Now, it is horizontal.

AndriiHorishniiMOC (changes requested):
Looks like we have issues with Git LFS

.gitattributes
```diff
@@ -1 +1,2 @@
 widgetssdk/src/test/snapshots/**/*.png filter=lfs diff=lfs merge=lfs -text
+*.png filter=lfs diff=lfs merge=lfs -text
```
AndriiHorishniiMOC, line 2: Do not need to add all PNG files with LFS.

AndriiHorishniiMOC (approved):
Good work!

PR #775, opened 2023-09-19
Feature/mob 2464 remote strings sm

Pelkar:


widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
```diff
@@ -970,7 +971,9 @@ internal class CallView(
                 callState.callStatus.formattedOperatorName,
                 ""
             )
-            if (callState.companyName != null) {
+            if (callState.companyName != null) { // TODO FIXME before release
```
andrews-moc, line 974: What exactly should be fixed within this TODO?

andrews-moc: I think it makes sense to be more specific because for me, at least it's not obvious here...

andrews-moc (approved):

gugalo (approved):
Could have better PR title )

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
Pelkar: The company name should be configurable from remote strings, so we need to have a way of using the companyName from remote strings instead of the one from GliaInit

PR #777, opened 2023-09-21
Feature/custom locales minor fixes

Pelkar:
Minor fixes here and there that include, but are not limited to:
* Some strings not being used in code
* Fixed call duration running 1000 times faster
* Fixed test with new string resource fetching
* Removed some unused tools annotated  parameters from xml files


andrews-moc (approved):
Looks good to me, although, rebase is needed because many changes already [exist](https://github.com/salemove/android-sdk-widgets/pull/775/files) in custom_locals branch.

AndriiHorishniiMOC (approved):
Here are lots of minor fixes :)
Good job!

gugalo (approved):

PR #780, opened 2023-09-25
Feature/mob 2691 var names

Pelkar:
Created a value based getRemoteString varargs for string insert value replacements

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/adapter/UploadAttachmentAdapter.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/adapter/UploadAttachmentAdapter.kt
@@ -188,7 +188,7 @@ class ViewHolder(
         itemView.contentDescription = stringProvider.getRemoteString(
             R.string.android_chat_file_accessibility,
             StringKeyPair(StringKey.NAME, displayName),
-            StringKeyPair(StringKey.SIZE, size ?: "N/A"),
+            StringKeyPair(StringKey.SIZE, size ?: stringProvider.getRemoteString(R.string.general_unknown)),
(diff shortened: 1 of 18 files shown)


gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/UploadAttachmentAdapter.kt
```diff
@@ -181,13 +183,13 @@ class ViewHolder(
...
-            displayName
+            StringKeyPair(StringKey.NAME, displayName)
         )
         itemView.contentDescription = stringProvider.getRemoteString(
             R.string.android_chat_file_accessibility,
-            displayName,
-            size,
-            getStatusIndicatorText(attachment.attachmentStatus)
+            StringKeyPair(StringKey.NAME, displayName),
+            StringKeyPair(StringKey.SIZE, size ?: "N/A"),
```
andrews-moc, line 191: @Pelkar Does it make sense to move "N/A" to strings (at least, non-translatable)?

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/UploadAttachmentAdapter.kt
Pelkar: There needs to be a value here, because if we otherwise size: {NOTHING} is weird, N/A should be the closest to multilanguage placeholder to this as we can manage. Otherwise we should start using different string res

AndriiHorishniiMOC (approved):

DavDo (approved):
LGTM

PR #782, opened 2023-09-29
Apply company name from custom locales

andrews-moc:
[Implement Company name from custom locales](https://glia.atlassian.net/browse/MOB-2685)

**Changes overview**
- Company name can now be taken from a single source of truth: `Dependencies.getSdkConfigurationManager().companyName`
- Dropped a way to specify company name through `Intent` (see migration guide)
- Company name priority setup for Android is going to be: custom locales (highest priority) => SDK configuration => default value, which is "Company Name".
- Renamed GliaWidgetsConfigManager to ExampleAppConfigManager to cross it out from the number of classes related to configuration.

[Additional details](https://salemove.slack.com/archives/C04JS5GH6AD/p1695891888361899) on taken decisions.

**Release notes:**

 - [x] Feature
 - [ ] Ignore
 - [x] Migration guide (If changes are needed for integrator already using the SDK - what needs to be communicated? Add underneath please)

**Migration guide:** _If you were specifying the company name by intent.putExtra(GliaWidgets.COMPANY_NAME, “Some value”) please specify it with Glia Widgets SDK initialization (set GliaWidgetsConfig.Builder#companyName) or customize the strings from GliaHub._

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/ExampleAppConfigManager.kt
+++ b/app/src/main/java/com/glia/exampleapp/ExampleAppConfigManager.kt
@@ -16,7 +16,7 @@ import com.glia.widgets.UiTheme
  * glia://widgets/secret?site_id={site_id}&api_key_secret={api_key_secret}&api_key_id={api_key_id}&queue_id={queue_id}&visitor_context_asset_id={visitor_context_asset_id}
  * where all query params are mandatory except visitor_context_asset_id
  */
-object GliaWidgetsConfigManager {
+object ExampleAppConfigManager {
     private const val SECRET_KEY = "secret"
     private const val SITE_ID_KEY = "site_id"
     private const val API_KEY_SECRET_KEY = "api_key_secret"
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
@@ -304,7 +304,7 @@ internal class CallView(
             operatorStatusView.showTransferring()
             operatorNameView.text = (stringProvider.getRemoteString(R.string.engagement_queue_transferring_message))
         } else {
-            operatorNameView.text = stringProvider.getRemoteString(R.string.general_company_name)
+            operatorNameView.text = Dependencies.getSdkConfigurationManager().companyName
             operatorNameView.hint = stringProvider.getRemoteString(R.string.chat_operator_name_accessibility_label)
             handleOperatorStatusViewOperatorImage(state)
         }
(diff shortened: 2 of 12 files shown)


DavDo (approved):
LGTM

app/src/main/java/com/glia/exampleapp/ExampleAppConfigManager.kt
```diff
@@ -16,7 +16,7 @@ import com.glia.widgets.UiTheme
  * glia://widgets/secret?site_id={site_id}&api_key_secret={api_key_secret}&api_key_id={api_key_id}&queue_id={queue_id}&visitor_context_asset_id={visitor_context_asset_id}
  * where all query params are mandatory except visitor_context_asset_id
  */
-object GliaWidgetsConfigManager {
+object ExampleAppConfigManager {
```
AndriiHorishniiMOC, line 19: 👍

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
```diff
@@ -304,8 +304,7 @@ internal class CallView(
             operatorStatusView.showTransferring()
             operatorNameView.text = (stringProvider.getRemoteString(R.string.engagement_queue_transferring_message))
         } else {
-            operatorNameView.text = stringProvider.getRemoteString(R.string.general_company_name)
-            operatorNameView.hint = stringProvider.getRemoteString(R.string.chat_operator_name_accessibility_label)
```
AndriiHorishniiMOC, line 308: What about accessibility?

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
andrews-moc: @Pelkar It seems to me that we need a separate string that will be used as an accessibility label. Basically, it should be a translation for the "Company name".

So, one string for the name itself and another one just as a translation for "Company name".

As far as I understand, currently, we have only one string for the name.

andrews-moc: Or I can just use the name of the company as an accessibility label.

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
```diff
@@ -974,9 +974,9 @@ internal class CallView(
                 callState.callStatus.formattedOperatorName,
                 ""
             )
-            if (callState.companyName != null) { // TODO FIXME Make CompanyName configurable from remote strings
+            if (callState.companyName != null) {
                 companyNameView.text = callState.companyName
-                companyNameView.hint = stringProvider.getRemoteString(R.string.general_company_name)
+                companyNameView.hint = callState.companyName
```
andrews-moc, line 979: @Pelkar  It seems to me that we need a separate string that will be used as an accessibility label. Basically, it should be a translation for the "Company name" wording.

So, one string for the name itself and another one just as a literal translation for "Company name".

As far as I understand, currently, we have only one string for the name. And at the moment, the name of the company serves as an accessibility label. Not critical, but could be improved.

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
andrews-moc: @AndriiHorishniiMOC Fixed

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
Pelkar: We discussed this with @gersonnoboa before he left for vacation and the decision was to have the accessibility text just be the company name itself. For now, this should remain the case, until further notice. We can always improve on this later on if the need arises.

Pelkar (approved):
Really good job, I like your approach

gugalo (approved):

PR #803, opened 2023-10-16
Acknowledgment dialog use site settings

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2698

**What was solved?**

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**


DavDo (approved):
LGTM

andrews-moc:
> @andrews-moc It should work for regular engagements and call visualizer engagements. It should always be displayed before those engagements are started. In the case of CV engagements, it would be displayed inside the client application screen.

Everything is correct then. I misunderstood regarding the Call Visualizer flow initially.
Okay, thank you.

PR #813, opened 2023-10-26
Prevented unread indicator to be shown during call visualizer engagement

Pelkar:
MOB-2681

**Jira issue:**
https://glia.atlassian.net/browse/MOB-2681

**What was solved?**
Fixed unread message indicator being shown on the bubble during Call Visualiser engagement

**Release notes:**

 - [ ] Feature
 - [ ] Ignore
 - [X] Release notes (Is it clear from the description here?)

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/view/head/ChatHeadView.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/view/head/ChatHeadView.kt
@@ -94,7 +94,9 @@ class ChatHeadView @JvmOverloads constructor(
 
     override fun showUnreadMessageCount(unreadMessageCount: Int) {
         post {
-            if (!isCallVisualizerUseCase.invoke()) {
+            if (isCallVisualizerUseCase.invoke()) {
+                binding.chatBubbleBadge.isVisible = false
+            } else {
                 binding.chatBubbleBadge.apply {
                     text = unreadMessageCount.toString()
                     isVisible = isDisplayUnreadMessageBadge(unreadMessageCount)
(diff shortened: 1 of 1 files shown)


widgetssdk/src/main/java/com/glia/widgets/view/head/ChatHeadView.kt
```diff
@@ -91,9 +94,11 @@ class ChatHeadView @JvmOverloads constructor(
...
     override fun showUnreadMessageCount(unreadMessageCount: Int) {
         post {
-            binding.chatBubbleBadge.apply {
-                text = unreadMessageCount.toString()
-                isVisible = isDisplayUnreadMessageBadge(unreadMessageCount)
+            if (!isCallVisualizerUseCase.invoke()) {
+                binding.chatBubbleBadge.apply {
+                    text = unreadMessageCount.toString()
+                    isVisible = isDisplayUnreadMessageBadge(unreadMessageCount)
+                }
```
gugalo, line 101: I would add a small comment as it might not be very intuitive that Call Visualizer does not support chat so it should never display an unread message indicator. Or alternatively, you could make the code more explicit about this behavior e.g:
```suggestion
            if (isCallVisualizerUseCase.invoke()) {
                binding.chatBubbleBadge.isVisible = false
            } else {
                binding.chatBubbleBadge.apply {
                    text = unreadMessageCount.toString()
                    isVisible = isDisplayUnreadMessageBadge(unreadMessageCount)
                }
```

Eather option is fine.

gugalo (approved):

DavDo (approved):
LGTM

AndriiHorishniiMOC (approved):

andrews-moc (approved):
Agree with Aleksander's comment. All the rest is fine.

PR #826, opened 2023-11-06
Incremented version to 2.2.2

Pelkar:
**Release notes:**

 - [ ] Feature
 - [X] Ignore

### DIFF
--- a/version.properties
+++ b/version.properties
@@ -1,4 +1,4 @@
-#Tue Oct 31 09:50:45 UTC 2023
+#Mon Nov 06 09:53:00 UTC 2023
 dependency.coreSdk.version=1.1.9
 widgets.versionCode=72
-widgets.versionName=2.2.0
+widgets.versionName=2.2.2


DavDo (approved):

version.properties
```diff
@@ -1,4 +1,4 @@
-#Tue Oct 31 09:50:45 UTC 2023
+#Mon Nov 06 09:53:00 UTC 2023
 dependency.coreSdk.version=1.1.9
 widgets.versionCode=72
-widgets.versionName=2.2.0
+widgets.versionName=2.2.2
```
andrews-moc, line 4: Taking into account that current version is 2.2.0, should the new one be 2.2.1?

gugalo: ➕

gugalo (approved):
What Andrew mentioned )

version.properties
Pelkar: Because iOS has already released 2.2.1, we decided to skip 2.2.1 and release this feature on both platforms under 2.2.2

andrews-moc: Aah, yes, we discussed it on Friday, indeed.

andrews-moc (approved):

PR #829, opened 2023-11-06
Backmerge/2.2.2

Pelkar:
Backmerge from master


DavDo (approved):

yurii-glia (changes requested):
we should not make cyclic merge commits, let's rebase development on top of master to re-order history. i will take a look on it. for now just mark "request changes" to avoid accidentally merge.

yurii-glia:
fixed in #830

PR #831, opened 2023-11-07
Refactor Dialogs to inflate views independently

DavDo:
MOB 2814

**Jira issue:**
https://glia.atlassian.net/browse/MOB-2814

**What was solved?**

- Created a mechanism to create GlobalColors from UiTheme. This helped to create AlertTheme from UiTheme and then merge it with provided by UnifiedUi, so now all the Dialog Views are configured with AlertTheme
- Created a mechanism to inflate all the Dialog Views independently and before Dialog is created. This will make all the Views(fully configured with themes) available outside of Alertdialogs and Dialog builders so that these views can be easily tested in snapshot tests.

All these changes will make it easy to change any of these views independently, without affecting other dialogs.

2nd part of improvements are here - https://github.com/salemove/android-sdk-widgets/pull/834

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
@@ -1,7 +1,6 @@
 package com.glia.widgets.call
 
 import android.content.Context
-import android.content.DialogInterface
 import android.content.Intent
 import android.content.res.Configuration
 import android.content.res.TypedArray
@@ -366,13 +365,9 @@ internal class CallView(
     }
 
     private fun showExitQueueDialog() {
-        alertDialog = Dialogs.showOptionsDialog(
-            context = this.context,
-            theme = theme,
-            title = stringProvider.getRemoteString(R.string.engagement_queue_leave_header),
-            message = stringProvider.getRemoteString(R.string.engagement_queue_leave_message),
-            positiveButtonText = stringProvider.getRemoteString(R.string.general_yes),
-            negativeButtonText = stringProvider.getRemoteString(R.string.general_no),
+        alertDialog = Dialogs.showExitQueueDialog(
+            context = context,
+            uiTheme = theme,
             positiveButtonClickListener = {
                 dismissAlertDialog()
                 callController?.endEngagementDialogYesClicked()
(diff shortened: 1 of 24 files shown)


andrews-moc (approved):

AndriiHorishniiMOC (approved):
LGTM!

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
```diff
@@ -366,13 +367,12 @@ internal class CallView(
     }
 
     private fun showExitQueueDialog() {
-        alertDialog = Dialogs.showOptionsDialog(
-            context = this.context,
-            theme = theme,
+        val payload = DialogPayload.Option(
```
Pelkar, line 370: I like this approach

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
```diff
@@ -382,49 +382,54 @@ internal class CallView(
...
                 callController?.endEngagementDialogDismissed()
-            },
-            isButtonsColorsReversed = true
-        )
+            }
+        }
     }
 
     private fun showAllowScreenSharingNotificationsAndStartSharingDialog() {
         if (alertDialog == null || !alertDialog!!.isShowing) {
```
Pelkar, line 398: There is still a bit of duplicated code here. Could you maybe extract the if statement into something more general and instead of showAllowScreenSharingNotificationsAndStartSharingDialog we would have something like createAllowSreenSharingAndStartSharingOptions(): DialogPayload.Option that would create only the option?

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
```diff
@@ -707,40 +719,26 @@ internal class CallView(
         }
     }
 
-    private fun showOptionsDialog(
```
Pelkar, line 710: Why was this deleted, Was this unused or was this merged with some other dialog?

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
```diff
@@ -809,12 +807,13 @@ internal class CallView(
     }
 
     private fun showOverlayPermissionsDialog() {
```
Pelkar, line 809: And shouldn't these dialog type definitions be somewhere else entirely? Why are they defined in View classes. This way they are not reusable. Maybe define them in Dialogs.kt instead and the view only adds listeners?

Pelkar (changes requested):
The overall approach is good, but I would like to have the dialog configurations out of views. That was one of the main problems with dialogs in the first place that we have view specific dialogs

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
DavDo: This was used only once, so I decided to use function from `Dialogs` instead

DavDo:
> The overall approach is good, but I would like to have the dialog configurations out of views. That was one of the main problems with dialogs in the first place that we have view specific dialogs

I agree with all of the suggestions provided, and I will proceed with implementing them. Once I have made the necessary changes, I will request a review again.

gugalo (approved):

DavDo:
Guys I made most of the requested changes, and separate functions for all the dialog types, so I'm re-requesting your review

DavDo:
2nd part of improvements are here - https://github.com/salemove/android-sdk-widgets/pull/834

AndriiHorishniiMOC (approved):

andrews-moc (approved):
Looks good to me. Even better now.

gugalo (approved):

PR #833, opened 2023-11-08
Hide Logger functionality from public usage, add unit tests for Logger class

andrews-moc:
[Developers want the Logger functions to be non-static in Widgets and Core SDK](https://glia.atlassian.net/browse/MOB-2822)

**What was solved?**
1. Disallow integrators to use our Logger directly so that they should not be able to call `com.glia.widgets.helper.Logger.d("Integrator", "Hello, Glia developers ;)")`.
2. Add unit tests for `Logger class`.

I decided to go with a simple and reliable solution instead of initially planned (see TL;DR for the details).

TL;DR
The initial idea was to make logging functions non-static in `Logger`. Make the `Logger` constructor private. Inject Logger to the components using DI (e.g. ControllerFactory, etc).

However, it turned out that this approach has several disadvantages:
- it's not possible to inject to all type of components, in some places I had to create separate `Logger` instance
- we will have introduce new rules for `Logger` usage internally and follow them, so it will add development complexity
- it's a good deal of changes (actually, I made them but was not satisfied)
- we will have many instances of logger

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/helper/Logger.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/helper/Logger.kt
@@ -6,7 +6,7 @@ import com.glia.androidsdk.LogLevel
 import com.glia.androidsdk.LoggingAdapter
 import java.util.function.Consumer
 
-object Logger {
+internal object Logger {
 
     const val SITE_ID_KEY = "site_id"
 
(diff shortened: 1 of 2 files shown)


DavDo (approved):

AndriiHorishniiMOC (approved):
👍

PR #839, opened 2023-11-15
Fix The dialog for screen sharing in the CV flow is not appearing

DavDo:
MOB 2847

**Jira issue:**
https://glia.atlassian.net/browse/MOB-2847

**What was solved?**
Add Screen sharing callback only when the `CallVisualizer` engagement is started.
This way we will not brake showing screen sharing dialog from `ActivityWatcherForCallVisualizer` only for CV flows and it will also be safe for configuration changes

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


Pelkar (approved):
LGTM and seems to work

andrews-moc (approved):
Looks good to me. 👍

PR #845, opened 2023-11-20
Fixed chat head duplications and crash

Pelkar:
MOB-2855

Audio channel state was restarted on engagement end and thus opening chat engagement after video had an empty audio state.

Also added a crash prevention mechanism to avoid "view already added" type of exceptions.

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/core/engagement/GliaEngagementTypeRepository.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/engagement/GliaEngagementTypeRepository.java
@@ -2,6 +2,7 @@ package com.glia.widgets.core.engagement;
 
 import com.glia.widgets.core.operator.GliaOperatorMediaRepository;
 import com.glia.widgets.core.visitor.GliaVisitorMediaRepository;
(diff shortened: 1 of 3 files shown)


DavDo (approved):
Good job!

widgetssdk/src/main/java/com/glia/widgets/core/engagement/GliaEngagementTypeRepository.java
```diff
@@ -2,6 +2,7 @@
 
 import com.glia.widgets.core.operator.GliaOperatorMediaRepository;
 import com.glia.widgets.core.visitor.GliaVisitorMediaRepository;
+import com.glia.widgets.helper.Logger;
```
andrews-moc, line 5: As far as I understand, this import is not used, is it?

andrews-moc (approved):
Good job!

widgetssdk/src/main/java/com/glia/widgets/core/engagement/GliaEngagementTypeRepository.java
Pelkar: Yes, thank you!

PR #851, opened 2023-11-24
Add links to confirmation dialog

AndriiHorishniiMOC:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2862

**What was solved?**
Added the ability for integrators to add links to their “terms and conditions” or/and “privacy policies” to the live observation confirmation dialog.

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
@@ -517,7 +517,9 @@ class ChatView(context: Context, attrs: AttributeSet?, defStyleAttr: Int, defSty
             theme = theme,
             companyName = companyName,
             positiveButtonClickListener = { onEngagementConfirmationDialogAllowed() },
-            negativeButtonClickListener = { onEngagementConfirmationDialogDismissed() }
+            negativeButtonClickListener = { onEngagementConfirmationDialogDismissed() },
+            link1ClickListener = {}, //TODO: will be implemented on the next task
+            link2ClickListener = {}
         )
     }
 
--- a/widgetssdk/src/main/res/values/new_strings.xml
+++ b/widgetssdk/src/main/res/values/new_strings.xml
@@ -202,8 +202,12 @@
     <string name="engagement_phone_title">Phone</string> <!-- TODO NEW -->
 
     <!--    Live Observation -->
-    <string name="live_observation_confirm_title">Before We Continue</string>
-    <string name="live_observation_confirm_message">Please allow the {companyName} representative to view this application screen for improved support.</string>
+    <string name="engagement_confirm_title">Before We Continue</string>
+    <string name="engagement_confirm_message">Please allow the {companyName} representative to view this application screen for improved support.</string>
+    <string name="engagement_confirm_link1_text">Terms and Conditions</string>
+    <string name="engagement_confirm_link1_url"/>
+    <string name="engagement_confirm_link2_text">Privacy Policies</string>
+    <string name="engagement_confirm_link2_url"/>
     <string name="live_observation_indicator_message">App screen is visible to the agent</string>
 
     <!--    GVA -->
--- a/app/src/main/res/raw/sample_unified_config.json
+++ b/app/src/main/res/raw/sample_unified_config.json
@@ -32,6 +32,48 @@
         ]
       }
     },
+    "linkButton": {
+      "background": {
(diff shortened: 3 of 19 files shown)


DavDo (changes requested):
I don't particularly appreciate adding new parameters to the Option payload and new options view and view-binding buttons that should be used only for a Confirmation dialog, because the options dialog/view is used in several scenarios with different texts. I'd suggest adding a separate view for this dialog with its Payload, ViewBinding, and other components.

widgetssdk/src/main/res/values/new_strings.xml
```diff
@@ -202,9 +202,13 @@
...
-    <string name="live_observation_confirm_title">Before We Continue</string>
-    <string name="live_observation_confirm_message">Please allow the {companyName} representative to view this application screen for improved support.</string>
-    <string name="live_observation_indicator_message">App screen is visible to the agent</string>
+    <string name="engagement_confirm_title">Before We Continue</string>
+    <string name="engagement_confirm_message">Please allow the {companyName} representative to view this application screen for improved support.</string>
+    <string name="engagement_indicator_message">App screen is visible to the agent</string>
+    <string name="engagement_confirm_link1_text">Terms and Conditions</string>
+    <string name="engagement_confirm_link1_url"/>
+    <string name="engagement_confirm_link2_text">Privacy Policies</string>
+    <string name="engagement_confirm_link2_url"/>
```
DavDo, line 211: Are these changes submitted to the remote strings?

andrews-moc (commented):
I agree with Davit's concerns.

widgetssdk/src/main/res/values/new_strings.xml
AndriiHorishniiMOC: PR in review.

AndriiHorishniiMOC:
I added the new dialogue type.

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
```diff
@@ -517,7 +517,9 @@ class ChatView(context: Context, attrs: AttributeSet?, defStyleAttr: Int, defSty
             theme = theme,
             companyName = companyName,
             positiveButtonClickListener = { onEngagementConfirmationDialogAllowed() },
-            negativeButtonClickListener = { onEngagementConfirmationDialogDismissed() }
+            negativeButtonClickListener = { onEngagementConfirmationDialogDismissed() },
+            link1ClickListener = {}, //TODO: will be implemented on the next task
```
Pelkar, line 521: Can we maybe make this a little more specific? We are expecting these to be ToS and PP right? Maybe we call the listeners accordingly? Would add a lot of useful context to reading the code

widgetssdk/src/main/res/values/new_strings.xml
```diff
@@ -205,6 +205,10 @@
     <string name="live_observation_confirm_title">Before We Continue</string>
     <string name="live_observation_confirm_message">Please allow the {companyName} representative to view this application screen for improved support.</string>
     <string name="live_observation_indicator_message">App screen is visible to the agent</string>
+    <string name="live_observation_confirm_link1_text">Terms and Conditions</string>
+    <string name="live_observation_confirm_link1_url"/>
```
Pelkar, line 209: rename to engagement_etc?

Pelkar (approved):

widgetssdk/src/main/res/values/new_strings.xml
AndriiHorishniiMOC: Updated with renamed strings.

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
AndriiHorishniiMOC: Actually, these are links one and two. The integrators can use them as they want.
Maybe it makes sense to change it to:
`linkClickListener = { url: String }`

gugalo (approved):
I have nothing to add

DavDo (approved):
Great job!
Are you planning to add snapshot tests in a separate task?

AndriiHorishniiMOC:
Yes, adding the snapshot tests is a separate ticket.

AndriiHorishniiMOC:
!squash

PR #872, opened 2024-01-03
Move screen-sharing stuff to the EngagementRepository

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2959

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


widgetssdk/src/main/java/com/glia/widgets/core/screensharing/ScreenSharingControllerImpl.kt
```diff
@@ -43,31 +53,27 @@ internal class ScreenSharingControllerImpl(
...
 
-    override fun onScreenSharingStarted() {
-        Logger.d(TAG, "screen sharing started")
+    private fun onScreenSharingStarted() {
         viewCallbacks.forEach { it.onScreenSharingStarted() }
     }
 
-    override fun onScreenSharingEnded() {
-        Logger.d(TAG, "screen sharing ended")
+    private fun onScreenSharingEnded() {
```
andrews-moc, line 60: Minor question. Maybe it makes sense to leave these debug logs for screen sharing started/ended events? It was useful for debugging. But it's not a big deal to add them if needed, so I don't insist, just asking your opinion and trying to understand why you decided to remove.

DavDo: These logs are duplicates since they are already logged where events are handled in the `EngagementRepositoryImpl`.

andrews-moc:
@DavDo Let me clarify, to make sure I'm understanding the goal of the changes. So, you have removed a separate `GliaScreenSharingRepository` and moved it's code to engagement repository, right?

From one side, lines of code in engagement repository have grown, which is a drawback. From the other side, it became easier to handle screen-sharing related events like all engagement related events from a single place.

Is my understanding correct, could you please elaborate?

widgetssdk/src/main/java/com/glia/widgets/core/screensharing/ScreenSharingControllerImpl.kt
andrews-moc: I see, okay, this is a good change then. 👍

DavDo:
> @DavDo Let me clarify, to make sure I'm understanding the goal of the changes. So, you have removed a separate `GliaScreenSharingRepository` and moved it's code to engagement repository, right?
> 
> From one side, lines of code in engagement repository have grown, which is a drawback. From the other side, it became easier to handle screen-sharing related events like all engagement related events from a single place.
> 
> Is my understanding correct, could you please elaborate?

Yes, you're right. From one side, lines of code in the engagement repository have grown but it still has a simple responsibility, to handle all the engagement-related events, and produce states based on them, so all the other classes need to observe these states instead of subscribing to raw events and unsubscribing from them. In my opinion, this is not a drawback, if we take into account that in the future this class will not grow anymore.

AndriiHorishniiMOC (approved):
Great work!

PR #874, opened 2024-01-05
Fix IllegalStateException: for NotificationRemovalService

andrews-moc:
[MOB-2837](https://glia.atlassian.net/browse/MOB-2837)

**Crash details:**
```
java.lang.IllegalStateException: Not allowed to start service Intent { com.glia.widgets.core.notification.NotificationRemovalService }: app is in background
```
Example from Sentry is [here](https://glia.sentry.io/issues/4766376086/?alert_rule_id=967509&alert_timestamp=1703689998756&alert_type=email&environment=prod-us&notification_uuid=097e6631-f490-4032-8f8f-622acb0a80d3&project=1887832&referrer=alert_email).

This happens if an app is in the background for 60+ seconds and `NotificationRemovalService` is attempted to be started.

**What was solved?**
Lets start `NotificationRemovalService` when app goes to 'OnStop' to make sure the service is not attempted to start after app being in background for over than 60 seconds.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore
 - [x] Release notes (Is it clear from the description here?)
 - [ ] Migration guide (If changes are needed for integrator already using the SDK - what needs to be communicated? Add underneath please)

**Additional info:**

 - [ ] Tests updated? Added? Unit, acceptance, snapshots?
 - [ ] Logging for future troubleshooting of client issues added?

**Release note:**
Fix IllegalStateException: for NotificationRemovalService

[MOB-2837]: https://glia.atlassian.net/browse/MOB-2837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #875, opened 2024-01-08
Fix IllegalStateException: for ChatHeadService

andrews-moc:
These changes will prevent the `IllegalStateException: Not allowed to start service ChatHeadService: app is in background` crash. A crash example is [here](https://glia.sentry.io/issues/4599728998/events/2fddaf124e8140beb2bcf286d8efcdea/?project=1887832).

[MOB-2974](https://glia.atlassian.net/browse/MOB-2974)

**What was solved?**
Crash details are described in the comments to the code. The reason is the same as for [IllegalStateException: for NotificationRemovalService](https://github.com/salemove/android-sdk-widgets/pull/874). However, the fix that I came up with is different. Taking into account two peculiarities:

1. We cannot start `ChatHeadService` before engagement is started or it's not straightforward to do this
2. `ChatHeadService` is attempted to be started once the app is resumed again

I decided that surrounding the `startService()` call with a `try` - `catch` block might prevent the crash, bubble will still be displayed once the app is in the foreground again.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

 - [x] Logging for future troubleshooting of client issues added?

**Release note:**
Fix IllegalStateException: for ChatHeadService

[MOB-2974]: https://glia.atlassian.net/browse/MOB-2974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/core/chathead/ChatHeadManager.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/chathead/ChatHeadManager.java
@@ -3,8 +3,12 @@ package com.glia.widgets.core.chathead;
 import android.content.Context;
 import android.content.Intent;
 
+import com.glia.widgets.helper.Logger;
+
 
 public class ChatHeadManager {
+    private static final String TAG = ChatHeadManager.class.getSimpleName();
+
     private final Context applicationContext;
     private final Intent chatHeadServiceIntent;
 
@@ -20,7 +24,24 @@ public class ChatHeadManager {
     public void startChatHeadService() {
         if (!isServiceStarted) {
             isServiceStarted = true;
-            applicationContext.startService(chatHeadServiceIntent);
+            try {
+                applicationContext.startService(chatHeadServiceIntent);
+            } catch (IllegalStateException exception) {
+                isServiceStarted = false;
+                Logger.w(TAG, "Application is in a state where the service can not be started" +
+                    " (such as not in the foreground in a state when services are allowed)");
+                // This prevents the 'Not allowed to start ChatHeadService: app is in background' crash.
+                // Example scenario:
+                // Disable the 'Display over other apps' Android setting for Widgets example app
+                // Open Widgets example app, tap 'Start new chat flow'
+                // Tap 'OK' for 'Screen Overlay Permissions Required' dialog, you will be navigated to Android Settings
+                // Open Widgets example app, go to the main screen to have chat bubble displayed
+                // Come back to Android Settings, allow the 'Display over other apps' setting
+                // Tap device Home button (bubble is not displayed)
+                // Go to Glia Hub, wait for 60+ seconds
+                // Accept engagement
+                // IllegalStateException will be caught. Once visitor opens Widgets example app, bubble will be displayed and start working.
+            }
         }
     }
 


widgetssdk/src/main/java/com/glia/widgets/core/chathead/ChatHeadManager.java
```diff
@@ -20,7 +24,24 @@ public ChatHeadManager(
     public void startChatHeadService() {
         if (!isServiceStarted) {
             isServiceStarted = true;
-            applicationContext.startService(chatHeadServiceIntent);
+            try {
+                applicationContext.startService(chatHeadServiceIntent);
+            } catch (IllegalStateException exception) {
+                isServiceStarted = false;
```
Pelkar, line 30: Is there validity for notifying the user as well?

Pelkar (approved):
LGTM! Left one comment

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/core/chathead/ChatHeadManager.java
andrews-moc: @Pelkar Ideally, yes, it might be good. Do you have any ideas on how to notify?

Pelkar: This could/should be discussed with Mari and Risto, but Risto is on vacation.

If I were to propose something, then either a general error dialog or a toast.

DavDo: What's the benefit of notifying the user about this? In my opinion, this is an internal error, and it should only be logged(most probably with a warning level). We should also take into account that the app should be in the background when this happens.

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/core/chathead/ChatHeadManager.java
andrews-moc: @DavDo Regarding log level - I have changed the log level from info to warning.

Regarding dialog or toast, I'm not sure it would be easy for the case when app is in background for 60+ seconds.
Also, let's take into account that:
1) It's kind of an edge case - 12 cases per 90 days in Sentry
2) Once the visitor opens the app, bubble will start working

@Pelkar do you still think I need to raise a question to Mari and Risto?

DavDo: > Regarding dialog or toast, I'm not sure it would be easy for the case when app is in background for 60+ seconds.

Besides those difficulties, I am unsure if providing such information to the user is necessary.

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/core/chathead/ChatHeadManager.java
Pelkar: Fair enough if it is such an extreme edge case.

andrews-moc:
!squash

PR #881, opened 2024-01-11
Refactor Dialog related functionality and covert to kotlin

DavDo:
MOB 2983

**Jira issue:**

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


widgetssdk/src/main/java/com/glia/widgets/core/dialog/model/DialogState.kt
```diff
@@ -0,0 +1,23 @@
+package com.glia.widgets.core.dialog.model
+
+import com.glia.androidsdk.comms.MediaUpgradeOffer
+
+internal sealed interface DialogState {
+    object None : DialogState
+    object OverlayPermission : DialogState
+    object ExitQueue : DialogState
+    data class StartScreenSharing(val operatorName: String?) : DialogState
```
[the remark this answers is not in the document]
Pelkar: What if you were to change the DialogState into an internal class that would have operatorName, mediaUpgradeOffer, mode and DialogType as arguments?

You would probably have to create custom contstructors for the argumented types, but this way the different types would be equivalent and the arguments would be there for the types that need them

Pelkar (approved):
overall looks solid, but the DialogState implementation does not look good to me. Overall I am not completely against it, but would prefer a solution that would contain interchangeable sealed class members

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/core/dialog/model/DialogState.kt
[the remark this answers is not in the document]
DavDo: After implementing the proposed solution using generics, I realized that it did not provide any benefits, so I decided to leave everything as it was. Let me explain in detail.

```kotlin
sealed interface Result<out T> {
    object Loading: Result<Nothing>
    data class Error(val ex: Exception): Result<Nothing>
    data class Success<out T>(val data: T): Result<T>
}
``` 
In the Result<T> interface, T can represent any type of value that might be returned upon the success of some operation. This allows us to keep the type safety by knowing that Result<User> will hold a User in case of success, Result<Post> will hold Post, and so on.

Now, let's compare this to the DialogState interface. 

In DialogState, each state is explicitly defined as an object or a data class with specifically defined parameters. There's no need for a generic type because each possible state of the dialog is explicitly enumerated in the interface. 

For example, StartScreenSharing holds an `operatorName` of type String and this would never be a different type (like User, Post, or Int). Similarly, MediaUpgrade holds a mediaUpgradeOffer of type MediaUpgradeOffer, and so on. 

Essentially, we don't use a generic type in DialogState because we don't have a situation where we might have a DialogState<User>, DialogState<Post>, DialogState<Int>, etc. Each state is distinct and holds specifically defined data related to that state.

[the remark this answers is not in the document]
DavDo: But during that local update, I found that we had set a lot of `DialogInterface.OnCancelListener` to the dialogs that are not cancelable.
Will provide that update soon.

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.kt
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+
+    override fun dismissVisitorCodeDialog() {
+        Logger.i(TAG, "Dismiss Visitor Code Dialog")
+        dialogManager.remove(DialogState.VisitorCode)
+    }
+
+    override fun showUnexpectedErrorDialog() {
+        // PRIORITISE THIS ERROR AS IT IS ENGAGEMENT FATAL ERROR INDICATOR (eg. GliaException:{"details":"Queue is closed","error":"Unprocessable entity"}) for example
```
andrews-moc, line 112: What does this comment mean? Is it a TODO? Does it mean that some additional changes needed?

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.kt
```diff
@@ -0,0 +1,177 @@
...
+        Logger.i(TAG, "Show Unexpected error Dialog")
+        dialogManager.addAndEmit(DialogState.UnexpectedError)
+    }
+
+    override fun showOverlayPermissionsDialog() {
+        Logger.i(TAG, "Show Overlay permissions Dialog")
+        setOverlayPermissionRequestDialogShownUseCase.execute()//TODO remove this shit
```
andrews-moc, line 119: It shouldn't be committed I suppose. 🙂

andrews-moc (approved):
Left two comments. Other than that - looks good to me.

gugalo (commented):
No additional comments )

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.kt
gugalo: Love this one! ❤️

DavDo: Yes, I solved it another way

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.kt
DavDo: This is just information about why we're doing so(prioritizing this dialog).

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.kt
DavDo: removed

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.kt
andrews-moc: Does it make sense to make it lowercase instead of uppercase?

DavDo: I don't know who wrote this comment, but I think it makes sense, and since I already merged this branch will do it on my next PR.

andrews-moc: Ah, no problem, we can leave it uppercase, it's not critical.

PR #885, opened 2024-01-15
Rename Controller interfaces to correspond to a Contract pattern

DavDo:
**Mostly Renamed files**

**Jira issue:**
https://glia.atlassian.net/browse/MOB-2991

**What was solved?**
I added some new interfaces that correspond to the pattern interface `Controller → class ControllerImpl`. But all the interfaces in our codebase that previously had an interface corresponding interface `Contract.Controller → class Controller`. It is preferable to have one pattern per layer, so all the controllers should be aligned to one pattern.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


app/src/main/java/com/glia/exampleapp/MainFragment.kt
```diff
@@ -458,14 +458,14 @@ class MainFragment : Fragment() {
         val visitorContext = getContextAssetIdFromPrefs(sharedPreferences)
         val cv = GliaWidgets.getCallVisualizer()
         if (!visitorContext?.trim { it <= ' ' }.isNullOrBlank()) {
-            cv.addVisitorContext(visitorContext)
+            cv.addVisitorContext(visitorContext!!)
```
andrews-moc, line 461: I would prefer to avoid force unwrapping. I would rather use the let{ } construction.

andrews-moc (approved):

app/src/main/java/com/glia/exampleapp/MainFragment.kt
DavDo: fixed [b2be851](https://github.com/salemove/android-sdk-widgets/pull/885/commits/b2be85199ae2ef8ac3475fbcd2177a406be6b3fc)

AndriiHorishniiMOC (approved):

DavDo:
!squash

PR #891, opened 2024-01-18
Queue for engagement after overlay dialog interaction

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2995

**What was solved?**
For details please check the ticket

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallController.kt
@@ -11,6 +11,7 @@ import com.glia.widgets.Constants
 import com.glia.widgets.call.CallStatus.EngagementOngoingAudioCallStarted
 import com.glia.widgets.call.CallStatus.EngagementOngoingVideoCallStarted
 import com.glia.widgets.call.domain.HandleCallPermissionsUseCase
+import com.glia.widgets.chat.domain.DecideOnQueueingUseCase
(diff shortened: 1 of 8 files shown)


andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallController.kt
```diff
@@ -176,20 +179,20 @@ internal class CallController(
         createNewTimerStatusCallback()
         initControllerCallbacks()
         initMessagesNotSeenCallback()
-        tryToQueueForEngagement(queueId, visitorContextAssetId, mediaType)
+        tryToQueueForEngagement(queueId)
```
Pelkar, line 182: So visitorContextAssetId is no longer passed through tryToQueueForEngagement and only during enqueueForEngagement? A would like an explanation on how this works now

Pelkar (approved):
All looks good, but I am having a hard time understanding the logic progression now. Maybe we can discuss in a little more detail during standup?

Approving for now

widgetssdk/src/main/java/com/glia/widgets/call/CallController.kt
DavDo: The properties mentioned are saved in the ChatState, a few steps before.

PR #895, opened 2024-01-19
Refactor base classes of snapshot tests

AndriiHorishniiMOC:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3001

**What was solved?**

- Paparazzi parameters were added to the base class constructor.
- Custom parameters for a single test are now available on the `SnapshotTest.snapshot` method.
- `SnapshotTest.snapshotFullSize` method was removed. Instead, the `SnapshotTest.fullWidthRenderMode` was added.
- To test animations (like `ProgressBar`), `animationRule` was added. The `SnapshotTest.dummyRule` can be used to turn off this rule. Added tests for cases where the animated element becomes visible with the new rule.
- The accuracy of screenshot difference detection is increased by decreasing `maxPercentDifference`. We need it because minor changes did not trigger test failures. Updated to current state screenshots were added to PR.
- Some VGA tests were refactored. The height for those views was calculated incorrectly. Instead of adding some fake data to this view to make them higher, the height was set manually. Now, the screenshots look better, and the tests do exactly what they need.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


Pelkar (approved):
LGTM

DavDo (approved):
LGTM

PR #910, opened 2024-01-26
Hide Configuration, CallIntentBuilder and CallIntentReader classes from integrators

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/GR-553

**What was solved?**
These classes are in Java, so to hide them I have to convert them to Kotlin and mark them 'internal'.
Did some refactoring in this classes

For release notes: just made two more classes less accessable to integrators

**Release notes:**

 - [ ] Feature
 - [ ] Ignore


widgetssdk/src/main/java/com/glia/widgets/call/CallIntentBuilder.kt
```diff
@@ -0,0 +1,26 @@
+package com.glia.widgets.call
+
+import android.content.Context
+import android.content.Intent
+import com.glia.widgets.GliaWidgets
+
+internal data class CallIntentBuilder(private val context: Context) {
```
DavDo, line 7: Since this class is only for internal usage and used only to build intent, I'd suggest converting it to a higher-level function like
```kotlin
fun buildCallIntent(context: Context, configuration: Configuration): Intent {
// build intent here
} 
```

widgetssdk/src/main/java/com/glia/widgets/call/CallIntentReader.kt
```diff
@@ -0,0 +1,37 @@
+package com.glia.widgets.call
+
+import androidx.appcompat.app.AppCompatActivity
+import com.glia.androidsdk.Engagement
+import com.glia.widgets.GliaWidgets
+import com.glia.widgets.core.configuration.GliaSdkConfiguration
+
+internal class CallIntentReader private constructor(activity: AppCompatActivity) {
```
DavDo, line 8: Same here

widgetssdk/src/main/java/com/glia/widgets/call/CallIntentBuilder.kt
gugalo: Something like this?
https://github.com/salemove/android-sdk-widgets/commit/d76bb9ca4f8754e5730ce6796e5e33de308c6d18

widgetssdk/src/main/java/com/glia/widgets/call/CallActivityIntentHelper.kt
```diff
@@ -0,0 +1,41 @@
+package com.glia.widgets.call
+
+import android.content.Context
+import android.content.Intent
+import androidx.appcompat.app.AppCompatActivity
+import com.glia.androidsdk.Engagement
+import com.glia.widgets.GliaWidgets
+import com.glia.widgets.core.configuration.GliaSdkConfiguration
+
+class CallActivityIntentHelper {
```
DavDo, line 10: I think you forgot to make it internal, also it's better to convert it to `object` instead of using `companion object`.

widgetssdk/src/main/java/com/glia/widgets/call/CallActivityIntentHelper.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    companion object {
+        @JvmStatic
+        fun createIntent(context: Context, configuration: Configuration?): Intent {
```
DavDo, line 14: `configuration` property should be non-null since we're checking for that inside this function. I think this `configuration ?: throw NullPointerException("Configuration can't be null")` check could be moved to the caller side if needed.

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/call/CallActivityIntentHelper.kt
gugalo: Fixed. Also please check the end result again. I also have added one more class into the mix of classes converted to Kotlin

DavDo (approved):
LGTM

andrews-moc (approved):

PR #911, opened 2024-01-26
Fixed Unified theme applying to the Header

AndriiHorishniiMOC:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3016

**What was solved?**
There were two bugs:
1. The title text style is stuck if the new theme doesn't contain those changes. It is fixed by resetting the title appearance before applying the new confirmation screen theme;
2. If the previous text color was a gradient, the new single color doesn't apply. It is fixed by clearing the text paint shader.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**
As you can see the title has not default title style. That is exactly what was fixed in this PR.
image


widgetssdk/src/main/java/com/glia/widgets/view/header/AppBarView.kt
```diff
@@ -204,6 +205,14 @@ internal class AppBarView @JvmOverloads constructor(
         leaveQueueIcon.isVisible = false
     }
 
+    internal fun resetTheme() {
+        val textAttrs = intArrayOf(com.google.android.material.R.attr.textAppearanceHeadline2)
+        context.obtainStyledAttributes(textAttrs).apply {
+            toolbarTitleText?.also { TextViewCompat.setTextAppearance(it, getResourceId(0, 0)) }
+            recycle()
+        }
```
DavDo, line 213:
```suggestion
        context.obtainStyledAttributes(textAttrs).use {
            toolbarTitleText?.also { TextViewCompat.setTextAppearance(it, getResourceId(0, 0)) }
        }
```

DavDo (approved):
LGTM, just one small suggestion.

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/view/header/AppBarView.kt
AndriiHorishniiMOC: Unfortunately:
```
   Explanation for issues of type "NewApi":
   This check scans through all the Android API calls in the application and
   warns about any calls that are not available on all versions targeted by
   this application (according to its minimum SDK attribute in the manifest).
   If you really want to use this API and don't need to support older devices
   just set the minSdkVersion in your build.gradle or AndroidManifest.xml
   files.
```

gugalo (approved):

PR #914, opened 2024-01-29
Converted couple of classes with very common names to internal

gugalo:
**What was solved?**
Did this to prevent integrators from accidentally using those classes because their name is very common in Android development. To make them internal had to also convert them to Kotlin

**Release notes:**

 - [ ] Feature
 - [ ] Ignore


DavDo (approved):
LGTM

widgetssdk/src/main/java/com/glia/widgets/call/CallController.kt
```diff
@@ -180,8 +180,8 @@ internal class CallController(
         initControllerCallbacks()
         initMessagesNotSeenCallback()
         tryToQueueForEngagement(queueId)
-        inactivityTimeCounter.addRawValueListener(inactivityTimerStatusListener)
-        connectingTimerCounter.addRawValueListener(connectingTimerStatusListener)
+        inactivityTimeCounter.addRawValueListener(inactivityTimerStatusListener!!) // Listener is created few lines before
```
AndriiHorishniiMOC, line 183: Honestly, I don't like to use `!!`.
The only place I would do it is in tests.

AndriiHorishniiMOC: And where is it a few lines before?
We could use the value here without the unsafe operator if a method were return the callback object.

AndriiHorishniiMOC (changes requested):

widgetssdk/src/main/java/com/glia/widgets/call/CallController.kt
gugalo: I will look into how to improve this.

gugalo: Fixed in 4a01134

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -232,7 +232,7 @@ internal class ChatController(
         addFileAttachmentsObserverUseCase.execute(fileAttachmentObserver)
         minimizeHandler.addListener { minimizeView() }
         createNewTimerCallback()
-        callTimer.addFormattedValueListener(timerStatusListener)
+        callTimer.addFormattedValueListener(timerStatusListener!!) // Listener is created one line before
```
andrews-moc, line 235: I would try to avoid the force unwrap operation. If it's not possible to use `lateinit` or `by lazy`, I would use the `let {}`construction:
`timerStatusListener?.let{ callTimer.addFormattedValueListener(it) }`

gugalo: Missed it when fixing the other one.
It is fixed now in cdc877a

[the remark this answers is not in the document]
gugalo: Missed it when fixing the other one.
It is fixed now in cdc877a

andrews-moc (approved):

AndriiHorishniiMOC (approved):

PR #915, opened 2024-01-29
Release/2.3.1

Pelkar:
**What was solved?**
Fixed a bug  for Java integrators that was introduced when migrating some files from Java to Kotlin.

**Release notes:**

 - [X] Feature
 - [ ] Ignore

### DIFF
--- a/version.properties
+++ b/version.properties
@@ -1,4 +1,4 @@
-#Fri Jan 19 18:08:25 UTC 2024
+#Fri Jan 29 08:00:00 UTC 2024
 dependency.coreSdk.version=1.2.0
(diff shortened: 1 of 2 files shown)


DavDo (approved):

version.properties
```diff
@@ -1,4 +1,4 @@
 #Fri Jan 19 18:08:25 UTC 2024
 dependency.coreSdk.version=1.2.0
 widgets.versionCode=74
```
gugalo, line 3: This should be incremented, right?

gugalo (changes requested):

gugalo (approved):

PR #918, opened 2024-01-29
Fix close button doesn't work on Messaging Chat screen

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3018

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
@@ -418,7 +418,7 @@ internal class ChatController(
         if (isQueueingOrOngoingEngagement) {
             dialogController.showExitQueueDialog()
         } else {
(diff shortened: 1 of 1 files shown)


Pelkar (approved):
But why was it destroy controllers? Maybe we need to navigate back and destroy controllers?

It does not hurt to check

DavDo:
> But why was it destroy controllers? Maybe we need to navigate back and destroy controllers?
> 
> It does not hurt to check

Actually no. I added destroyControllers at some point during refactoring and then forgot to check/fix it.
On the other hand, `navigateBack()` particularly in this case will keep the ChatController alive - will call `onDestory(true)`, to not fetch chat history every time the user opens the Chat screen.

andrews-moc (approved):

AndriiHorishniiMOC (approved):

PR #919, opened 2024-01-29
Operator image attachment snapshots

AndriiHorishniiMOC:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3022

**What was solved?**
Added snapshot tests for the Operator Image Attachment chat item

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


widgetssdk/src/test/snapshots/images/com.glia.widgets.chat.adapter.holder_OperatorImageAttachmentViewHolderSnapshotTest_withHeader.png
Pelkar, line 1: Why are there so many identical ones? does it make sense to make identical images? I would avoid making identical  images just because, if there is no reason for them to be different, then maybe we should remove them

AndriiHorishniiMOC: I hesitate whether it is really necessary. But on the other hand, our Unified theme is not perfect. Our future fixes may affect some issues with screenshots that look identical today.
I believe that tests are necessary. However, the current Paparazzi implementation doesn't allow setting one image for all of the similar screenshot tests.

DavDo: I think we should rely on differences between test cases rather than generated screenshots.

andrews-moc (approved):

widgetssdk/src/test/snapshots/images/com.glia.widgets.chat.adapter.holder_OperatorImageAttachmentViewHolderSnapshotTest_withHeader.png
AndriiHorishniiMOC: One additional thing: We currently have an issue applying a Unified theme for this view. Some of these similar screenshots should be different.

PR #930, opened 2024-02-01
Consolidate Media upgrade and screen sharing request in one place

DavDo:
This PR is solving only the _**Media Upgrade**_ part

**Jira issue:**
https://glia.atlassian.net/browse/MOB-3039

**What was solved?**
Currently, media upgrade and screen-sharing requests are handled in three separate places on both platforms - Call Visualizer, Call Screen, and Chat Screen. This has resulted in code duplication, which is not scalable or maintainable. Furthermore, there is also a limitation to display media upgrade and screen-sharing requests only in Call and Chat screens for non-CV engagements.

The decision is to do refactoring.

- Handle all the media upgrade and screen-sharing requests in one place
- Show these requests in all possible screens of an integrator app as we do it currently for CV
- Navigate to the Call screen after the media upgrade request is allowed as we do it currently for CV

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


widgetssdk/src/main/java/com/glia/widgets/core/permissions/PermissionManager.kt
```diff
@@ -36,12 +37,17 @@ internal class PermissionManager(
     fun hasScreenSharingNotificationChannelEnabled(): Boolean =
         applicationContext.areNotificationsEnabledForChannel(NotificationFactory.NOTIFICATION_SCREEN_SHARING_CHANNEL_ID)
 
-    fun hasPermission(permission: String) = checkSelfPermission(applicationContext, permission) == PackageManager.PERMISSION_GRANTED
+    private fun hasPermission(permission: String) = checkSelfPermission(applicationContext, permission) == PackageManager.PERMISSION_GRANTED
```
Pelkar, line 40: Eventually all permissions should be handled in the core, right?

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.kt
```diff
@@ -17,9 +15,6 @@ internal interface DialogContract {
         fun dismissDialogs()
         fun showExitQueueDialog()
         fun showExitChatDialog()
-        fun showUpgradeAudioDialog(mediaUpgradeOffer: MediaUpgradeOffer, operatorName: String?)
```
Pelkar, line 20: Why have you divided the dialog logic? To my mind the dialog controller and state should remain more or less unchanged as the Dialog UI creator and restoration classes.

widgetssdk/src/main/java/com/glia/widgets/operator/RequestHandlerActivityWatcher.kt
```diff
@@ -0,0 +1,107 @@
...
+
+    private fun ensureGliaActivity(activity: Activity, callback: Activity.() -> Unit) {
+        if (activity.isGlia) {
+            callback(activity)
+        } else {
+            launchDialogHandlerActivity(activity)
+        }
+    }
+
+    private fun showAlertDialogOnUiThreadWithStyledContext(
```
Pelkar, line 89: Is this function name implying that other dialog calls are running on background threads?

Pelkar (commented):
Overall good job, but I do not like the separation of dialog creation/restoration logic

widgetssdk/src/main/java/com/glia/widgets/core/permissions/PermissionManager.kt
DavDo: I don't think so.
The main reason is that it is not core responsibility, so in my opinion, the core must check permissions and not start media upgrades or screen sharing if necessary permission is not granted. Widgets is fully responsible for handling permissions. Also with the current implementation we have to override `public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults)` method in every activity to pass permissions results to core this is a headache in case, we use activity watcher.

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.kt
DavDo: I hesitated on this question for a long time. DialogController aims to handle dialog queues, priorities, and cross-screen dialogs, but we're dealing with none of this here. The same was in `EngagementCompletionActivityWatcher`. So my feeling is that `DialogController` adds complexity in these two cases.
I would like more opinions on this matter.

widgetssdk/src/main/java/com/glia/widgets/operator/RequestHandlerActivityWatcher.kt
DavDo: Of course not! Very well noticed. Will shorten it.

widgetssdk/src/main/java/com/glia/widgets/core/dialog/DialogController.kt
DavDo: I think I will return these to DialogController

DavDo:
This commit broke the CV because its business logic relies on the `DialogController` callback. I'll fix it after moving the screen-sharing functionality here.

AndriiHorishniiMOC (approved):
Good job!

widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
```diff
@@ -385,4 +379,15 @@ public EngagementCompletionContract.Controller getEndEngagementController() {
         }
         return engagementCompletionController;
     }
+
+    @NonNull
+    public OperatorRequestHandlerContract.Controller getRequestHandlerController() {
```
andrews-moc, line 384: First association with `getRequestHandlerController()` name is HTTP requests. 🙂 Maybe it makes sense to rename it to `getMediaRequestHandlerController()`?

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizerController.kt
```diff
@@ -261,47 +230,9 @@ internal class ActivityWatcherForCallVisualizerController(
         }
     }
 
-    @VisibleForTesting
-    internal fun onMediaUpgradeAccept(offer: MediaUpgradeOffer) {
-        Logger.d(TAG, "Media upgrade request accepted by visitor")
-        if (offer.video != null && offer.video != MediaDirection.NONE) {
-            watcher.openCallActivity()
-        } else {
-            Logger.e(TAG, "Audio upgrade offer in call visualizer", Exception("Audio upgrade offer in call visualizer"))
```
andrews-moc, line 270: I expected to see this log in a new place. However, looks like it's just about to be removed. Don't we need it?

widgetssdk/src/main/java/com/glia/widgets/operator/OperatorRequestHandlerContract.kt
```diff
@@ -0,0 +1,24 @@
...
+import com.glia.widgets.helper.OneTimeEvent
+import io.reactivex.Flowable
+
+internal interface OperatorRequestHandlerContract {
+    sealed interface State {
+        data class RequestMediaUpgrade(val data: MediaUpgradeOfferData) : State
+        data class OpenCallActivity(val mediaType: Engagement.MediaType) : State
+        object DismissAlertDialog : State
+
+        //add request permissions states
```
andrews-moc, line 16: Is it a TODO?

andrews-moc (approved):
Overall idea is good. I just think we need to conduct an additional exploratory testing session to make sure we don't miss anything.

widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
DavDo: This controller handles all the requests from the operator including screen-sharing requests, what about `getOperatorRequestHandlerController`?

DavDo:
> Overall idea is good. I just think we need to conduct an additional exploratory testing session to make sure we don't miss anything.

Sure, that is something that must be done.

widgetssdk/src/main/java/com/glia/widgets/operator/OperatorRequestHandlerContract.kt
DavDo: Yes

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizerController.kt
DavDo: we're logging this event in `EngagementRepositoryImpl` line 449

```kotlin
private fun handleMediaUpgradeOffer(mediaUpgradeOffer: MediaUpgradeOffer) {
        Logger.d(TAG, "Media upgrade requested: $mediaUpgradeOffer")
        _mediaUpgradeOffer.onNext(mediaUpgradeOffer)
    }
``` 
I think it contains all the necessary information. What do you think about this matter?

widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
gugalo: Adding my 5 cents )
`getUpgradeRequestsController` ?

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
DavDo: Is a request for screen-sharing considered an upgrade request?

andrews-moc: The matter is that `getOperatorRequestHandlerController` is still associating with HTTP requests to me. 🙂 
> Is a request for screen-sharing considered an upgrade request?

andrews-moc: > Is a request for screen-sharing considered an upgrade request?

andrews-moc: > Is a request for screen-sharing considered an upgrade request?
Not really, but I like `getUpgradeRequestsController()` or `getEngagementUpgradeController()`.

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizerController.kt
andrews-moc: They are different, aren't they? The current one logs an error. The new one logs incoming offer.

widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
andrews-moc: Or `getEngagementUpgradeOffersController()`.

widgetssdk/src/main/java/com/glia/widgets/operator/OperatorRequestHandlerContract.kt
andrews-moc: Let's add it as a TODO instead of a comment then. 😉

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizerController.kt
DavDo: Yes, I somehow missed the 2-nd log.
That's a case that should not happen because we're displaying media upgrade dialogs based on these values' presence, so on a step when the request is accepted, those parameters should stay the same as before.

widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
DavDo: Guys, I think in this case we need to change the name for all the components, controller, activitywatcher, and so on.
What do you think regarding these names? Or maybe this will inspire you to suggest a better one.

1. OperatorRequestOrchestratorController
2. OperatorCommandController
3. OperatorRequestManagementController
4. OperatorRequestController

I prefer the 4-th one.
I think an association with an HTTP request is related to the request + handler combination, so this name will reflect the whole functionality.

andrews-moc: Agree, 4th option is fine to me.

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/ActivityWatcherForCallVisualizerController.kt
andrews-moc: Okay, got it.

PR #953, opened 2024-02-14
[Backport] Deprecate Styles

DavDo:
**Jira issue:**

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/app/src/main/res/values/themes.xml
+++ b/app/src/main/res/values/themes.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources xmlns:tools="http://schemas.android.com/tools">
+<resources>
 
     <style name="Theme.GliaAndroidSDKWidgetsExample" parent="Theme.MaterialComponents.Light.NoActionBar">
         <item name="colorPrimary">@color/color_primary</item>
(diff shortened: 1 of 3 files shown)


Pelkar (approved):
LGTM. However I am a little concerned that we are deprecating all finer customisation. This will still be available to those who want to use those, but still. We might want to go over all the finer styling types and make sure we want to not have this from this point onward

andrews-moc (commented):
To double check, you are not just deprecating but removing right away? I was under impression that we agreed to mark as deprecated and remove from the docs.

DavDo:
Regarding deprecation, I think that's the only way atm because these styles are working partially(applying everything except colors), so yes, we must investigate if we need them at all or not.

DavDo:
> To double check, you are not just deprecating but removing right away? I was under impression that we agreed to mark as deprecated and remove from the docs.

No, we're just deprecating, I removed examples from the example app.

andrews-moc (approved):

andrews-moc:
> No, we're just deprecating, I removed examples from the example app.

Ah, got it, thanks.

PR #957, opened 2024-02-26
Fix CV confirmation dialog absence when returning from WebBrowser

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3140

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerActivityWatcher.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerActivityWatcher.kt
@@ -32,9 +32,10 @@ internal class CallVisualizerActivityWatcher(
             event.consumed -> Logger.d(TAG, "skipping.., event is already consumed")
             activity == null || activity.isFinishing -> Logger.d(TAG, "skipping.. activity is null or finishing")
             activity is WebBrowserActivity && state is ControllerState.DisplayConfirmationDialog -> Logger.d(TAG, "skipping.. WebBrowser is open")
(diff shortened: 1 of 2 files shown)


andrews-moc (commented):
Could you please provide the steps to reproduce? Also, ticket link opens 3138 instead of 3140.

DavDo:
> Could you please provide the steps to reproduce? Also, ticket link opens 3138 instead of 3140.

Sure. I'd like to remind you that this is happening on the refactoring branch only.

1. Open visitor code dialog
2. input code from the operator side and request CV
3. When the Confirmation dialog is opened click on link1 or link2
4. Link should be opened inside WebBrowserActivity
5. Return from the WebBrowser activity

**Expected Result:**
Confirmation dialog should be re-displayed

**Actual Result:**
Nothing Happens

andrews-moc (approved):

Pelkar:
I have an issue with this. 

I would argue that the correct implementation is to close the dialog, because the timeout on the backend side happens at 15 seconds.

Ofc the best UX would be to display a new dialog, that explains that the attempt has timed out and we need to try again.

The current implementation kind-of connects and visitor is not informed that the connection has failed

DavDo:
> I have an issue with this.
> 
> I would argue that the correct implementation is to close the dialog, because the timeout on the backend side happens at 15 seconds.
> 
> Ofc the best UX would be to display a new dialog, that explains that the attempt has timed out and we need to try again.
> 
> The current implementation kind-of connects and visitor is not informed that the connection has failed

These changes were made to align the behavior after refactoring with the current one. I like your suggestion, but I think it should be discussed separately and implemented as a separate improvement.

Pelkar (approved):

PR #971, opened 2024-03-07
Do not end engagement after authentication

andrews-moc:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3075

**What was solved?**

- Do not end engagement after authentication
- Show application bubble after authentication if there is ongoing engagement

**Release notes:**

 - [x] Feature
 - [x] Ignore

**Additional info:**


DavDo:
I think it's better to define a separate function to call from the authentication flow, the current solution is used in `clearVisitorSession` so this implementation will break that logic.

DavDo:
one more interesting question is what should happen in case a user authenticates /deauthenticates during queueing.

gugalo (commented):
It's only one last commit that needs reviewing, right?
"https://github.com/salemove/android-sdk-widgets/pull/971/commits/0bd83d93c7137b7586cde803ba61617f5eaf9f62"

andrews-moc:
@gugalo 
> It's only one last commit that needs reviewing, right?
Yes. I rebased the branch on top of development now and there is only one commit for review again. 😉

andrews-moc:
> During authentication, we need to keep the engagement state, and maybe the best way, in this case, is to just unsubscribe from the old engagement events inside the EngagementRepositoryImpl when a new engagement has arrived.

Honestly, it sounds like a bug-prone solution to me. It seems to me we don't need to complicate the solution with this suggestion. However, I would like to ask you for more points on why we need to keep the engagement state.

andrews-moc:
> one more interesting question is what should happen in case a user authenticates /deauthenticates during queueing.

I would expect that it queueing should stop. However, it doesn't. I will look into it on Monday. Also, included this case to exploratory testing scenarios.

andrews-moc:
> I think it's better to define a separate function to call from the authentication flow, the current solution is used in clearVisitorSession so this implementation will break that logic.

Good point, I will check it on Monday, thank you.

DavDo:
> Honestly, it sounds like a bug-prone solution to me. It seems to me we don't need to complicate the solution with this suggestion. However, I would like to ask you for more points on why we need to keep the engagement state.

I need to debug this to understand it more deeply, but my first thought is that the new engagement should update all the states because all states produced by `EngagementRepository` are strictly related to the ongoing engagement. However, even if we decide to reset all the states when a new engagement is emitted, it should happen inside this repository, because that's in the scope of the repository.

andrews-moc:
> I think it's better to define a separate function to call from the authentication flow, the current solution is used in `clearVisitorSession` so this implementation will break that logic.

Done. Please have a look.

> one more interesting question is what should happen in case a user authenticates /deauthenticates during queueing.
As I said, I think queueing should stop. Implemented this behavior in Widgets. However, I think it should be done on Core SDK level, the same as for current engagement. But to do this, queue ticket Core SDK interface needs queue ticket id. We don't keep it in Core. Thoughts?

andrews-moc:
> I need to debug this to understand it more deeply, but my first thought is that the new engagement should update all the states because all states produced by EngagementRepository are strictly related to the ongoing engagement. However, even if we decide to reset all the states when a new engagement is emitted, it should happen inside this repository, because that's in the scope of the repository.

Could you please check the latest fixup bd3661447af5d2f8b17b85dd18ce58ecbb868e8d and let me know if it covers your comment or not?

DavDo:
> > I think it's better to define a separate function to call from the authentication flow, the current solution is used in `clearVisitorSession` so this implementation will break that logic.
> 
> Done. Please have a look.
> 
> > one more interesting question is what should happen in case a user authenticates /deauthenticates during queueing.
> 
> As I've said already, I think queueing should stop. Implemented this behavior in Widgets. However, I think it should be done on Core SDK level, the same as for current engagement. But to do this, queue ticket Core SDK interface needs queue ticket id. We don't keep it in Core. Thoughts?

Yes, it's better to cancel queueing in widgets SDK. I think that's a very very corner case, and enough to cover here.

DavDo (approved):

DavDo:
> > I need to debug this to understand it more deeply, but my first thought is that the new engagement should update all the states because all states produced by EngagementRepository are strictly related to the ongoing engagement. However, even if we decide to reset all the states when a new engagement is emitted, it should happen inside this repository, because that's in the scope of the repository.
> 
> @DavDo Could you please check the latest fixup [bd36614](https://github.com/salemove/android-sdk-widgets/commit/bd3661447af5d2f8b17b85dd18ce58ecbb868e8d) and let me know if it covers your comment or not?

Overall looks good to me, and I'm sure that it should work, but I think it's safer to unsubscribe from the old engagement as soon as the new one is received.
```kotlin
//new function to unsubscribe from the old engagement
 private fun unsubscribeFromCurrentEngagement() {
        unsubscribeFromEvents(currentEngagement ?: return)
        currentEngagement = null
    }

    private fun handleOmniCoreEngagement(engagement: OmnicoreEngagement) {
        Logger.i(TAG, "Omnicore Engagement started")
        unsubscribeFromCurrentEngagement()//unsubscribe here
        currentEngagement = engagement
        operatorRepository.emit(engagement.state.operator)
        _engagementState.onNext(State.StartedOmniCore)

        subscribeToEngagementEvents(engagement)
        subscribeToEngagementMediaEvents(engagement.media)
        subscribeToEngagementChatEvents(engagement.chat)
        subscribeToScreenSharingEvents(engagement.screenSharing)
    }

    private fun handleCallVisualizerEngagement(engagement: OmnibrowseEngagement) {
        Logger.i(TAG, "Call Visualizer Engagement started")
        unsubscribeFromCurrentEngagement()//unsubscribe here
        currentEngagement = engagement
        operatorRepository.emit(engagement.state.operator)
        _engagementState.onNext(State.StartedCallVisualizer)

        subscribeToEngagementEvents(engagement)
        subscribeToEngagementMediaEvents(engagement.media)
        subscribeToScreenSharingEvents(engagement.screenSharing)
        //No need for chat events here
    }
```

andrews-moc:
> If you implement the above code, please make sure to update the `EngagementRepository` tests to check these changes.

@DavDo Added the changes you've suggested. Thank you. Working on the tests, don't seem to be easy to extend tests because new changes are in internal functions, looks like they are not tested directly, something indirect, I guess.

DavDo:
> > If you implement the above code, please make sure to update the `EngagementRepository` tests to check these changes.
> 
> @DavDo Added the changes you've suggested. Thank you. Working on the tests, don't seem to be easy to extend tests because new changes are in internal functions, looks like they are not tested directly, something indirect, I guess.

Sure, I can help you with that. It would be more efficient for me to investigate the tests instead of you having to spend time understanding the entire code in order to make the necessary modifications.

gugalo (approved):

PR #985, opened 2024-03-21
Increment project version to 2.3.4

sm-deployer:


andrews-moc:
Need to increment minor version instead of patch.

PR #990, opened 2024-03-26
Bugfix/mob 3206 broken download

Pelkar:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3206

**What was solved?**
Hotfix for file download on api 31+

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
@@ -878,7 +878,7 @@ internal class ChatView(context: Context, attrs: AttributeSet?, defStyleAttr: In
     }
 
     override fun onFileDownloadClick(file: AttachmentFile) {
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S || ContextCompat.checkSelfPermission(
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q || ContextCompat.checkSelfPermission(
                 context, Manifest.permission.WRITE_EXTERNAL_STORAGE
             ) == PackageManager.PERMISSION_GRANTED
         ) {
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
@@ -8,6 +8,7 @@ import android.content.Intent
 import android.content.pm.PackageManager
 import android.content.res.TypedArray
 import android.net.Uri
+import android.os.Build
 import android.provider.MediaStore
 import android.provider.Settings
 import android.text.Editable
@@ -877,7 +878,7 @@ internal class ChatView(context: Context, attrs: AttributeSet?, defStyleAttr: In
     }
 
     override fun onFileDownloadClick(file: AttachmentFile) {
-        if (ContextCompat.checkSelfPermission(
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S || ContextCompat.checkSelfPermission(
                 context, Manifest.permission.WRITE_EXTERNAL_STORAGE
             ) == PackageManager.PERMISSION_GRANTED
         ) {
--- a/widgetssdk/src/main/java/com/glia/widgets/filepreview/data/source/local/DownloadsFolderDataSource.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/filepreview/data/source/local/DownloadsFolderDataSource.java
@@ -171,30 +171,18 @@ public class DownloadsFolderDataSource {
 
     private Completable downloadFileToDownloadsOld(String fileName, InputStream inputStream) {
         return Completable.create(emitter -> {
-            OutputStream fos = null;
-            try {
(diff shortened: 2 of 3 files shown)


widgetssdk/src/main/java/com/glia/widgets/filepreview/data/source/local/DownloadsFolderDataSource.java
```diff
@@ -171,30 +171,18 @@ private Completable downloadFileToDownloadsAPI29(String fileName, String content
 
     private Completable downloadFileToDownloadsOld(String fileName, InputStream inputStream) {
         return Completable.create(emitter -> {
-            OutputStream fos = null;
-            try {
-                String imagesDir = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS).toString();
-                File file = new File(imagesDir, fileName);
-                fos = new FileOutputStream(file);
+            String imagesDir = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS).toString();
```
DavDo, line 174: I like this change 👍

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
```diff
@@ -877,7 +878,7 @@ internal class ChatView(context: Context, attrs: AttributeSet?, defStyleAttr: In
     }
 
     override fun onFileDownloadClick(file: AttachmentFile) {
-        if (ContextCompat.checkSelfPermission(
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S || ContextCompat.checkSelfPermission(
```
DavDo, line 881: This function calls the `downloadFileToDownloads` under the hood, which checks for `Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q`, I think it's better to have the same version check in this function.
```java
public Completable downloadFileToDownloads(String fileName, String contentType, InputStream inputStream) {
        if (fileName == null) return Completable.error(new NullPointerException("File name cannot be null"));

        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
            return downloadFileToDownloadsAPI29(fileName, contentType, inputStream);
        } else {
            return downloadFileToDownloadsOld(fileName, inputStream);
        }
    }
```

DavDo:
I would also recommend adding `maxSdkVersion` in manifest for the `READ_EXTERNAL_STORAGE` permission
```xml
<uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"
        android:maxSdkVersion="32" />
```

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
DavDo: https://developer.android.com/training/data-storage#scoped-storage
To give users more control over their files and to limit file clutter, apps that target Android 10 (API level 29) and higher are given scoped access into external storage, or scoped storage, by default. Such apps have access only to the app-specific directory on external storage, as well as specific types of media that the app has created.

andrews-moc: I'm not sure I'm following. @DavDo Based on your last comment, do you suggest to change the condition to `Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q || ContextCompat.checkSelfPermission`?

DavDo: Yes, permissions are not necessary for versions 29 and above, according to my interpretation of the mentioned functions.

DavDo:
> > I would also recommend adding `maxSdkVersion` in manifest for the `READ_EXTERNAL_STORAGE` permission
> > ```
> > <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"
> >         android:maxSdkVersion="32" />
> > ```
> 
> @DavDo Could you please explain why you think so? I'm asking, because [there is a warning](https://developer.android.com/guide/topics/manifest/uses-sdk-element#max) related to **android:maxSdkVersion**: `Warning: We don't recommend declaring this attribute. First, there is no need to set the attribute as a means of blocking deployment of your application onto new versions of the Android platform as they are released. By design, new versions of the platform are fully backward-compatible. Your application works properly on new versions, provided it uses only standard APIs and follows development best practices. Second, in some cases declaring the attribute can result in your application being removed from users' devices after a system update to a higher API level. Most devices on which your application is likely to be installed receive periodic system updates over the air, so consider their effect on your application before setting this attribute.`
> 
> I'm not sure if it's okay to have application removed from users' devices after a system update to a higher API level.

This is under the `<uses-sdk>` tag.
we need [this](https://developer.android.com/guide/topics/manifest/uses-permission-element) one.

AndriiHorishniiMOC (commented):
Nothing to add

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
Pelkar: Done

DavDo (approved):
LGTM

gugalo (approved):

andrews-moc (approved):

AndriiHorishniiMOC (approved):

PR #991, opened 2024-03-26
Implement mechanisms to check Firebase availability in widgets SDK

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3209

**What was solved?**

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.java
@@ -4,7 +4,6 @@ import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
 
-import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import com.glia.widgets.GliaWidgets;
@@ -116,18 +115,6 @@ public final class CallActivity extends FadeTransitionActivity {
     public void onUserInteraction() {
         super.onUserInteraction();
         callView.onUserInteraction();
-    }
-
-    @Override
-    protected void onActivityResult(int requestCode, int resultCode, Intent data) {
-        GliaWidgets.onActivityResult(requestCode, resultCode, data);
-        super.onActivityResult(requestCode, resultCode, data);
-    }
-
-    @Override
-    public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
-        super.onRequestPermissionsResult(requestCode, permissions, grantResults);
-        GliaWidgets.onRequestPermissionsResult(requestCode, permissions, grantResults);
     }
 
     private void startCall() {
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
@@ -1,11 +1,8 @@
 package com.glia.widgets.call
 
 import android.content.Context
-import android.content.Intent
 import android.content.res.Configuration
 import android.content.res.TypedArray
-import android.net.Uri
-import android.provider.Settings
 import android.util.AttributeSet
 import android.view.Gravity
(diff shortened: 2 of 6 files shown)


widgetssdk/src/main/java/com/glia/widgets/push/notifications/PushNotificationsSetUpUseCase.kt
```diff
@@ -0,0 +1,13 @@
...
+    operator fun invoke(): Boolean
+}
+
+internal class PushNotificationsSetUpUseCaseImpl(private val firebaseAvailabilityCheckManager: FirebaseAvailabilityCheckManager) :
+    PushNotificationsSetUpUseCase {
+    override fun invoke(): Boolean = firebaseAvailabilityCheckManager.run {
+        isFcmDependencyAvailable && isFcmServiceDeclaredInManifest
+    }
+
+}
```
andrews-moc, line 13: I suppose, you will add code that will set up push notifications later, right?

DavDo: No, will request notification permission before the first engagement request If push notifications are set up.

andrews-moc: In this case, I think that `PushNotificationsSetUpUseCase` and `PushNotificationsSetUpUseCaseImpl` name does not correspond to the actual work it does. I would rename it to something like `PushNotificationseAvailabilityUseCase`.

Also, do we really need 3 classes for this: FirebaseAvailabilityCheckManager, `PushNotificationsSetUpUseCase` and `PushNotificationsSetUpUseCaseImpl`? Maybe it's enough to have one use case class?

widgetssdk/src/main/java/com/glia/widgets/push/notifications/PushNotificationsSetUpUseCase.kt
```diff
@@ -0,0 +1,13 @@
+package com.glia.widgets.push.notifications
+
+internal interface PushNotificationsSetUpUseCase {
+    operator fun invoke(): Boolean
+}
+
+internal class PushNotificationsSetUpUseCaseImpl(private val firebaseAvailabilityCheckManager: FirebaseAvailabilityCheckManager) :
```
Pelkar, line 7: I don't think that we have so far used interface/impl pattern on usecases and I don't think we should start now.

DavDo: I used this approach for all the use cases for the engagement refactored layer, because this way we can mock it without any problems. But if you have something against this, I can change this one.

widgetssdk/src/main/java/com/glia/widgets/push/notifications/PushNotificationsSetUpUseCase.kt
DavDo: Makes sense, will rename it and remove the manager layer.

andrews-moc (approved):
Looks more clear and straightforward to me. The only thing that might be very good to add is - unit tests. 🙂

DavDo:
> Looks more clear and straightforward to me. The only thing that might be very good to add is - unit tests. 🙂

To be honest, I'm not sure if it is worth adding unit tests for this class, it is a pain since we have to mock a lot of system stuff to check one condition.

andrews-moc:
> > Looks more clear and straightforward to me. The only thing that might be very good to add is - unit tests. 🙂
> 
> To be honest, I'm not sure if it is worth adding unit tests for this class, it is a pain since we have to mock a lot of system stuff to check one condition.

Indeed, I know, okay. 👍

PR #996, opened 2024-04-01
Fixed narrow font on older devices

Pelkar:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2843

Fixed narrow font styling on API 24 and lower devices

**Release notes:**

 - [ ] Feature
 - [ ] Ignore
 - [X] Release notes (Is it clear from the description here?)


DavDo (approved):

andrews-moc (approved):
Strange case, indeed. I wish I knew the reason. Anyway, the fix should work.

AndriiHorishniiMOC (approved):

PR #1002, opened 2024-04-02
Undo the 'De-authenticate' button renaming in testing app to keep int…

andrews-moc:
**What was solved?**
I was too quick yesterday with renaming. After the recent sync with Igor we agreed to rename ONLY logs sent to Kibana and keep 'as is' all the rest.

That's why I'm rolling back this minor change.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**


Pelkar (approved):
This will then also fix acceptance tests

gugalo (approved):

PR #1006, opened 2024-04-08
Migrated dependencies to version catalogs

Pelkar:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2460

**What was solved?**
- Migrated dependencies to version catalogs  

**Release notes:**

 - [ ] Feature
 - [X] Ignore

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -101,28 +101,28 @@ afterEvaluate {
 }
 
 dependencies {
-    implementation "androidx.appcompat:appcompat:$appCompatVersion"
-    implementation "com.google.android.material:material:$materialVersion"
-    implementation "androidx.constraintlayout:constraintlayout:$constraintLayoutVersion"
-    implementation "androidx.navigation:navigation-fragment:$navVersion"
-    implementation "androidx.navigation:navigation-ui:$navVersion"
-    implementation "androidx.preference:preference:$preferenceVersion"
-    implementation platform("com.google.firebase:firebase-bom:$firebaseVersion")
-    implementation "com.google.firebase:firebase-messaging"
+    implementation libs.java.appcompat
+    implementation libs.java.material
+    implementation libs.java.constraintlayout
+    implementation libs.java.navigation.fargment
+    implementation libs.java.navigation.ui
+    implementation libs.java.preference
+    implementation platform(libs.firebase.bom)
+    implementation libs.firebase.messaging
 
     debugImplementation project(path: ':widgetssdk')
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -85,14 +85,16 @@ apply from: "${rootDir}/scripts/version-updater.gradle"
...
-  appCompatVersion = '1.3.1'
-  materialVersion = '1.6.1'
+  appCompatVersion = '1.6.1'
+  materialVersion = '1.11.0'
   constraintLayoutVersion = '2.1.4'
   navVersion = '2.3.5'
-  lifecycle_process = '2.3.1'
-  rxAndroid = '2.1.0'
-  rxJava = '2.2.6'
+  lifecycle_process = '2.7.0'
```
andrews-moc, line 97: Shall the whole `ext` section be also migrated to a version catalog approach?

Pelkar: Yes, I will cleanup

app/build.gradle
```diff
@@ -2,7 +2,7 @@ apply plugin: 'com.android.application'
 apply plugin: 'org.jetbrains.kotlin.android'
 
 android {
-    compileSdkVersion 33
+    compileSdkVersion 34
```
DavDo, line 5: As far as I'm informed, this property is deprecated and must be replaced to the `compileSdk 34`

widgetssdk/build.gradle
```diff
@@ -4,7 +4,7 @@ apply plugin: 'kotlin-parcelize'
 apply plugin: 'app.cash.paparazzi'
 
 android {
-  compileSdkVersion 33
+  compileSdkVersion 34
```
DavDo, line 7: We agreed to upgrade the compile SDK version to 33, so integrators aren't forced to upgrade to the latest version. Upgrading this to the latest available API version will force the integrators to upgrade their dependencies, especially `core-ktx` to the latest version. I'd prefer to keep the testing app compile and target SDK versions latest API version(34), and Widget and core sdk target and compile sdk versions one version lowe(33).

DavDo (changes requested):
I think it's a great initiative. 
However, I suggest handling the migration to RxJava RX3 and the migration of dependencies to the version catalogs separately. Also, I don't recommend upgrading the compile version of the widgets SDK, as it may cause certain libraries to be updated to their latest versions.

widgetssdk/build.gradle
andrews-moc: Yes, indeed. I was about to write this also.

andrews-moc (approved):
Just update the PR name for clarity. 😉

PR #1010, opened 2024-04-10
Add button in testing app to simulate end screen sharing button click from notifiations

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3238

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


gugalo (approved):
But why do we need it when we can just use the notification itself?

DavDo:
> But why do we need it when we can just use the notification itself?

Could you please look at [this](https://salemove.slack.com/archives/CHN6UB7UH/p1712748541153809) thread, I explained it in detail with screenshots and videos.

PR #1012, opened 2024-04-12
Fix Screen sharing starts with delay

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3240

**What was solved?**
After notification improvements, screen sharing started to be initiated with a huge ~15-second delay.

https://github.com/salemove/android-sdk-widgets/assets/19994238/fd3c1643-6217-48d9-82f5-b8a66cd8d816

As my investigation showed, that was related to the notification channel importance, which was set to a LOW level to prevent the notification from disturbing the user.

**Solution:**
- I returned the notification channel importance to the HIGH level for both call and screen-sharing notification channels.
- Put the `silent` flag to true for those notifications
> If {@code true}, silences this instance of the notification, regardless of the sounds or vibrations set on the notification or notification channel

So the notifications will appear at the top of the notification list, but still will be silent and will not pop up. And since the importance is HIGH system will initialize the media projection taking that into account.

https://github.com/salemove/android-sdk-widgets/assets/19994238/6c0ecb3e-8d5f-4042-80dc-1b5b3a15eab7

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/AndroidManifest.xml
+++ b/widgetssdk/src/main/AndroidManifest.xml
@@ -83,6 +83,7 @@
 
         <service
             android:name=".core.screensharing.MediaProjectionService"
+            android:exported="false"
             android:foregroundServiceType="mediaProjection" />
 
         <service android:name=".core.notification.NotificationRemovalService" />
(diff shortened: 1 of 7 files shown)


Pelkar (approved):
LGTM

andrews-moc (approved):

PR #1020, opened 2024-04-18
Fix Bubble pops up in call screen during configuration changes

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3256

I know this is not the best solution for this problem, but to do it more beautifully, a lot of refactoring is needed. So I decided to pick the most straightforward solution.

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


Pelkar (approved):

gugalo (commented):
I think the code looks fine.
But even though the ticket has overview of problematic behavior prior to the fix I don't understand why the old approach didn't work as expected.

DavDo:
> I think the code looks fine. But even though the ticket has overview of problematic behavior prior to the fix, I don't understand why the old approach didn't work as expected.

The problem is that we are using the root view of the current activity to determine an appropriate screen we need and decide to show or hide the bubble. However, since the `DialogHolderActivity` doesn't have a specific view, it is impossible to recognize `DialogHolderActivity` based on the root view name. Since the current logic of showing the chat bubble is based on this scheme, it will cause a lot of changes to refactor it and make it independent from the view name. So I decided to take the more straightforward approach and just added the root view to the activity.

AndriiHorishniiMOC (approved):
LGTM!

PR #1028, opened 2024-05-06
Feature/mob 3281 show bubble internally with overlay false

Pelkar:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3281

**What was solved?**
Even with setUseOverlay(false) we are still showing the chathead in the integrator application. Also added relevant javadoc changes, but those will hopefully be confirmed/modified by Kadri today.

**Release notes:**

 - [X] Feature
 - [ ] Ignore

**Additional info:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/core/chathead/domain/IsDisplayChatHeadUseCase.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/chathead/domain/IsDisplayChatHeadUseCase.kt
@@ -23,8 +23,12 @@ internal abstract class IsDisplayChatHeadUseCase(
 ) {
     abstract fun isDisplayBasedOnPermission(): Boolean
 
-    open operator fun invoke(viewName: String?): Boolean {
-        return isBubbleEnabled() && isDisplayBasedOnPermission() && isShowForEngagement(viewName)
+    open operator fun invoke(viewName: String?, internal: Boolean = false): Boolean {
+        return if (internal) {
+            isDisplayBasedOnPermission() && isShowForEngagement(viewName)
+        } else {
+            isBubbleEnabled() && isDisplayBasedOnPermission() && isShowForEngagement(viewName)
+        }
     }
 
     private fun isShowForEngagement(viewName: String?) =
--- a/widgetssdk/src/main/java/com/glia/widgets/core/chathead/domain/IsDisplayChatHeadUseCase.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/chathead/domain/IsDisplayChatHeadUseCase.kt
@@ -25,7 +25,7 @@ internal abstract class IsDisplayChatHeadUseCase(
 
     open operator fun invoke(viewName: String?, internal: Boolean = false): Boolean {
         return if (internal) {
-            isDisplayBasedOnPermission() && isShowForEngagement(viewName)
+            isShowForEngagement(viewName)
         } else {
             isBubbleEnabled() && isDisplayBasedOnPermission() && isShowForEngagement(viewName)
         }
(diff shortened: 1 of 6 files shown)


DavDo (approved):
LGTM

AndriiHorishniiMOC (approved):
Looks good, but it would be great to have a unit tests for those cases.

Pelkar:
> Looks good, but it would be great to have a unit tests for those cases.

Testing this is really complicated because 1. most of its components are untested 2. main logic is in an abstract class and would thus need testing in multiple final classes that override the abstract class. I can do this if we think this is warranted, but it will probably take an additional day or so

widgetssdk/src/main/java/com/glia/widgets/core/chathead/domain/IsDisplayChatHeadUseCase.kt
```diff
@@ -23,8 +23,12 @@ internal abstract class IsDisplayChatHeadUseCase(
 ) {
     abstract fun isDisplayBasedOnPermission(): Boolean
 
-    open operator fun invoke(viewName: String?): Boolean {
-        return isBubbleEnabled() && isDisplayBasedOnPermission() && isShowForEngagement(viewName)
+    open operator fun invoke(viewName: String?, internal: Boolean = false): Boolean {
```
andrews-moc, line 26: Could you please help me to understand, where is the value of `USE_OVERLAY` setting taken into account? I see that for `isDisplayApplicationChatHeadUseCase` `internal` is always `true`. I see that if not passed , internal is `false`. But what if permission for showing a bubble outside of the app is granted and `USE_OVERLAY` is `true` - will the bubble be shown outside of the app? I cannot find this case in the code. 🤔

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/core/chathead/domain/IsDisplayChatHeadUseCase.kt
andrews-moc: Based on my testing, with permission enabled and `USE_OVERLAY` set to false, bubble is still shown inside and outside the app. Which is not expected, as far as I understand, isn't it? As far as I understand, bubble should not be shown if `USE_OVERLAY` is false. Could you please test this case on your own manually?

Pelkar: I am guessing you had the overlay permission already set to allowed before changing the use_overlay parameter to false, right?

andrews-moc: Right.

Pelkar: I fixed the issue in the latest fixup

Pelkar: The result of the last fix is:
* Bubble is always shown internally
* Bubble is shown outside application with permission and USE_OVERLAY is true
* Even with granted permission, but USE_OVERLAY false the bubble will not be shown outside the application

andrews-moc:
@Pelkar I would like to take some time to review PR carefully once again because there are lots of [bubble test cases](https://glia.atlassian.net/wiki/spaces/ENG/pages/3305340989/Bubble). By the way, have you tested them after the changes?

Pelkar:
> bubble test cases

I have tested the feature extensively, but I will go through the specific test cases right now

Pelkar:
tested more thoroughly, one issue came up, for some reason there are two bubbles shown, all functionality is correct, but there is just a duplicate

andrews-moc:
> tested more thoroughly, one issue came up, for some reason there are two bubbles shown, all functionality is correct, but there is just a duplicate

It might be that one is a "local" bubble, the second - is "global".

andrews-moc:
I've prepared an alternative solution: https://github.com/salemove/android-sdk-widgets/pull/1052
Based on my thorough manual testing, it works well. ✅

DavDo:
> I've prepared an alternative solution: #1052 Based on my thorough manual testing, it works well. ✅

If it replaces this PR, closing this one makes sense.

PR #1032, opened 2024-05-15
Fix: Video views do not appear during a video call

AndriiHorishniiMOC:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3327

**What was solved?**
It seems that the startWithItem does not work well when called after some map functions. The combined visitor and operator media events can never occur in that case. Added dummy initial state that solves the issue.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/call/CallController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/call/CallController.kt
@@ -131,17 +131,14 @@ internal class CallController(
     // This is done to prevent errors such as "does not support one-way audio calls" only because the
     // other participant media state is passed few milliseconds late.
     private fun subscribeToMediaState() {
-        val dummyState = object : MediaState {
-            override fun getVideo(): Video? {
-                return null
-            }
-            override fun getAudio(): Audio? {
-                return null
-            }
+        val initialState = object : MediaState {
+            override fun getVideo(): Video? = null
+            override fun getAudio(): Audio? = null
         }
+
         Flowable.combineLatest(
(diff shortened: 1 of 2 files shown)


widgetssdk/src/main/java/com/glia/widgets/call/CallController.kt
```diff
@@ -129,9 +131,17 @@ internal class CallController(
     // This is done to prevent errors such as "does not support one-way audio calls" only because the
     // other participant media state is passed few milliseconds late.
     private fun subscribeToMediaState() {
+        val dummyState = object : MediaState {
```
Pelkar, line 134: dummyState seems something that should not be in production code. Can we have something like defaultState?

andrews-moc: It can be renamed to `initialState` or `initialMediaState`, I suppose.

andrews-moc, line 134: It can be renamed to initialState or initialMediaState, I suppose.

AndriiHorishniiMOC: fixed d61ebfc1b85167b43c48705c7bb76b994037a50f

AndriiHorishniiMOC: fixed d61ebfc1b85167b43c48705c7bb76b994037a50f

DavDo (approved):

andrews-moc (approved):

Pelkar (approved):

PR #1046, opened 2024-06-12
Show call notification only on Android APIs prior to 31

andrews-moc:
**Jira issue:**
[Crashes in Glia Android SDK (Monument)](https://glia.atlassian.net/browse/MOB-3358)

**What was solved?**
In my opinion, call notifications are not needed on Android APIs 31+ (Android 12+) because of the built-in Android [privacy indicators](https://source.android.com/docs/core/permissions/privacy-indicators). 

As a result, there is no need to start `NotificationRemovalService` on 31+ APIs. 

So the following crash should not happen: `Not allowed to start service Intent ... NotificationRemovalService ... app is in background` anymore.

See more details in [Slack](https://salemove.slack.com/archives/CHN6UB7UH/p1718174711873439?thread_ts=1717669165.207099&cid=CHN6UB7UH).

**Release notes:**
Fix `Not allowed to start service Intent ... NotificationRemovalService ... app is in background` crash that sometimes happen on Android 12+.

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**
Screenshot 2024-06-12 at 09 41 58

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/core/notification/device/NotificationManager.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/notification/device/NotificationManager.kt
@@ -58,6 +58,11 @@ internal class NotificationManager(private val applicationContext: Application) 
     }
 
     override fun showAudioCallNotification() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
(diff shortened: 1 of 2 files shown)


gugalo (approved):
👍

PR #1051, opened 2024-06-20
Default queues

AndriiHorishniiMOC:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3403

**What was solved?**
Added ability to start an engagement without passing queue ID. In that case, a ticket will be assigned to all default queues.

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**


app/src/main/res/layout/main_fragment.xml
```diff
@@ -74,6 +74,15 @@
...
 
+            <androidx.appcompat.widget.SwitchCompat
+                android:id="@+id/default_queues_switch"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:text="Default queues "
+                app:layout_constraintTop_toBottomOf="@id/settings_button"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintEnd_toEndOf="parent" />
+
```
andrews-moc, line 85: Could you please add a screenshot to PR description?
In my opinion, it's better to move this switch somewhere to the settings screen to avoid Main screen overloading. In fact, this switch is a setting, so it's logical to put it there, I think.

widgetssdk/src/main/java/com/glia/widgets/chat/ChatActivity.java
```diff
@@ -52,13 +53,13 @@ public final class ChatActivity extends FadeTransitionActivity {
      * Creates and fills out Intent for starting ChatActivity
      * @param context   - Context object
      * @param contextId - Context asset ID
-     * @param queueId   - Queue ID
+     * @param queueId   - Queue ID or `null` to request the default queues
```
gugalo, line 56: NIT: I would change **request** to **use**
```suggestion
     * @param queueId   - Queue ID or `null` to use the default queues
```

gugalo: Same in other places

gugalo (approved):
👍

widgetssdk/src/main/java/com/glia/widgets/di/GliaCoreImpl.kt
```diff
@@ -86,19 +86,15 @@ internal class GliaCoreImpl : GliaCore {
         Glia.getQueues(requestCallback)
     }
 
-    override fun queueForEngagement(queueId: String, visitorContextAssetId: String?, callback: Consumer<GliaException?>) {
-        Glia.queueForEngagement(queueId, visitorContextAssetId, callback)
-    }
-
     override fun queueForEngagement(
-        queueId: String,
+        queueIds: Array<String>,
```
DavDo, line 90: Maybe this should have been written in core sdk PR, but anyways, why we're using `Array` instead of more flexible and recommended by Kotlin `List`

DavDo (approved):
LGTM!

widgetssdk/src/main/java/com/glia/widgets/di/GliaCoreImpl.kt
AndriiHorishniiMOC: But you approved it https://github.com/salemove/android-sdk/pull/681

DavDo: Yes, but that was `Java` code.
This is not critical, but I would prefer changing such things to Kotlin's recommended way during migration.

app/src/main/res/layout/main_fragment.xml
AndriiHorishniiMOC: I decided sometimes we want to start one engagement with a specific queue and immediately another without.

AndriiHorishniiMOC: Of course, I can move it to the setting screen.

AndriiHorishniiMOC: OK, I decided to move this toggle to the settings.

widgetssdk/src/main/java/com/glia/widgets/di/GliaCoreImpl.kt
AndriiHorishniiMOC: Let's discuss and implement this in a separate task.

andrews-moc (approved):

AndriiHorishniiMOC:
!squash

PR #1053, opened 2024-07-02
Fix useOverlays property having different values during configuration

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3416
https://glia.atlassian.net/browse/MOB-3405

**What was solved?**

**The Problem:**
In certain environments, particularly cross-platform ones, the UI can become stuck in the engagement-request state while actually no request is being made.

**The Reason:**
We had different default values for `useOverlays` property, so it was returning different values during initialization steps, as a result to start a queuing we were waiting for the overlay dialog, but in call or chat screen we were receiving `false` and skipping the overlay dialog.

**The Solution**
I left only one default value for the `useOverlays` property inside `GliaSdkConfigurationManager` and move the overlay check from the `DecideOnQueuingUseCase` init block to avoid concurrency issues and mark it completed manually from the chat or call screens.

**Release notes:**
We resolved the problem with concurrency that caused the SDK to become stuck in a `connecting` state without making an engagement request.

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
@@ -223,11 +223,11 @@ internal class ChatView(context: Context, attrs: AttributeSet?, defStyleAttr: In
         companyName: String?,
         queueId: String?,
         visitorContextAssetId: String?,
-        useOverlays: Boolean = false,
+        useOverlays: Boolean? = null,
         screenSharingMode: ScreenSharing.Mode? = null,
         chatType: ChatType = ChatType.LIVE_CHAT
     ) {
-        Dependencies.getSdkConfigurationManager().isUseOverlay = useOverlays
+        useOverlays?.also { Dependencies.getSdkConfigurationManager().isUseOverlay = it }
         Dependencies.getSdkConfigurationManager().screenSharingMode = screenSharingMode
         dialogCallback?.also { dialogController?.addCallback(it) }
         controller?.initChat(companyName, queueId, visitorContextAssetId, chatType)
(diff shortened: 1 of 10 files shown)


Pelkar (approved):
LGTM

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
```diff
@@ -223,11 +223,11 @@ internal class ChatView(context: Context, attrs: AttributeSet?, defStyleAttr: In
         companyName: String?,
         queueId: String?,
         visitorContextAssetId: String?,
-        useOverlays: Boolean = false,
+        useOverlays: Boolean? = null,
         screenSharingMode: ScreenSharing.Mode? = null,
         chatType: ChatType = ChatType.LIVE_CHAT
     ) {
-        Dependencies.getSdkConfigurationManager().isUseOverlay = useOverlays
+        useOverlays?.also { Dependencies.getSdkConfigurationManager().isUseOverlay = it }
```
andrews-moc, line 230: Hm, so engagement value for `useOverlays` will override the SDK initialisation value. 🤔

DavDo: Yes!

Nothing has changed in this sense.

widgetssdk/src/main/java/com/glia/widgets/core/configuration/GliaSdkConfigurationManager.java
```diff
@@ -107,7 +117,6 @@ public GliaSdkConfiguration createWidgetsConfiguration() {
                 .companyName(companyName)
                 .screenSharingMode(screenSharingMode)
                 .useOverlay(useOverlay)
-                .manualLocaleOverride(manualLocaleOverride)
```
andrews-moc, line 110: Is removing of this line somehow related to bubble?

widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
andrews-moc: Ah, indeed, you are right.

andrews-moc (approved):

PR #1056, opened 2024-07-04
Get rid of Kotlin reflection

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3418

**What was solved?**
Replaced the Kotlin reflection-based `Merge` functionality with manually written ones to make the SDK less sensitive to proguard on the integrator side.

Related issues could be found in - https://glia.atlassian.net/browse/MOB-3337

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/view/dialog/base/DialogViewFactory.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/view/dialog/base/DialogViewFactory.kt
@@ -16,13 +16,13 @@ import com.glia.widgets.view.dialog.screensharing.ScreenSharingDialogViewInflate
 import com.glia.widgets.view.dialog.screensharing.VerticalScreenSharingDialogViewInflater
 import com.glia.widgets.view.dialog.update.UpgradeDialogViewInflater
 import com.glia.widgets.view.dialog.update.VerticalUpgradeDialogViewInflater
-import com.glia.widgets.view.unifiedui.deepMerge
+import com.glia.widgets.view.unifiedui.nullSafeMerge
 import com.glia.widgets.view.unifiedui.theme.AlertDialogConfiguration
 import com.glia.widgets.view.unifiedui.theme.UnifiedTheme
 
 internal class DialogViewFactory(context: Context, uiTheme: UiTheme, unifiedTheme: UnifiedTheme?) {
     private val configuration: AlertDialogConfiguration = uiTheme.alertTheme(context).run {
-        copy(theme = this.theme.deepMerge(unifiedTheme?.alertTheme)!!)
+        copy(theme = theme nullSafeMerge unifiedTheme?.alertTheme)
     }
 
     private val layoutInflater: LayoutInflater = LayoutInflater.from(context)
--- a/widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/Merge.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/Merge.kt
@@ -1,10 +1,14 @@
 package com.glia.widgets.view.unifiedui
 
-import kotlin.reflect.KClass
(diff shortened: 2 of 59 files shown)


AndriiHorishniiMOC (commented):
Because there are so many changes, it is difficult to check whether every merge method is correct. But it looks good.
The unit tests are still needed.

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/Merge.kt
```diff
@@ -1,10 +1,14 @@
 package com.glia.widgets.view.unifiedui
 
-import kotlin.reflect.KClass
-import kotlin.reflect.KParameter
-import kotlin.reflect.KProperty1
-import kotlin.reflect.full.declaredMemberProperties
-import kotlin.reflect.full.primaryConstructor
+/**
+ * Interface for classes that can be merged with another instance of the same class
```
andrews-moc, line 4: Lets maybe add that it is for data classes?
with another instance of the same class => with another instance of the same **data** class

widgetssdk/src/main/java/com/glia/widgets/view/dialog/base/DialogViewFactory.kt
```diff
@@ -16,13 +16,13 @@ import com.glia.widgets.view.dialog.screensharing.ScreenSharingDialogViewInflate
...
 import com.glia.widgets.view.dialog.update.VerticalUpgradeDialogViewInflater
-import com.glia.widgets.view.unifiedui.deepMerge
+import com.glia.widgets.view.unifiedui.nullSafeMerge
 import com.glia.widgets.view.unifiedui.theme.AlertDialogConfiguration
 import com.glia.widgets.view.unifiedui.theme.UnifiedTheme
 
 internal class DialogViewFactory(context: Context, uiTheme: UiTheme, unifiedTheme: UnifiedTheme?) {
     private val configuration: AlertDialogConfiguration = uiTheme.alertTheme(context).run {
-        copy(theme = this.theme.deepMerge(unifiedTheme?.alertTheme)!!)
+        copy(theme = theme nullSafeMerge unifiedTheme?.alertTheme)
```
andrews-moc, line 25: Could you please explain, for my understanding, why do you use `nullSafeMerge` instead of `merge` here?

widgetssdk/src/test/java/com/glia/widgets/view/unifiedui/extensions/MergeKtTest.kt
andrews-moc, line 1: Are you going to remove the code disabled in this class?

widgetssdk/src/main/java/com/glia/widgets/view/dialog/base/DialogViewFactory.kt
DavDo: because of the return type, `nullSafeMerge` returns non-null value.

DavDo:
> Because there are so many changes, it is difficult to check whether every merge method is correct. But it looks good. The unit tests are still needed.

I'm adding tests, but it will cover the merge function functionality, not every `Mergable.merge` function implementation.

andrews-moc:
I will approve as soon as you finish with unit tests because, anyway, I will review them. 😉

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/Merge.kt
DavDo: There is no restriction on the usage of any other classes, which is why I preferred to not mention the `data` class here.

andrews-moc (approved):

AndriiHorishniiMOC (approved):

DavDo:
!squash

PR #1057, opened 2024-07-05
Deprecate UiTheme to make its removal smooth

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3421

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
@@ -39,7 +39,12 @@ public class GliaWidgets {
     /**
      * Use with {@link android.os.Bundle} to pass in a {@link UiTheme} as a navigation argument when
      * navigating to {@link com.glia.widgets.chat.ChatActivity}
+     *
+     * @deprecated While UiTheme can still be used for UI customization,
+     * we strongly encourage adopting remote configurations {@link GliaWidgetsConfig.Builder#setUiJsonRemoteConfig(String)}.
+     * The remote configurations approach is more versatile and better suited for future development.
      */
+    @Deprecated
     public static final String UI_THEME = "ui_theme";
     /**
      * Use with {@link android.os.Bundle} to pass in the name of your company as a navigation
--- a/widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsConfig.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsConfig.kt
@@ -3,7 +3,7 @@ package com.glia.widgets
 import android.content.Context
 import com.glia.androidsdk.SiteApiKey
 import com.glia.androidsdk.screensharing.ScreenSharing
-import com.glia.widgets.helper.Logger.i
+import com.glia.widgets.helper.Logger
 import com.glia.widgets.helper.TAG
 
 /**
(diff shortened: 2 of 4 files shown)


andrews-moc (commented):
I think we need to add `@deprecated` with a comment and `@Deprecated` for each element. Like it was discussed recently [here](https://github.com/salemove/android-sdk/pull/730#discussion_r1662294725).

AndriiHorishniiMOC:
Icons customization is still available only on UiTheme.

DavDo:
> Icons customization is still available only on UiTheme.

Yes, but with resource id's only, so I think it would still be possible with XML arguments.

DavDo:
> I think we need to add `@deprecated` with a comment and `@Deprecated` for each element. Like it was discussed recently [here](https://github.com/salemove/android-sdk/pull/730#discussion_r1662294725).



> I think we need to add `@deprecated` with a comment and `@Deprecated` for each element. Like it was discussed recently [here](https://github.com/salemove/android-sdk/pull/730#discussion_r1662294725).

I think this applies to Java only. 
The @Deprecated annotation in Kotlin supports a message parameter, and optionally, replaceWith parameter to suggest a replacement. And in addition to this, Kotlin uses `KDoc` for documentation that doesn't have a `@deprecated` tag.

AndriiHorishniiMOC (approved):

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -39,7 +39,12 @@ public class GliaWidgets {
     /**
      * Use with {@link android.os.Bundle} to pass in a {@link UiTheme} as a navigation argument when
      * navigating to {@link com.glia.widgets.chat.ChatActivity}
+     *
+     * @deprecated While UiTheme can still be used for UI customization,
+     * we strongly encourage adopting remote configurations {@link GliaWidgetsConfig.Builder#setUiJsonRemoteConfig(String)}.
+     * This newer approach is more versatile and better suited for future development.
```
andrews-moc, line 45: NIT (I'm not insisting): This newer approach => Remote configurations approach.

andrews-moc (approved):
Does it make sense to ask tech writes to review the text? I don't insist, it's up to you.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -39,7 +39,12 @@ public class GliaWidgets {
     /**
      * Use with {@link android.os.Bundle} to pass in a {@link UiTheme} as a navigation argument when
      * navigating to {@link com.glia.widgets.chat.ChatActivity}
+     *
+     * @deprecated While UiTheme can still be used for UI customization,
+     * we strongly encourage adopting remote configurations {@link GliaWidgetsConfig.Builder#setUiJsonRemoteConfig(String)}.
+     * Remote configurations approach is more versatile and better suited for future development.
```
1toomas, line 45: Please update the same sentence everywhere.
(Seems that this is again one of such cases (previously discussed [here](https://github.com/salemove/android-sdk/pull/730#discussion_r1660855313)) that the generator does not render for the [SDK docs](http://android-sdk-docs.glia.com.s3-website-us-east-1.amazonaws.com/widgets/javadoc/com/glia/widgets/GliaWidgets.html).)
```suggestion
     * The remote configurations approach is more versatile and better suited for future development.
```

1toomas (approved):
One tiny suggestion.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
DavDo: We can migrate to an `HTML` type of documentation, that is more Kotlin-friendly and has a way better UI and user experience.

image
image

DavDo:
!squash

PR #1060, opened 2024-07-10
Automatic update of string on locale change

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3378

**What was solved?**
- Default SDK company name is now ""
- Widgets SDK company name from configurations has lower priority than remote company name also used as the fallback if remote is unavailable or empty
- If the first run and locales are not yet available Widgets will use local strings but once the remote becomes available the UI is reloaded with remote strings

N.B: I am still doing minor fixes but would appreciate early reviews.

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**
[Screen_recording_20240710_164420.webm](https://github.com/salemove/android-sdk-widgets/assets/10000880/3f0e29a4-e842-43ef-a004-d71d8f6053a7)


DavDo (commented):
Added some thoughts.

widgetssdk/src/main/java/com/glia/widgets/StringProviderImpl.kt
```diff
@@ -1,41 +1,85 @@
...
 
-internal class StringProviderImpl(private val resourceProvider: ResourceProvider) : StringProvider {
+@Parcelize
+data class LocaleString(
+    @StringRes val stringKey: Int,
+    val values: Array<StringKeyPair> = emptyArray()
+): Parcelable {
+
+    companion object {
+        fun from(@StringRes stringKey: Int, vararg values: StringKeyPair = emptyArray()) = LocaleString(stringKey, arrayOf(*values))
```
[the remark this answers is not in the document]
gugalo: My first attempt was also to use the secondary contractor but for some reason during compilation it return an error 
image

But after your other suggestions I have tried it with list instead of array and it seems to work in this case with secondary constractor.

widgetssdk/src/main/java/com/glia/widgets/StringProviderImpl.kt
```diff
@@ -1,41 +1,85 @@
...
 import com.glia.androidsdk.GliaException
-import com.glia.widgets.helper.Logger
 import com.glia.widgets.helper.ResourceProvider
+import kotlinx.parcelize.Parcelize
 
-internal class StringProviderImpl(private val resourceProvider: ResourceProvider) : StringProvider {
+@Parcelize
+data class LocaleString(
+    @StringRes val stringKey: Int,
+    val values: Array<StringKeyPair> = emptyArray()
```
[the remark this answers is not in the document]
gugalo: So would a list be better? Or what should I use?
I think list also uses arrays internally

widgetssdk/src/main/java/com/glia/widgets/StringProviderImpl.kt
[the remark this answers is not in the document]
DavDo: Wow! Indeed, We have `arrays` in both cases, I think if you change `array` to a `list` it might be possible with `values.toList`.

widgetssdk/src/main/java/com/glia/widgets/StringProviderImpl.kt
```diff
@@ -1,41 +1,85 @@
...
+    override fun registerLocaleListener(key: String, listener: () -> Unit) {
+        localeListeners[key] = listener
+        startListener()
+    }
+
+    override fun unregisterLocaleListener(key: String) {
+        localeListeners.remove(key)
+    }
+
+    // Priority for remote company name but not found or no locales it will foll back to hardcoded one or "" if no hardcoded one
```
andrews-moc, line 60: NIT: foll => fall

widgetssdk/src/main/java/com/glia/widgets/call/CallView.kt
```diff
@@ -368,7 +373,7 @@ internal class CallView(
             if (isActivated) activatedContentDescription else notActivatedContentDescription
 
         floatingActionButton.isActivated = isActivated
-        floatingActionButton.contentDescription = stringProvider.getRemoteString(contentDescription)
+        floatingActionButton.setLocaleContentDescription(contentDescription)
```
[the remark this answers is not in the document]
gugalo: I agree with you. Karl also suggested re-loading the whole activity instead of separate strings. So I have shortly looked into it but that also will have some problems.

I want to finish this solution for now and then try to improve it. One option is to improve on the current view/view-state/view-controller logic that is already there but not yet on the level where it is easy just to reload the locales.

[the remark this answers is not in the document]
DavDo: I think reloading the whole activity is a bad idea. We could replace texts but control/trigger them from a single place to minimize the chance of missing the upgrade step. 
Like I described it here - https://github.com/salemove/android-sdk-widgets/pull/1060#discussion_r1672369458. 
If you're interested in more details, we can have a short discussion on this topic.

gugalo:
Everything mentioned before should be resolved in last 3 fixups

widgetssdk/src/main/java/com/glia/widgets/locale/StringKeyPair.kt
```diff
@@ -0,0 +1,53 @@
+package com.glia.widgets.locale
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+/**
```
DavDo, line 6: why not make these classes internal instead of adding the @hide tag?

widgetssdk/src/main/java/com/glia/widgets/helper/ViewExtensions.kt
```diff
@@ -145,3 +150,86 @@ internal val View.layoutInflater: LayoutInflater get() = LayoutInflater.from(thi
...
+    val localeListener = {
+        // It is strange that it should already be UI thread according to:
+        // https://github.com/salemove/android-sdk/blob/4dbf361ecf108b478b997e415734240778af2e41/androidSdk/src/main/java/com/glia/androidsdk/internal/locale/LocaleManagerInternal.kt#L71
+        // But without `runOnUiThread` it sometimes does not work
+        context.asActivity()?.runOnUiThread {
+            listener.invoke(localeManager.getString(stringKey, values.toList()))
+        }
+    }
+    localeListener.invoke()
+    val disposable = localeManager.getLocaleObservable().subscribe(
```
DavDo, line 219: have you tried to add `subscribeOn(AndroidSchedulers.mainThread())` here as well?

widgetssdk/src/main/java/com/glia/widgets/helper/ViewExtensions.kt
```diff
@@ -145,3 +150,86 @@ internal val View.layoutInflater: LayoutInflater get() = LayoutInflater.from(thi
...
+}
+
+private fun View.registerLocaleListener(@StringRes stringKey: Int, vararg values: StringKeyPair, listener: (String) -> Unit) {
+    val localeManager = Dependencies.getLocaleProvider()
+    val localeListener = {
+        // It is strange that it should already be UI thread according to:
+        // https://github.com/salemove/android-sdk/blob/4dbf361ecf108b478b997e415734240778af2e41/androidSdk/src/main/java/com/glia/androidsdk/internal/locale/LocaleManagerInternal.kt#L71
+        // But without `runOnUiThread` it sometimes does not work
+        context.asActivity()?.runOnUiThread {
+            listener.invoke(localeManager.getString(stringKey, values.toList()))
```
DavDo, line 215: can safely remove `.invoke`, since the listener is `non-null`

widgetssdk/src/main/java/com/glia/widgets/helper/ViewExtensions.kt
```diff
@@ -145,3 +150,86 @@ internal val View.layoutInflater: LayoutInflater get() = LayoutInflater.from(thi
...
+        }
+    }
+    localeListener.invoke()
+    val disposable = localeManager.getLocaleObservable().subscribe(
+        { localeListener.invoke() },
```
DavDo, line 220: can safely remove `.invoke`, since the localeListener is `non-null`

DavDo, line 220: What do you think about this solution, it is simpler and will avoid listener invocation in case of the same value.
```suggestion
   val disposable = localeManager.getLocaleObservable()
        .startWithItem("stub")
        .map { localeManager.getString(stringKey, values.toList()) }
        .distinctUntilChanged()
        .subscribe { post { listener(it) } }
```

widgetssdk/src/main/java/com/glia/widgets/helper/ViewExtensions.kt
```diff
@@ -145,3 +150,86 @@ internal val View.layoutInflater: LayoutInflater get() = LayoutInflater.from(thi
...
+        { /* no-op */ }
+    )
+
+    addOnAttachStateChangeListener(object : View.OnAttachStateChangeListener {
+        override fun onViewAttachedToWindow(v: View) {
+            // TODO: to add listener back here or not to add, that is the question
```
DavDo, line 226: Is there a case when the same view can be re-attached to Window?

widgetssdk/src/main/java/com/glia/widgets/StringKeyPair.kt
```diff
@@ -1,23 +1,34 @@
...
- * Retrieve custom locales strings that should have placeholders replaced with real values.
- */
-data class StringKeyPair(val key: StringKey, val value: String)
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
 
-/**
- * List placeholders for [StringKeyPair].
- */
+@Deprecated("Feature removed for internal use only")
```
andrews-moc, line 6: This wording is ambiguous to me. "Feature is for internal use only" sounds more clear in my opinion.

widgetssdk/src/main/java/com/glia/widgets/StringProvider.java
```diff
@@ -3,12 +3,18 @@
...
 import androidx.annotation.StringRes;
 
+import com.glia.androidsdk.Glia;
+import com.glia.androidsdk.locale.LocaleManager;
+
 /**
- * Retrieving string from custom locales.
- *
- * See [Android Custom Locales guide][https://docs.glia.com/glia-mobile/docs/android-custom-locales] for more details.
+ * @depricated
```
andrews-moc, line 10: depricated => deprecated

widgetssdk/src/main/java/com/glia/widgets/StringKeyPair.kt
andrews-moc: Or lets at least transform into two sentences: "Feature was removed. It's for internal use only now."

widgetssdk/src/main/java/com/glia/widgets/base/FadeTransitionActivity.kt
```diff
@@ -34,4 +35,13 @@ open class FadeTransitionActivity : AppCompatActivity() {
         overridePendingTransition(android.R.anim.fade_in, android.R.anim.fade_out)
     }
 
+    fun setTitle(locale: LocaleString?) {
+        if (locale == null) {
+            super.setTitle(null)
+            return
+        }
+
+        // TODO: handle locale updated
```
andrews-moc, line 44: Is it a leftover, or it's really out of the current scope?

widgetssdk/src/main/java/com/glia/widgets/locale/StringKeyPair.kt
gugalo: They are used in java public class. So I first need to convert it to Kotlin then I will be able to make it also internal because it is not supposed to be public and then I will be able to make these classes internal as well. Plan to do that but in a separate PR.

widgetssdk/src/main/java/com/glia/widgets/helper/ViewExtensions.kt
gugalo: Yes, but I don't know how common it is. It's probably a corner case. I remember (from very long ago) that it can happen when using multiple `Fragment`s.

widgetssdk/src/main/java/com/glia/widgets/base/FadeTransitionActivity.kt
gugalo: Actually forgot to implement

widgetssdk/src/main/java/com/glia/widgets/helper/ViewExtensions.kt
gugalo: Nice one 👍

gugalo:
Another round of fixes.
Everything mentioned above should be fixed or responded to.

widgetssdk/src/main/java/com/glia/widgets/base/FadeTransitionActivity.kt
gugalo: Now should be fixed

andrews-moc (approved):

AndriiHorishniiMOC (approved):

DavDo (approved):
LGTM

gugalo:
Fixed snapshot tests and android instrumented tests. Please review latest changes.

widgetssdk/src/main/java/com/glia/widgets/helper/ViewExtensions.kt
```diff
@@ -145,3 +152,82 @@ internal val View.layoutInflater: LayoutInflater get() = LayoutInflater.from(thi
...
+        .subscribe(
+//            { post { listener(it) } },
+            { listener(it) },
+            { /* no-op */ }
+        )
+
+    addOnAttachStateChangeListener(object : View.OnAttachStateChangeListener {
+        override fun onViewAttachedToWindow(v: View) {
+            // TODO: to add listener back here or not to add, that is the question
+            // TODO: also not clear if that is even possible without causing a memory leak
```
andrews-moc, line 225: Are these todos still needed?

widgetssdk/src/main/java/com/glia/widgets/locale/LocaleProvider.kt
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    private fun getLocaleManager(): LocaleManager {
+        if (this::localeManager.isInitialized.not()) {
+            localeManager = Glia.getLocaleManager()
+            localeManager.onLocaleChanged { localeCode, error ->
+                if (error == null && localeCode != null) {
+                    localeEmitter.onNext(localeCode)
+                } else {
+                    // TODO: report error or do nothing since it is already reporting it on Core side?
```
andrews-moc, line 60: I think it makes sense to decide on this and remove todo, doesn't it?

widgetssdk/src/main/java/com/glia/widgets/locale/LocaleProvider.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    @OpenForTesting
+    open fun getStringInternal(stringKey: Int, values: List<StringKeyPair> = emptyList()): String {
+        val key = resourceProvider.getResourceKey(stringKey)
+        return try {
+            getLocaleManager().getRemoteString(key)?.let {
+                values.fold(it, ::replaceInsertedValues).run(::cleanupRemainingReferences)
+            } ?: resourceProvider.getString(stringKey, *values.map { pair -> pair.value }.toTypedArray())
+        } catch (e: GliaException) {
+            // TODO: report error or is there even possibility that exception will be thrown?
```
andrews-moc, line 99: I think it makes sense to add a debug log at lease.

widgetssdk/src/main/java/com/glia/widgets/StringKeyPair.kt
```diff
@@ -1,23 +1,44 @@
...
+    @Deprecated("Feature was removed. It's for internal use only now.")
     BADGE_VALUE("badgeValue"),
+    @Deprecated("Feature was removed. It's for internal use only now.")
     FILE_SENDER("fileSender"),
-    BUTTON_TITLE("buttonTitle")
+    @Deprecated("Feature was removed. It's for internal use only now.")
+    BUTTON_TITLE("buttonTitle");
+
+    init {
+        Logger.logDeprecatedClassUse(TAG)
```
andrews-moc, line 42: Great that you added this log. 👍

widgetssdk/src/main/java/com/glia/widgets/StringKeyPair.kt
```diff
@@ -1,23 +1,44 @@
...
+import kotlinx.parcelize.Parcelize
 
-/**
- * List placeholders for [StringKeyPair].
- */
+@Deprecated("Feature was removed. It's for internal use only now.")
+@Parcelize
+data class StringKeyPair(val key: StringKey, val value: String): Parcelable {
+    init {
+        Logger.logDeprecatedClassUse(TAG)
```
andrews-moc, line 12: Great that you added this log. 👍

DavDo (approved):
LGTM

gugalo:
!merge

PR #1062, opened 2024-07-17
Default queues for Secure Messaging

AndriiHorishniiMOC:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3438

**What was solved?**
If no queueId is passed to start messaging, then the SDK uses default queues supporting messaging and are not closed to send secure message to.

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**


widgetssdk/src/test/java/com/glia/widgets/chat/controller/ChatControllerTest.kt
```diff
@@ -64,6 +67,8 @@ import org.mockito.kotlin.verify
 import org.mockito.kotlin.whenever
 import org.robolectric.RobolectricTestRunner
 
+private const val s = "CompanyName"
```
andrews-moc, line 70: Although it's a unit test, I would suggest using a meaningful name instead of "s".

AndriiHorishniiMOC: I have no idea what it is :)

widgetssdk/src/main/java/com/glia/widgets/messagecenter/MessageCenterController.kt
```diff
@@ -59,17 +62,26 @@ internal class MessageCenterController(
...
     override fun setView(view: MessageCenterContract.View) {
         this.view = view
+    }
 
+    override fun initialize() {
         if (isAuthenticatedUseCase()) {
             initComponents()
         } else {
             dialogController.showUnauthenticatedDialog()
+            Logger.w(TAG, "Secure Messaging is unavailable because the visitor is not authenticated.")
```
andrews-moc, line 84: It seems to me that 'info' level is enough for this event. Visitor just has to authenticate and it's a normal case/flow. It's not something that we should definitely keep an eye on.

AndriiHorishniiMOC: Maybe you're right.
@EgorovEI, what do you think? Let's change it to `info`.

gugalo (approved):

andrews-moc (approved):

DavDo:
Should we review this pull request, or will it change the way we enqueue to default queues?

PR #1065, opened 2024-08-01
Dismiss CV dialog on timeout

AndriiHorishniiMOC:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3456

**What was solved?**
The incoming engagement request dialog is dismissed if the request is timeout.

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

 - [x] Did you add logging beneficial for troubleshooting of customer issues?

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/engagement/domain/OnIncomingEngagementRequestTimeoutUseCase.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/engagement/domain/OnIncomingEngagementRequestTimeoutUseCase.kt
@@ -1,0 +1,18 @@
+package com.glia.widgets.engagement.domain
+
+import com.glia.androidsdk.EngagementRequest.Outcome
+import com.glia.widgets.engagement.EngagementRepository
+import io.reactivex.rxjava3.core.Flowable
+
+internal interface OnIncomingEngagementRequestTimeoutUseCase {
+    operator fun invoke(): Flowable<Unit>
+}
+
+internal class OnIncomingEngagementRequestTimeoutUseCaseImpl(
+    private val engagementRepository: EngagementRepository
+) : OnIncomingEngagementRequestTimeoutUseCase {
+
+    override fun invoke(): Flowable<Unit> = engagementRepository.engagementOutcome
+        .filter { it == Outcome.TIMEOUT }
+        .map { Unit }
+}
--- a/widgetssdk/src/test/java/com/glia/widgets/engagement/EngagementRepositoryTest.kt
+++ b/widgetssdk/src/test/java/com/glia/widgets/engagement/EngagementRepositoryTest.kt
@@ -4,6 +4,7 @@ import android.LOGGER_PATH
 import android.app.Activity
 import com.glia.androidsdk.Engagement
 import com.glia.androidsdk.Engagement.MediaType
+import com.glia.androidsdk.EngagementRequest
 import com.glia.androidsdk.Glia
 import com.glia.androidsdk.GliaException
 import com.glia.androidsdk.IncomingEngagementRequest
@@ -407,6 +408,7 @@ class EngagementRepositoryTest {
     @Test
     fun `engagementRequest will emit value when engagement is requested`() {
         val incomingEngagementRequest: IncomingEngagementRequest = mockk()
+        every { incomingEngagementRequest.on(EngagementRequest.Events.OUTCOME, any()) } returns Unit
         val incomingEngagementRequestTest = repository.engagementRequest.test()
 
         engagementRequestCallbackSlot.captured.accept(incomingEngagementRequest)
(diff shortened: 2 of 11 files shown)


widgetssdk/src/test/java/com/glia/widgets/engagement/EngagementRepositoryTest.kt
```diff
@@ -477,6 +488,31 @@ class EngagementRepositoryTest {
...
+        val engagementOutcomeTest = repository.engagementOutcome.test()
+
+        engagementRequestCallbackSlot.captured.accept(incomingEngagementRequest)
+        outcomeSlot.captured.accept(EngagementRequest.Outcome.ACCEPTED)
+
+        engagementOutcomeTest.assertNoErrors().assertValue(EngagementRequest.Outcome.ACCEPTED)
+    }
+
+    @Test
+    fun `unsubscribe from engagement outcome event will happen when outcome is requested`() {
```
andrews-moc, line 505: "when outcome is requested" or "accepted"?

andrews-moc: In my understanding the outcome is received by SDK but it's not explicitly requested. Am I wrong?

andrews-moc (approved):
Good job 👍

widgetssdk/src/test/java/com/glia/widgets/engagement/EngagementRepositoryTest.kt
AndriiHorishniiMOC: We can get it from the `visitor_state`. To do it, we need to subscribe for the `outcome`. But after we get it, we don't need the subscription anymore.

AndriiHorishniiMOC:
I made a small change, but it is related to the DialogHolderActivity, so you may have something to say about my decision.
We need it if the state changes when the Browser Activity is opened or if an application is in the background. In those cases, the DialogHolderActivity can stuck and block user interaction with the application when they return to the application.

widgetssdk/src/main/java/com/glia/widgets/engagement/domain/OnIncomingEngagementRequestTimeoutUseCase.kt
```diff
@@ -0,0 +1,18 @@
...
+    operator fun invoke(): Flowable<Unit>
+}
+
+internal class OnIncomingEngagementRequestTimeoutUseCaseImpl(
+    private val engagementRepository: EngagementRepository
+) : OnIncomingEngagementRequestTimeoutUseCase {
+
+    override fun invoke(): Flowable<Unit> = engagementRepository.engagementOutcome
+        .filter { it == Outcome.TIMEOUT }
+        .map { }
```
DavDo, line 17: What is the benefit of converting this to `Unit` and then ignoring?
I think since we are ignoring the value, we can avoid this step.

widgetssdk/src/main/java/com/glia/widgets/helper/DialogHolderActivity.kt
```diff
@@ -26,6 +26,11 @@ internal class DialogHolderActivity : FadeTransitionActivity() {
         setContentView(DialogHolderView(this))
     }
 
+    override fun onPause() {
+        super.onPause()
+        finish()
```
DavDo, line 31: This seems to me a bit dangerous.
Calling one lifecycle method in another one could lead to unpredictable results.
Lets try to fix the cause of the problem.

andrews-moc (commented):
+1 to Davit's comments

widgetssdk/src/main/java/com/glia/widgets/helper/DialogHolderActivity.kt
AndriiHorishniiMOC: I guess I found a better solution.

AndriiHorishniiMOC: https://github.com/salemove/android-sdk-widgets/pull/1065/files#diff-e12c01bfa33793e79d7ece617032b133f97f172f3502dd1e2726770ea52135ebR37

widgetssdk/src/main/java/com/glia/widgets/engagement/domain/OnIncomingEngagementRequestTimeoutUseCase.kt
DavDo: No, I mean we ignore the result when subscribing to this `Flowable`, so no need to map it to `Unit`

widgetssdk/src/main/java/com/glia/widgets/helper/DialogHolderActivity.kt
DavDo: Looks good!

DavDo (approved):

PR #1070, opened 2024-08-13
Fix hardcoded company name

gugalo:
**Jira issue:**

**What was solved?**
There was a problem that the company name specified in SDK configs on `init` was ignored even when no remote was loaded.

**Release notes:**

 - [x] Feature: Part of custom locales improvement changes
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/di/Dependencies.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/di/Dependencies.java
@@ -64,7 +64,7 @@ public class Dependencies {
     public static void onAppCreate(Application application) {
         schedulers = new GliaWidgetsSchedulers();
         resourceProvider = new ResourceProvider(application.getBaseContext());
-        localeProvider = new LocaleProvider(resourceProvider, sdkConfigurationManager.getCompanyName());
+        localeProvider = new LocaleProvider(resourceProvider);
         notificationManager = new NotificationManager(application);
         DownloadsFolderDataSource downloadsFolderDataSource = new DownloadsFolderDataSource(application);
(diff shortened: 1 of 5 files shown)


andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/di/Dependencies.java
```diff
@@ -196,6 +196,7 @@ public static void init(@NonNull GliaWidgetsConfig gliaWidgetsConfig) {
         controllerFactory.init();
         repositoryFactory.getEngagementRepository().initialize();
         sdkConfigurationManager.fromConfiguration(gliaWidgetsConfig);
+        localeProvider.setCompanyName(gliaWidgetsConfig.companyName);
```
AndriiHorishniiMOC, line 199: It isn't possible to call the `internal` function from the Kotlin class in the Java class.

AndriiHorishniiMOC (changes requested):

widgetssdk/src/main/java/com/glia/widgets/di/Dependencies.java
gugalo: It is interesting that it compiled and worked just fine when I test it locally

gugalo: Also the class itself is also internal but lint is only complaining about the method

gugalo: Fixed

PR #1073, opened 2024-08-22
Show Snackbar if an engagement request is timed out

AndriiHorishniiMOC:
MOB-3495

**Jira issue:**

**What was solved?**
The Snackbar is displayed to inform visitors if an engagement request is timed out.

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/controller/CallVisualizerController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/controller/CallVisualizerController.kt
@@ -26,6 +26,7 @@ internal interface CallVisualizerContract {
         object DisplayVisitorCodeDialog : State
         data class DisplayConfirmationDialog(val links: ConfirmationDialogLinks) : State
         object DismissDialog : State
+        object ShowTimeoutSnackBar : State
         object CloseHolderActivity : State
         data class OpenWebBrowserScreen(val title: LocaleString, val url: String) : State
     }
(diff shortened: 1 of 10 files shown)


andrews-moc (approved):

DavDo:
There is a conflict now after I merged my [fix](https://github.com/salemove/android-sdk-widgets/pull/1074).
I'll wait for the resolution of this conflict.

AndriiHorishniiMOC:
@DavDo, conflicts were resolved.

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/controller/CallVisualizerController.kt
```diff
@@ -107,7 +108,7 @@ internal class CallVisualizerController(
 
     private fun onIncomingEngagementRequestTimeout() {
         dialogController.dismissCVEngagementConfirmationDialog()
-        closeHolderActivity()
+        _state.onNext(CallVisualizerContract.State.ShowTimeoutSnackBar)
```
DavDo, line 111: I would recommend keeping the `closeHolderActivity()` call here before pushing a new state instead of moving it to the ActivityWatcher.
Each state should be responsible for its action like closing holder activity, showing snackbar, etc.

DavDo (approved):
Overall looks good! Added one minor comment.

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/controller/CallVisualizerController.kt
AndriiHorishniiMOC: Unfortunately, it doesn't work. It isn't a stack of events. Only the last event will apply if the application was in the background during the timeout.

AndriiHorishniiMOC: I can only suggest renaming the state to `CloseHolderActivityAndShowTimeoutSnackBar`.

andrews-moc (approved):
I haven't tested manually, the changes are looking good to me.

PR #1080, opened 2024-09-04
Upgrade compile, target and AGP versions

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3551

**What was solved?**
- Upgrade Widgets SDK Compile SDK version to 34
- Upgrade Testing app Compile and target SDK versions to 35
- Upgrade the AGP version to the latest version(8.6)
- Upgrade third-party dependencies
- Upgrade the Core SDK dependency version to 1.8.0

**Release notes:**
- Android 15(Target SDK 35) support.
- Upgrade the minimum supported compileSDK to 34

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**


EgorovEI (approved):

gugalo (approved):
Are these changes backword compatible?
Can you publish to locale maven and use them with CortexFinacial without incrementing the gradle version or changing the target SDK to API 35?

DavDo:
> Are these changes backward compatible? Can you publish to locale maven and use this Widget version with CortexFinacial without incrementing the gradle version or changing the target SDK to API 35?

these changes should be backwards compatible, but I will check before merging them.

DavDo:
there are mostly no differences between the old and new snapshots, except the Operator attachment-related cases where the Paparazzi is having some problems with the Formatter so the `KB` somehow gets replaced with the `kilobyteShort` but in runtime, there is no such problem. Also, upgrading Paparazzi to the latest version fixes this, but since it is bumping the Kotlin version to 1.9 for an entire project, we can’t do it now.
image

DavDo:
It seems all the failing snapshots are related to the paparazzi-related changes, so we will probably face the same problems when upgrading the Paparazzi version with Kotlin being 1.9

PR #1085, opened 2024-09-16
Refactor configurations

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3587

**What was solved?**
- Migrate Call and Chat Activities to Kotlin
- Refactor CallConfiguration
- Refactor EngagementConfiguration

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


AndriiHorishniiMOC (approved):
LGTM!

widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.kt
```diff
@@ -0,0 +1,162 @@
...
+        startActivity(newIntent)
+    }
+
+    private fun navigateToWebBrowser(title: LocaleString, url: String) {
+        val newIntent = intent(this, title, url)
+        startActivity(newIntent)
+    }
+
+    internal companion object {
+        private val TAG: String = CallActivity::class.java.simpleName
```
gugalo, line 150: I think this line can be replaced with the static import of `Logger.kt` -> `Any.TAG`

widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.kt
```diff
@@ -0,0 +1,162 @@
...
+    private fun startCall() {
+        val engagementConfiguration = callConfiguration.engagementConfiguration!!
+        callView.startCall(
+            engagementConfiguration.companyName!!,
+            engagementConfiguration.queueIds,
+            engagementConfiguration.contextAssetId,
+            engagementConfiguration.screenSharingMode!!,
+            callConfiguration.isUpgradeToCall,
+            callConfiguration.mediaType
+        )
```
gugalo, line 129: Could you check if it is possible to remove force null unwraps? I think at least first one should be possible to make cleaner.

widgetssdk/src/main/java/com/glia/widgets/call/CallActivityIntentHelper.kt
```diff
@@ -6,33 +6,28 @@ import androidx.appcompat.app.AppCompatActivity
...
-        val mediaType = intent.getSerializableExtra(GliaWidgets.MEDIA_TYPE) as Engagement.MediaType?
+        val engagementConfiguration = EngagementConfiguration(intent)
+        val mediaType = intent.getSerializableExtraCompat<Engagement.MediaType>(GliaWidgets.MEDIA_TYPE)
         val isUpgradeToCall = intent.getBooleanExtra(GliaWidgets.IS_UPGRADE_TO_CALL, false)
-        return CallConfiguration.Builder()
-            .setEngagementConfiguration(engagementConfiguration)
-            .setMediaType(mediaType)
-            .setIsUpgradeToCall(isUpgradeToCall)
-            .build()
+        return CallConfiguration(engagementConfiguration = engagementConfiguration, mediaType = mediaType, isUpgradeToCall = isUpgradeToCall)
```
gugalo, line 31: I think that method property names could be dropped here because they provide 0 benefits to the code readability in this case

widgetssdk/src/main/java/com/glia/widgets/chat/ChatActivity.kt
```diff
@@ -0,0 +1,186 @@
...
+    private fun backToCallScreen() {
+        startActivity(CallActivity.getIntent(applicationContext, defaultCallConfiguration))
+        finish()
+    }
+
+    /**
+     * Just commenting to avoid lint error, this should be removed in next tasks
+     */
+    companion object {
+        private val TAG: String = ChatActivity::class.java.simpleName
```
gugalo, line 144: Can replace TAG with the static import of Logger class TAG implementation

widgetssdk/src/main/java/com/glia/widgets/chat/ChatActivity.kt
```diff
@@ -0,0 +1,186 @@
...
+            contextId: String?,
+            queueIds: List<String>,
+            chatType: ChatType?
+        ): Intent {
+            return Intent(context, ChatActivity::class.java)
+                .putExtra(GliaWidgets.CONTEXT_ASSET_ID, contextId)
+                .putExtra(GliaWidgets.QUEUE_IDS, ArrayList(queueIds))
+                .putExtra(GliaWidgets.COMPANY_NAME, "Legacy company")
+                .putExtra(GliaWidgets.CHAT_TYPE, chatType as Parcelable?)
+        }
```
gugalo, line 184: I know the is clearly out of the scope of this PR, but could you extract this into `CallActivityIntentHelper` class and then rename that class to something like `WidgetsActivityIntentHelper`?
But ignore this suggestion if you are busy with something else or it ends up as bigger refactoring

widgetssdk/src/main/java/com/glia/widgets/core/configuration/EngagementConfiguration.kt
gugalo, line 1: I am lost about how you pulled it out. This file was 126 lines and now is 35 and nothing broke without moving it anywhere else ))
👍

gugalo (approved):
Added small change requists that are not directly impactful on the PR

widgetssdk/src/main/java/com/glia/widgets/chat/ChatActivity.kt
DavDo: This and other Intent-related things will be refactored in the next PR.

widgetssdk/src/main/java/com/glia/widgets/core/configuration/EngagementConfiguration.kt
DavDo: Advantages of Kotlin)

widgetssdk/src/main/java/com/glia/widgets/chat/ChatActivity.kt
andrews-moc: It definitely makes sense to have Intent creation functions in one place. What I think is confusing - we have `CallActivityIntentHelper` but don't have `ChatActivityIntentHelper`.

widgetssdk/src/main/java/com/glia/widgets/chat/ChatActivity.kt
```diff
@@ -0,0 +1,186 @@
...
+
+    override fun onDestroy() {
+        chatView.onDestroyView()
+        Logger.i(TAG, "Destroy Chat screen")
+        super.onDestroy()
+    }
+
+    private fun createEngagementConfiguration(intent: Intent): EngagementConfiguration = EngagementConfiguration(intent)
+
+    //TODO: Check why theme attribute is not anymore used
```
andrews-moc, line 124: Is it a leftover?

widgetssdk/src/main/java/com/glia/widgets/core/configuration/EngagementConfiguration.kt
```diff
@@ -6,121 +6,30 @@ import com.glia.widgets.GliaWidgets
...
 
-        fun contextAssetId(contextAssetId: String?): Builder {
-            this.contextAssetId = contextAssetId
-            return this
-        }
-
-        @Deprecated(
-            "While UiTheme can still be used for UI customization, we strongly encourage adopting remote configurations(GliaWidgetsConfig.Builder.setUiJsonRemoteConfig). " +
-                "The remote configurations approach is more versatile and better suited for future development."
-        )
```
andrews-moc, line 81: I think we should keep this deprecation notice, shouldn't we?

widgetssdk/src/main/java/com/glia/widgets/chat/ChatActivity.kt
DavDo: No, it is more for the next tickets in this story.

widgetssdk/src/main/java/com/glia/widgets/core/configuration/EngagementConfiguration.kt
DavDo: No, I removed it because this is an internal class. I didn't touch any public deprecated methods.
Since these features will be released in version 3.0, we should recheck and remove any remaining deprecations. However, this should only be done after completing the Navigator approach, as several public interfaces that contain deprecations will also be removed.

widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.kt
DavDo: I prefer to leave this as it is because I will completely change this approach to a more straightforward one during the 'Navigator' implementation.

DavDo:
Guys, I added some suggestions from your review, but merging this PR without re-requesting review because this is a simple preparation PR, and most of the changed classes should be completely removed during this process.

PR #1086, opened 2024-09-17
Entry Widget (embedded view)

AndriiHorishniiMOC:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3478

**What was solved?**
Added Entry Widget (embedded view). By this PR added just UI.

An interface for integrators will be added in the next tasks.
The Unified UI support and UI improvements will be added in the next tasks.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetView.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetView.kt
@@ -1,0 +1,44 @@
+package com.glia.widgets.entrywidget
+
+import android.content.Context
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import com.glia.widgets.R
+import com.glia.widgets.di.Dependencies
+import com.glia.widgets.entrywidget.adapter.EntryWidgetAdapter
+import com.glia.widgets.entrywidget.adapter.EntryWidgetItemDecoration
+import com.glia.widgets.helper.getDrawableCompat
+
+internal class EntryWidgetView(
+    context: Context
+) : RecyclerView(context, null, 0), EntryWidgetContract.View {
+    var onDismissListener: (() -> Unit)? = null
+
+    private lateinit var controller: EntryWidgetContract.Controller
+    private val adapter = EntryWidgetAdapter().also { setAdapter(it) }
+
+    init {
+        layoutManager = LinearLayoutManager(context)
+        getDrawableCompat(R.drawable.bg_separator_gray)?.let {
+            addItemDecoration(EntryWidgetItemDecoration(it))
+        }
+        adapter.onItemClickListener = {
+            controller.onItemClicked(it)
+        }
+
+        setController(Dependencies.controllerFactory.entryWidgetController)
+    }
+
+    override fun setController(controller: EntryWidgetContract.Controller) {
+        this.controller = controller
+        controller.setView(this)
+    }
+
+    override fun showItems(items: List<EntryWidgetContract.ItemType>) {
+        adapter.items = items
+    }
+
+    override fun dismiss() {
+        onDismissListener?.invoke()
+    }
+}
--- a/widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/di/ControllerFactory.java
@@ -17,6 +17,8 @@ import com.glia.widgets.core.dialog.DialogContract;
 import com.glia.widgets.core.dialog.DialogController;
 import com.glia.widgets.engagement.completion.EngagementCompletionContract;
 import com.glia.widgets.engagement.completion.EngagementCompletionController;
+import com.glia.widgets.entrywidget.EntryWidgetContract;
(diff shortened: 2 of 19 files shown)


DavDo:
Will the loading be a group of gray views, or will it have a shimmer effect?

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetView.kt
```diff
@@ -0,0 +1,44 @@
...
+import androidx.recyclerview.widget.RecyclerView
+import com.glia.widgets.R
+import com.glia.widgets.di.Dependencies
+import com.glia.widgets.entrywidget.adapter.EntryWidgetAdapter
+import com.glia.widgets.entrywidget.adapter.EntryWidgetItemDecoration
+import com.glia.widgets.helper.getDrawableCompat
+
+internal class EntryWidgetView(
+    context: Context
+) : RecyclerView(context, null, 0), EntryWidgetContract.View {
```
andrews-moc, line 14: There is this concern in the project plan:
>If we use RecyclerView for showing the engagement items (Chat, Audio, Video, Messaging), integrators [would have to use](https://armanco.medium.com/how-to-use-vertical-recyclerview-inside-scrollview-in-android-b72337285517) androidx.core.widget.NestedScrollView as a parent view to handle embedded view height change. We cannot limit the integrators. Taking into account that the number of items is final and not that big, we can show them using ConstraintLayout. In this case, embedded view height change will be animated and handled properly.

You mentioned this yesterday, during our sync but I forgot to clarify if the following case will be handled properly (see last step):
- Entry Widget displays two items
- media types availability is updated and one more item becomes available
- will Entry Widget be able to redraw items and occupy more space?

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetView.kt
```diff
@@ -0,0 +1,44 @@
...
+    init {
+        layoutManager = LinearLayoutManager(context)
+        getDrawableCompat(R.drawable.bg_separator_gray)?.let {
+            addItemDecoration(EntryWidgetItemDecoration(it))
+        }
+        adapter.onItemClickListener = {
+            controller.onItemClicked(it)
+        }
+
+        setController(Dependencies.controllerFactory.entryWidgetController)
```
andrews-moc, line 29: Does it make sense to call `setController` `before adapter.onItemClickListener` to prevent any chance of `controller.onItemClicked` call before controller is set?

andrews-moc (commented):
Good job, just have a look at my comments, please. 👍

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetView.kt
DavDo: Will it help if we disable scroll in RecyclerView?

AndriiHorishniiMOC: I didn't add animation, but the view should resize with no problem.

https://github.com/user-attachments/assets/79c0887e-e8ae-4529-87f6-fb6b6d77259a

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetView.kt
AndriiHorishniiMOC: But without the controller, there is nothing to display. So, no click events should be before `setController`.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetView.kt
DavDo: but what about the embedded view?

AndriiHorishniiMOC: @DavDo The bottom sheet includes exactly the embedded view. The only difference is the border, which will be added after the default theme is implemented.

AndriiHorishniiMOC:
@DavDo I didn't add animation to this PR because, honestly, I don't know how this animation should look. In my view, it is better to implement it in a separate PR.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetView.kt
andrews-moc: Bottom sheet view container adjusts the height, that's good. However, could you please add the embedded view to the MainFragment hierarchy and test the same resizing there?

DavDo (approved):
LGTM

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetView.kt
andrews-moc: Hm, looks like it's working well. 👍 
We will still need to brainstorm various ideas how integrators can use it and test.

andrews-moc (approved):

AndriiHorishniiMOC:
!squash

PR #1089, opened 2024-09-19
Create new internal layer responsible for opening screens

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3589

**What was solved?**
- Create ActivityLauncher
- Rename FilePreview... classes to ImagePreview...

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


AndriiHorishniiMOC (approved):
Looks good.
But you also need to refactor snapshot tests.

widgetssdk/src/main/java/com/glia/widgets/filepreview/ui/ImagePreviewActivity.kt
```diff
@@ -69,7 +63,7 @@ internal class FilePreviewActivity : AppCompatActivity(), FilePreviewContract.Vi
         enableEdgeToEdge()
         super.onCreate(savedInstanceState)
         SimpleWindowInsetsAndAnimationHandler(binding.root, binding.toolbar)
-        Logger.i(TAG, "Create File Preview screen")
+        Logger.i(TAG, "Create Image Preview screen")
```
andrews-moc, line 66: It's better to leave existing message because it's aligned with iOS platform...

widgetssdk/src/main/java/com/glia/widgets/filepreview/ui/ImagePreviewActivity.kt
```diff
@@ -132,8 +126,8 @@ internal class FilePreviewActivity : AppCompatActivity(), FilePreviewContract.Vi
     }
 
     override fun onDestroy() {
-        Logger.i(TAG, "Destroy File Preview screen")
-        filePreviewController?.onDestroy()
+        Logger.i(TAG, "Destroy Image Preview screen")
```
andrews-moc, line 129: The same here - it's better to leave the existing wording because it's aligned with iOS.

andrews-moc (approved):
👍 I don't have other comments except the logging messages.

widgetssdk/src/main/java/com/glia/widgets/filepreview/ui/ImagePreviewActivity.kt
DavDo: But what benefits can we expect from aligning this with iOS?

I think it's better to log it in a way that matches the current Android screen.
I renamed this activity and related items because it is only for previewing images. I think it is confusing to name it FilePreview.

Also, why not just rename iOS either?

andrews-moc: > what benefits can we expect from aligning this with iOS?
There was a goal to have as much the same messages as possible. To be able to search in Kibana and compare iOS/Android behavior in some cases.

> I think it's better to log it in a way that matches the current Android screen.
Yes, agree. Let's ask iOS to change it?

DavDo: I will create a ticket for iOS.

PR #1091, opened 2024-09-23
Downgrade 'one-way audio calls' error to info

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3583

**What was solved?**

This problame consists of multiple parts:
- At the moment it should be impossible to start one-way audio
- Visitor and Operator media states are received from different APIs and are async
- ChatController does not care about visitor media state

As a result, this error is 'expected' so I decided to downgrade it to an info log

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


DavDo (approved):
👍

andrews-moc (approved):

PR #1094, opened 2024-09-26
Mob 3604 devs want to implement business logic for the engagement launcher approach

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3604

**What was solved?**
- Remove all the configurations in favor of ConfigurationManager
- Get rid of configurations passed through Activity extras
- Move extra keys to internal class
- Make activities internal
- Minor refactoring

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -38,99 +38,6 @@
...
-     * If set to true then the SDK will ask for overlay permissions and try to always show the navigation bubble outside
-     * the application. However, it will be shown only if the user has accepted the permissions.
-     * If false, then overlay permission is not requested and the navigation bubble is shown when the application is active.
-     * Default value is true.
-     *
-     * @deprecated Use {@link com.glia.widgets.GliaWidgetsConfig#isEnableBubbleOutsideApp() and
-     * {@link GliaWidgetsConfig#isEnableBubbleInsideApp()}
-     */
-    @Deprecated
-    public static final String USE_OVERLAY = "use_overlay";
```
gugalo, line 103: Remove of depricated properties should be a separate commit and even separate PR

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -208,6 +115,7 @@ private static void setupQueueIds(@Nullable List<String> queueIds) {
      *
      * <p>This method is no-op for other non-Glia triggered results.</p>
      */
+    //TODO check the possibilities to remove this method
```
gugalo, line 118: Is this going to be done in this PR or should we create a separate jira ticket for this?

widgetssdk/src/main/java/com/glia/widgets/UiTheme.kt
```diff
@@ -30,10 +30,6 @@ import kotlinx.parcelize.Parcelize
         "The remote configurations approach is more versatile and better suited for future development."
 )
 data class UiTheme(
-    /**
-     * Text to be shown on the top of the app bar in the chat
-     */
-    val appBarTitle: String? = null,
```
gugalo, line 36: This should also be a separate PR. Also, the release notes should be mentioned in the PR section as breaking changes. Same for the removed deprecated properties so that we could communicate this in the 3.0.0 release and migration guides

widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.kt
```diff
@@ -40,9 +42,8 @@ import kotlin.properties.Delegates
  * startActivity(intent);
 </pre> *
  */
-class CallActivity : FadeTransitionActivity() {
+internal class CallActivity : FadeTransitionActivity() {
```
gugalo, line 45: 👍 Finally! I tried to do this twice only to find there are still public methods/classes that require this to be public. Very cool that it is internal at last!

widgetssdk/src/main/java/com/glia/widgets/chat/ChatActivity.kt
```diff
@@ -95,14 +60,11 @@ class ChatActivity : FadeTransitionActivity() {
...
-        chatView.startChat(
-            engagementConfiguration.companyName,
-            engagementConfiguration.queueIds,
-            engagementConfiguration.contextAssetId,
-            engagementConfiguration.screenSharingMode,
-            engagementConfiguration.chatType ?: ChatType.LIVE_CHAT
-        )
+
+        val chatType = intent.getIntExtra(ExtraKeys.CHAT_TYPE, -1).takeIf { it != -1 }
+            ?.let { ChatType.values()[it] } ?: ChatType.LIVE_CHAT
```
gugalo, line 65: Could you extract this code to some to some `IntentExtentions` class with a method name `fun Intent.getChatType` and also extract any calls to `intent.putExtra(CHAT_TYPE, ChatType.SECURE_MESSAGING as Parcelable)` to the same class with a name `intent.putChatType()`. 

I know that technically your solution is completely safe even for any future changes to enums but I really would still prefer to have such logic in one place.

widgetssdk/src/main/java/com/glia/widgets/call/CallActivity.kt
```diff
@@ -89,7 +79,10 @@ class CallActivity : FadeTransitionActivity() {
         callView.setOnNavigateToWebBrowserListener(onNavigateToWebBrowserListener)
 
         if (savedInstanceState == null) {
-            startCall()
+            val mediaType = intent.getIntExtra(ExtraKeys.MEDIA_TYPE, -1)
+                .takeIf { it != -1 }?.let { Engagement.MediaType.values()[it] }
```
gugalo, line 83: Commented similar logic below

gugalo (changes requested):
You did great work but could youplease separate your changes to smaller PRs. It is very hard to review such big PRs

app/src/main/java/com/glia/exampleapp/Application.java
```diff
@@ -43,35 +35,4 @@ private void initFirebase() {
...
-            @Override
-            public void onActivityStopped(@NonNull Activity activity) {}
-
-            @Override
-            public void onActivitySaveInstanceState(@NonNull Activity activity, @NonNull Bundle outState) {}
-
-            @Override
-            public void onActivityDestroyed(@NonNull Activity activity) {}
-        });
-    }
```
andrews-moc, line 76: As far as I see, this was needed to init Glia Widgets using a deep link. Are you sure it is not needed anymore?

DavDo: Yes, I moved this logic inside MainFragment and this way we have more control on it

app/src/main/java/com/glia/exampleapp/MainFragment.kt
```diff
@@ -60,6 +55,20 @@ class MainFragment : Fragment() {
...
+        PreferenceManager.getDefaultSharedPreferences(requireContext())
+    }
+
+    private val engagementLauncher: EngagementLauncher by lazy {
+        ensureInitialized()
+        GliaWidgets.getEngagementLauncher(getQueueIdsFromPrefs(sharedPreferences))
+    }
+
+    private val entryWidget: EntryWidget by lazy {
+        ensureInitialized()
```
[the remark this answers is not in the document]
DavDo: Good point. I thought this was necessary for ATs, but we’re initializing the SDK immediately when we receive the deep-link intent.
So, most probably, this check is safe to completely remove.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
DavDo: I removed all the properties that are `useless` after these PR changes because they’re mainly used for intent extras that we closed from integrators, so there aren’t only the deprecated ones.
This is part of the breaking changes we will introduce in version 3.0.0. Do you think we need to somehow separate deprecated ones?

app/src/main/java/com/glia/exampleapp/MainFragment.kt
```diff
@@ -523,7 +485,7 @@ class MainFragment : Fragment() {
         GliaWidgets.init(
             createDefaultConfig(
                 context = requireActivity().applicationContext,
-//                uiJsonRemoteConfig = UnifiedUiConfigurationLoader.fetchRemoteConfiguration(),
+                uiJsonRemoteConfig = UnifiedUiConfigurationLoader.fetchLocalConfigSample(requireContext()),
```
andrews-moc, line 488: Maybe it makes sense to keep this code disabled?

app/src/main/java/com/glia/exampleapp/Utils.java
```diff
@@ -115,7 +109,6 @@ public static UiTheme getRunTimeThemeByPrefs(SharedPreferences sharedPreferences
 
         UiTheme.UiThemeBuilder builder = new UiTheme.UiThemeBuilder();
 
-        builder.setAppBarTitle(title);
```
andrews-moc, line 118: Just out of interest, what is the reason for removing `setAppBarTitle()`?

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -226,6 +134,7 @@ public static void onRequestPermissionsResult(int requestCode, String[] permissi
...
```
andrews-moc, line 137: The same as Alex asked for another TODO, is this going to be done in this PR or should we create a separate Jira ticket for this?

widgetssdk/src/main/java/com/glia/widgets/UiTheme.kt
```diff
@@ -1185,7 +1170,6 @@ data class UiTheme(
         }
 
         fun setTheme(theme: UiTheme) {
-            appBarTitle = theme.appBarTitle
```
andrews-moc, line 1188: For my understanding, why don't we need it anymore?

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/EndScreenSharingView.kt
```diff
@@ -109,10 +101,9 @@ internal class EndScreenSharingView(
         binding.root.applyLayerTheme(theme.background)
     }
 
-    private fun applyRuntimeTheme(theme: UiTheme?) {
-        if (theme == null) {
-            Logger.d(TAG, "UiTheme is null!")
-            return
+    private fun applyRuntimeTheme(theme: UiTheme) {
+        theme.iconEndScreenShare?.let {
+            binding.endSharingButton.icon = ResourcesCompat.getDrawable(context.resources, it, null)
```
andrews-moc, line 106: We were not applying style to the End Screen Sharing icon?

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/holder/OperatorStatusViewHolder.kt
```diff
@@ -85,21 +85,21 @@ internal class OperatorStatusViewHolder(
...
 
-    private fun applyInQueueState(companyName: String?) {
+    private fun applyInQueueState() {
         statusPictureView.showPlaceholder()
         applyChatStartingViewsVisibility()
         applyChatStartedViewsVisibility(false)
         itemView.setLocaleContentDescription(
             R.string.android_chat_queue_message_accessibility_label,
-            StringKeyPair(StringKey.COMPANY_NAME, companyName ?: "")
+            StringKeyPair(StringKey.COMPANY_NAME, "This should be ignored")
```
andrews-moc, line 102: It's safer to leave an empty String, in my opinion.

app/src/main/java/com/glia/exampleapp/Utils.java
DavDo: This property internally was not in use, so had no effect for a while, that is why I removed it.
Regarding release notes and migration guides, relying on the fact that we made activities internal and introduced EngagementLauncher instead should fully cover the migration part.

widgetssdk/src/main/java/com/glia/widgets/view/head/ChatHeadLayout.kt
```diff
@@ -159,7 +146,7 @@ internal class ChatHeadLayout @JvmOverloads constructor(
         visibility = GONE
         initConfigurations()
         setupViewActions()
-        readTypedArray(attrs, defStyleAttr, defStyleRes)
+//        readTypedArray(attrs, defStyleAttr, defStyleRes) TODO most probably we don't need this anymore
```
andrews-moc, line 149: Is it a leftover?

andrews-moc:
@DavDo You've done a good deal of work! 💪  Agree with @gugalo, it would be much easier to review and spot issues if a huge PR is split into smaller parts.

DavDo:
Thank you guys for your review.
I will try to address comments and split this PR into separate smaller parts, but since there are a lot of connections, I'm afraid that there will still be one that is comparatively big.

andrews-moc:
Taking into account the work done already, it makes sense to proceed with this PR, isn't it? Otherwise we would have to review everything once again. Not sure how to do better.

DavDo:
> Taking into account the work done already, it makes sense to proceed with this PR, isn't it? Otherwise we would have to review everything once again. Not sure how to do better.

I'd also prefer to proceed with this PR, but I'm also ready to split it if you really want it.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
DavDo: I will remove TODOs, I think it makes sense to have separate tickets for these two functions.
Do you think it makes sense to include this in this project?
This should be unused since we introduced `ActivityWatchers` for screen sharing and permissions, but unfortunately, we forgot to deprecate it.

widgetssdk/src/main/java/com/glia/widgets/UiTheme.kt
DavDo: I have no clue how it happened, but this property wasn’t in use in the project and, therefore, had no effect.

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/EndScreenSharingView.kt
DavDo: we were, but just before calling this function.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
andrews-moc: I think a separate ticket is better. It doesn't seem to be essential for our current project.

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
DavDo: I will mark these methods as deprecated since we're not using them anymore.

PR #1095, opened 2024-09-27
Devs want to implement the logic for showing available engagement types

andrews-moc:
[[Android] Devs want to implement the logic for showing available engagement types](https://glia.atlassian.net/browse/MOB-3411)

**What was solved?**
- Passing queues specified by integrator OR default queues to QueueMonitoring component
- Queue monitoring component subscribes to queue updates
- Each time queue is updated, the Queue monitoring component notifies the subscriber(s)
- Entry Widget Controller refreshes the available engagement types

**What has to be done?**
- Maybe extract standalone MonitorAvailableEngagementTypes component - separate PR
- Add unit tests - separate PR
- Add logs and align them with iOS (in "Devs want to add logging and telemetry" tickets)

**Release notes:**

 - [x] Feature
 - [x] Ignore

**Additional info:**

**Screen recording:**

https://github.com/user-attachments/assets/c7031e6f-564a-4d34-8251-a9bea118a3c7

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
@@ -91,7 +91,11 @@ public class GliaWidgets {
      */
     @NonNull
     public synchronized static EntryWidget getEntryWidget(@Nullable List<String> queueIds) {
-        setupQueueIds(queueIds);
+        if (Dependencies.glia().isInitialized()) {
+            setupQueueIds(queueIds);
+        } else {
+            Logger.w(TAG, "Attempt to get EntryWidget before SDK initialization");
+        }
 
         return Dependencies.getEntryWidget();
     }
--- a/widgetssdk/src/main/java/com/glia/widgets/core/queue/QueueMonitor.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/queue/QueueMonitor.kt
@@ -1,0 +1,98 @@
+package com.glia.widgets.core.queue
+
+import android.annotation.SuppressLint
+import com.glia.androidsdk.RequestCallback
+import com.glia.androidsdk.queuing.Queue
+import com.glia.widgets.core.queue.domain.FetchQueuesUseCase
+import com.glia.widgets.core.queue.domain.SubscribeToQueueUpdatesUseCase
+import com.glia.widgets.core.queue.domain.UnsubscribeFromQueueUpdatesUseCase
+import com.glia.widgets.helper.Data
+import com.glia.widgets.launcher.ConfigurationManager
+import io.reactivex.rxjava3.core.BackpressureStrategy
+import io.reactivex.rxjava3.core.Flowable
+import io.reactivex.rxjava3.processors.BehaviorProcessor
+import java.util.function.Consumer
+
+internal sealed interface QueueMonitorState {
+    object Loading : QueueMonitorState
+    data class Queues(val queues: List<Queue>) : QueueMonitorState
+    object Empty : QueueMonitorState
+    object Error : QueueMonitorState
+}
+
+internal class QueueMonitor(
+    configurationManager: ConfigurationManager,
+    private val fetchQueuesUseCase: FetchQueuesUseCase,
+    private val subscribeToQueueUpdatesUseCase: SubscribeToQueueUpdatesUseCase,
+    private val unsubscribeFromQueueUpdatesUseCase: UnsubscribeFromQueueUpdatesUseCase
+) {
+
+    // Queues received from an integrator or default queues
+    private val _integratorQueues: Flowable<List<Queue>> = configurationManager.queueIdsObservable
+        .flatMap(::findIntegratorQueues)
+
+    private val _observableIntegratorQueues: BehaviorProcessor<QueueMonitorState> = BehaviorProcessor.createDefault(QueueMonitorState.Loading)
+    val observableIntegratorQueues = _observableIntegratorQueues.hide()
+
+    private val queueUpdateCallback: Consumer<Queue> = Consumer {
+        updateQueuesList(it)
+    }
+
+    init {
+        subscribeToQueueUpdates()
+    }
+
+    private fun findIntegratorQueues(queueIds: Data<List<String>>): Flowable<List<Queue>> =
+        fetchSiteQueues()
+            .map { mapCurrentQueues(it, queueIds) }
+
+    private fun fetchSiteQueues(): Flowable<List<Queue>> = Flowable.create({
+        fetchQueuesUseCase(RequestCallback { queues, exception ->
+            if (exception == null) {
+                it.onNext(queues?.toList().orEmpty())
+            } else {
+                it.onError(exception)
+            }
+        })
+    }, BackpressureStrategy.LATEST)
+
+    @SuppressLint("CheckResult")
+    private fun subscribeToQueueUpdates() {
+        _integratorQueues.subscribe({
+            updateQueuesList(it)
+        }, {
+            _observableIntegratorQueues.onNext(QueueMonitorState.Error)
+        })
+    }
+
+    private fun updateQueuesList(queues: List<Queue>) {
+        unsubscribeFromQueueUpdatesUseCase(null, queueUpdateCallback)
+        if (queues.isEmpty()) {
+            _observableIntegratorQueues.onNext(QueueMonitorState.Empty)
+            return
+        }
+
+        _observableIntegratorQueues.onNext(QueueMonitorState.Queues(queues))
+        subscribeToQueueUpdatesUseCase(queues.map { it.id }.toTypedArray(), {}, queueUpdateCallback)
+    }
+
+    private fun updateQueuesList(queue: Queue) {
+        val updatedQueues = _observableIntegratorQueues.value
+            ?.let { it as? QueueMonitorState.Queues }
+            ?.run { queues.map { if (it.id == queue.id) queue else it } } ?: listOf(queue)
+
+        _observableIntegratorQueues.onNext(QueueMonitorState.Queues(updatedQueues))
+    }
+
+    private fun mapCurrentQueues(siteQueues: List<Queue>, queueIds: Data<List<String>>): List<Queue> {
+        return when (queueIds) {
+            is Data.Value -> {
+                siteQueues.filter { queueIds.result.contains(it.id) }
+            }
+
+            else -> {
+                siteQueues.filter { it.isDefault == true }
+            }
+        }
+    }
+}
--- a/widgetssdk/src/main/java/com/glia/widgets/core/queue/domain/FetchQueuesUseCase.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/queue/domain/FetchQueuesUseCase.kt
@@ -1,0 +1,15 @@
+package com.glia.widgets.core.queue.domain
+
+import com.glia.androidsdk.RequestCallback
+import com.glia.androidsdk.queuing.Queue
+import com.glia.widgets.di.GliaCore
+
+internal interface FetchQueuesUseCase {
+    operator fun invoke(callback: RequestCallback<Array<Queue>?>)
+}
+
+internal class FetchQueuesUseCaseImpl(private val gliaCore: GliaCore): FetchQueuesUseCase {
+    override fun invoke(callback: RequestCallback<Array<Queue>?>) {
+        gliaCore.getQueues(callback)
+    }
+}
--- a/widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
@@ -1,25 +1,66 @@
 package com.glia.widgets.entrywidget
 
-internal class EntryWidgetController : EntryWidgetContract.Controller {
+import android.annotation.SuppressLint
+import com.glia.androidsdk.Engagement
+import com.glia.androidsdk.queuing.Queue
+import com.glia.androidsdk.queuing.QueueState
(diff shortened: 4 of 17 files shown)


widgetssdk/src/main/java/com/glia/widgets/core/queue/QueueMonitor.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    private fun mapCurrentQueues(siteQueues: List<Queue>, queueIds: Data<List<String>>): List<Queue> {
+        return when (queueIds) {
+            is Data.Value -> {
+                siteQueues.filter { queue -> queueIds.result.contains(queue.id) }
+            }
+
+            else -> {
+                siteQueues.filter { it.isDefault == true }
+            }
```
gugalo, line 79: I think it would be nice if you use in both `when` cases consistent approach:

```suggestion
            is Data.Value -> {
                siteQueues.filter { queueIds.result.contains(it.id) }
            }

            else -> {
                siteQueues.filter { it.isDefault == true }
            }
```

OR

```suggestion
            is Data.Value -> {
                siteQueues.filter { queue -> queueIds.result.contains(queue.id) }
            }

            else -> {
                siteQueues.filter { queue -> queue.isDefault == true }
            }
```

widgetssdk/src/main/java/com/glia/widgets/core/queue/QueueMonitor.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    init {
+        fetchSiteQueues()
+    }
+
+    private fun fetchSiteQueues() {
+        val callback: (Array<Queue>?, GliaException?) -> Unit = { queues, error ->
+            if (error == null && !queues.isNullOrEmpty()) {
+                _siteQueues.onNext(queues.toList())
+                subscribeToQueueUpdates()
```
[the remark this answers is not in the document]
andrews-moc: Fixed 81d993003a86b1d917e84b5cdbd6f87798a236ac

widgetssdk/src/main/java/com/glia/widgets/core/queue/QueueMonitor.kt
andrews-moc: Good catch, thank you.

andrews-moc: Fixed fe69ca30d1fd76534e309d3d92e65eea8a270bd0

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -184,7 +184,11 @@ public synchronized static EngagementLauncher getEngagementLauncher(@Nullable Li
      */
     @NonNull
     public synchronized static EntryWidget getEntryWidget(@Nullable List<String> queueIds) {
-        setupQueueIds(queueIds);
+        if (Dependencies.glia().isInitialized()) {
```
DavDo, line 187: What if integrator keeps an `EntryWidget` instance and reuses after initialization?
We will have default queues, instead of provided queues, I think it will be correct to somehow save the queueIds.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -184,7 +184,11 @@ public synchronized static EngagementLauncher getEngagementLauncher(@Nullable Li
...
+            setupQueueIds(queueIds);
+        } else {
+            Logger.w(TAG, "Attempt to get EntryWidget before SDK initialization");
```
DavDo, line 190: But actually what is the purpose of letting integrators to get the entrywidget without SDK initialization?

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetView.kt
```diff
@@ -55,7 +57,9 @@ internal class EntryWidgetView(
     }
 
     override fun showItems(items: List<EntryWidgetContract.ItemType>) {
-        viewAdapter.items = items
+        Handler(Looper.getMainLooper()).post {
```
DavDo, line 60: What is the purpose of upgrading items property on main thread?

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
```diff
@@ -1,20 +1,40 @@
...
+            EntryWidgetContract.ItemType.EMPTY_STATE
         )
         view.showItems(items)
+
+        queueMonitor.observableIntegratorQueues.subscribe {
+            // TODO: code below should be moved to a standalone component
+            val openQueues = it.filter { queueFromList -> queueFromList.state.status == QueueState.Status.OPEN }
+            val allMedias = openQueues.flatMap { queueFromList -> queueFromList.state.medias.toList() }.distinct()
+
+            items.clear()
```
DavDo, line 24: I'd prefer to have immutable list here.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetView.kt
andrews-moc: Otherwise, UI is not updated dynamically.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
andrews-moc: Changed: f91942e868436ea28597504e3817387ba1e0bb5c. Please have a look to see if it's what you suggested.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
andrews-moc: As per requirements:
> Integrators can launch the Entry Widget view after the SDK setup. If SDK hasn’t been initialized or was initialized with an error - Widgets SDK should not crash, it should show an [Entry Widget with an error state](https://www.figma.com/design/x5BLxm2UWpm3TuerdqmYi8/Glia-Mobile-App-Pages?node-id=11753-4938&node-type=SECTION&t=csKDTOMlup7L17Hn-0).

Initially, I suggested throwing an exception and crashing the app so that this misconfiguration issue was discovered early during development time.

I remember @gugalo brought up the idea to avoid crashing at the kick-off meeting and it has been approved.

@DavDo What alternatives do you have in mind?

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
DavDo: sorry for the poorly described request.
I meant not to keep the mutable list of items and clear it each time and then add new items.

I meant smth like this
```kotlin
override fun setView(view: EntryWidgetContract.View) {
        this.view = view
        view.showItems(listOf(EntryWidgetContract.ItemType.EMPTY_STATE))

        queueMonitor.observableIntegratorQueues.subscribe {
            // TODO: code below should be moved to a standalone component
            val openQueues = it.filter { queueFromList -> queueFromList.state.status == QueueState.Status.OPEN }
            val allMedias = openQueues.flatMap { queueFromList -> queueFromList.state.medias.toList() }.distinct()

            val items = buildList {
                if (allMedias.contains(Engagement.MediaType.TEXT)) add(EntryWidgetContract.ItemType.CHAT)
                if (allMedias.contains(Engagement.MediaType.AUDIO)) add(EntryWidgetContract.ItemType.AUDIO_CALL)
                if (allMedias.contains(Engagement.MediaType.VIDEO)) add(EntryWidgetContract.ItemType.VIDEO_CALL)
                if (allMedias.contains(Engagement.MediaType.MESSAGING)) add(EntryWidgetContract.ItemType.SECURE_MESSAGE)

                if (allMedias.isEmpty()) {
                    add(EntryWidgetContract.ItemType.EMPTY_STATE)
                } else {
                    add(EntryWidgetContract.ItemType.PROVIDED_BY)
                }
            }
            view.showItems(items)
        }
    }
``` 

Or maybe one more way is to use Rx advantages like this
```kotlin
    override fun setView(view: EntryWidgetContract.View) {
        this.view = view
        view.showItems(listOf(EntryWidgetContract.ItemType.EMPTY_STATE))

        // TODO: code below should be moved to a standalone component
        queueMonitor.observableIntegratorQueues
            .map { queues -> queues.map { it.state } }
            .map { states -> states.filter { it.status == QueueState.Status.OPEN } }
            .map { openQueues -> openQueues.flatMap { it.medias.toList() }.distinct() }
            .map { allMedias ->
                buildList {
                    if (allMedias.contains(Engagement.MediaType.TEXT)) add(EntryWidgetContract.ItemType.CHAT)
                    if (allMedias.contains(Engagement.MediaType.AUDIO)) add(EntryWidgetContract.ItemType.AUDIO_CALL)
                    if (allMedias.contains(Engagement.MediaType.VIDEO)) add(EntryWidgetContract.ItemType.VIDEO_CALL)
                    if (allMedias.contains(Engagement.MediaType.MESSAGING)) add(EntryWidgetContract.ItemType.SECURE_MESSAGE)

                    if (allMedias.isEmpty()) {
                        add(EntryWidgetContract.ItemType.EMPTY_STATE)
                    } else {
                        add(EntryWidgetContract.ItemType.PROVIDED_BY)
                    }
                }
            }
            //This way we can avoid switching to main thread in Adapter
            .observeOn(AndroidSchedulers.mainThread())
            .subscribe(view::showItems)
    }
```

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
DavDo: > Initially, I suggested throwing an exception and crashing the app so that this misconfiguration issue was discovered early during development time.

I'm still up for this suggestion because this initialization part and errors thrown during misconfiguration must be disruptive to show that misconfiguration during development.
I think if it goes to the production, that is the integrator's problem.

However, if we decide to go with this solution, I think the only way is to still configure queues and log errors.
```java
public synchronized static EntryWidget getEntryWidget(@Nullable List<String> queueIds) {
        Dependencies.getConfigurationManager().setQueueIds(queueIds);

        if (!Dependencies.glia().isInitialized()) {
            Logger.w(TAG, "Attempt to get EntryWidget before SDK initialization");
        }

        return Dependencies.getEntryWidget();
    }
```

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
andrews-moc: Ah, I understand now. Yes, the Rx approach is better because it also avoids switching to main thread in adapter. 👍 
Made changes here 5b65432c8abc8bf2b5d37939cca3fdbede9bbfa6. Thank you.

widgetssdk/src/main/java/com/glia/widgets/core/queue/domain/FetchQueuesUseCase.kt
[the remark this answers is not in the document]
andrews-moc: The main reason to create a UseCase is a need for `GliaCore`.

By moving `gliaCore.getQueues(callback)` into a UseCase, I can easily pass `gliaCore` to `FetchQueuesUseCase` in `UseCaseFactory`. If I don’t separate `gliaCore.getQueues(callback)` into a UseCase, then `gliaCore` is needed in `QueueMonitor`. But `QueueMonitor` is created in `ControllerFactory`, which doesn’t have `gliaCore`.
This was an indicator for me that UseCase is needed although, the purpose of this Use Case is just to call `gliaCore.getQueues(callback)`.

@DavDo @gugalo Please correct me if I'm wrong.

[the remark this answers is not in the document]
DavDo: Good point.

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/core/queue/domain/FetchQueuesUseCase.kt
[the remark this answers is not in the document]
gugalo: So if we could improve our dependency injection solution, could we eliminate those use cases? Or is there something else that is related to this?

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
gugalo: Could we upgrade the log to error?
I think we want to get alerted if integrators are attempting to use entry widgets before SDK is init

widgetssdk/src/main/java/com/glia/widgets/core/queue/domain/FetchQueuesUseCase.kt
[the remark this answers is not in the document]
DavDo: In my opinion, yes. 
But we also need to define in which cases we will use usecases and refactor the codebase accordingly.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
andrews-moc: GitHub shows the diff a little unclear, I will explain what has been changed:
1. Setup queues no matter if SDK is initialized
2. Change log level for this case to 'error'
3. Moved the setupQueueIds function closer to the code which calls it.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
andrews-moc: I've made changes as per the code that you suggested.
Also, raised this question in Slack.

DavDo (approved):

gugalo (approved):

PR #1097, opened 2024-10-02
Upgrade Dokka to latest version

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3617

**What was solved?**
In new version of dokka changes of note:
- Support for @author, @since and @return tags in javadocs
- Significantly improve the mobile layout and overall responsiveness There are other changes also. For details see:
https://github.com/Kotlin/dokka/releases

**Release notes:**

 - [ ] Feature
 - [ ] Ignore
 - [x] Release notes: API docs engene update

**Additional info:**

**Screenshots:**
 < - Left side is the old version ===================== Right side is the new version -> 
image

In most cases however nothing changed and things are still rendered incorrectly
image

P.S: Most things however haven't changed and it seems that @Depricated tag is still rendered incorrectly or not rendered at all
image

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,7 +2,7 @@
 javaVersion = "17"
 
 kotlinVersion = "1.8.22" # Kotlin, Kotlin Reflect and Dokka versions should have the same minor version to support the same kotlin version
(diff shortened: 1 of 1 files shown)


DavDo (approved):

DavDo:
🤔 Maybe it's the right time to switch to `Kotlin` docs, which are much cleaner and beautiful.

andrews-moc (approved):

PR #1118, opened 2024-10-16
Fix! Messages are not added to the chat immediately after input

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3664

**What was solved?**
### Before

https://github.com/user-attachments/assets/cc791fef-0324-4b41-a68c-58e8071e9a69
### After

https://github.com/user-attachments/assets/b3da1f36-2e9a-4303-93b5-231094efa8c4

**Release notes:**
Visitor messages will appear immediately after the send button is clicked.

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
@@ -52,8 +52,8 @@ import com.glia.widgets.core.notification.domain.CallNotificationUseCase
 import com.glia.widgets.core.permissions.domain.RequestNotificationPermissionIfPushNotificationsSetUpUseCase
 import com.glia.widgets.core.permissions.domain.WithCameraPermissionUseCase
 import com.glia.widgets.core.permissions.domain.WithReadWritePermissionsUseCase
-import com.glia.widgets.core.secureconversations.domain.IsSecureEngagementUseCase
 import com.glia.widgets.core.secureconversations.domain.GetAvailableQueueIdsForSecureMessagingUseCase
+import com.glia.widgets.core.secureconversations.domain.IsSecureEngagementUseCase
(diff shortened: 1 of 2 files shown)


AndriiHorishniiMOC (approved):
LGTM!

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -153,8 +153,12 @@ internal class ChatController(
             emitViewState { chatState.setLastTypedText("").setShowSendButton(isShowSendButtonUseCase("")) }
         }
 
-        override fun errorOperatorNotOnline(message: Unsent) {
-            onSendMessageOperatorOffline(message)
+        override fun onMessagePrepared(message: Unsent) {
+            appendUnsentMessage(message)
+        }
+
+        override fun errorOperatorNotOnline() {
```
andrews-moc, line 160: Just as an idea, maybe it will become a bit more readable if we rename `errorOperatorNotOnline()` to `errorOperatorOffline()`.

andrews-moc (approved):
To me, it sometimes (like here) makes sense to record a video of how it looks like with the changes. It helps with review, proves the changes and could help in future investigations.

DavDo:
> To me, it sometimes (like here) makes sense to record a video of how it looks like with the changes. It helps with review, proves the changes and could help in future investigations.

Good Idea, I will add.

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
DavDo: changed

andrews-moc:
@DavDo Thank you for adding the videos. 👍

PR #1120, opened 2024-10-16
Refactor queue management to avoid race conditions

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3698

**What was solved?**
In my previous change, I didn't consider that, in some circumstances, we're requesting relevant queues before we receive the site queues from the server. Therefore, I made everything reactive, improved readability and added an initialization check in EntryWidgetController to show an `Error` state immediately if the SDK is not initialized.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/app/src/main/res/values/themes.xml
+++ b/app/src/main/res/values/themes.xml
@@ -11,7 +11,7 @@
     <!-- Must be included for dialog styling -->
     <style name="Application.Glia.Activity.Style" parent="Theme.MaterialComponents.DayNight.NoActionBar">
         <!-- example usage of compile time styling -->
-        <item name="gliaChatStyle">@style/Theme.GliaAndroidSdkWidgetsExample.Chat</item>
+        <!--        <item name="gliaChatStyle">@style/Theme.GliaAndroidSdkWidgetsExample.Chat</item>-->
     </style>
 
     <style name="Theme.GliaAndroidSdkWidgetsExample.Chat" parent="Application.Glia.Chat">
--- a/widgetssdk/src/main/java/com/glia/widgets/core/queue/QueueRepository.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/queue/QueueRepository.kt
@@ -1,5 +1,6 @@
 package com.glia.widgets.core.queue
 
+import com.glia.androidsdk.GliaException
 import com.glia.androidsdk.queuing.Queue
 import com.glia.widgets.di.GliaCore
 import com.glia.widgets.helper.Logger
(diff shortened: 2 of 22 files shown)


andrews-moc (commented):
> in some circumstances, we're requesting relevant queues before we receive the site queues from the server
Could you please describe these cases? Maybe we will need to include them to the test cases for exploratory testing.

DavDo:
> > in some circumstances, we're requesting relevant queues before we receive the site queues from the server
> 
> Could you please describe these cases? Maybe we will need to include them to the test cases for exploratory testing.

Sure! I mistakenly thought it was impossible to request an engagement right after initialization. So we can get IDs directly. But, when you tap the chat button without initialization, it will immediately initialize SDK and open the chat screen, which will start enqueueing with a (still) empty list.
I also somehow skipped the case when the Entry widget was requested without initializing the SDK, and I fixed this case as well.

app/src/main/res/values/themes.xml
```diff
@@ -11,7 +11,7 @@
     <!-- Must be included for dialog styling -->
     <style name="Application.Glia.Activity.Style" parent="Theme.MaterialComponents.DayNight.NoActionBar">
         <!-- example usage of compile time styling -->
-        <item name="gliaChatStyle">@style/Theme.GliaAndroidSdkWidgetsExample.Chat</item>
+        <!--        <item name="gliaChatStyle">@style/Theme.GliaAndroidSdkWidgetsExample.Chat</item>-->
```
andrews-moc, line 14: 👍

widgetssdk/src/main/java/com/glia/widgets/core/queue/QueueRepository.kt
```diff
@@ -32,57 +40,65 @@ internal class QueueRepositoryImpl(private val gliaCore: GliaCore, private val c
...
+    override val relevantQueueIds: Single<List<String>>
+        get() = _relevantQueueIds.firstOrError()
+
+    override fun initialize() {
+        fetchQueues()
     }
 
     private fun fetchQueues() {
-        if (siteQueues.value?.isSuccess != true) {
+        if (siteQueues.value?.isSuccess != true && gliaCore.isInitialized) {
```
andrews-moc, line 57: I think it's logical to change the order to `(gliaCore.isInitialized && siteQueues.value?.isSuccess != true)`.

widgetssdk/src/main/java/com/glia/widgets/core/queue/QueueRepository.kt
```diff
@@ -32,57 +40,65 @@ internal class QueueRepositoryImpl(private val gliaCore: GliaCore, private val c
...
+    private fun siteQueuesReceived(queues: List<Queue>) {
+        siteQueues.onNext(Result.success(queues))
+        subscribeToQueues()
+        subscribeToQueueUpdates()
+    }
+
+    private fun getSiteQueuesError(exception: GliaException?) {
+        siteQueues.onNext(Result.failure(exception ?: RuntimeException("Fetching queues failed")))
+        Logger.e(TAG, "Fetching queues failed", exception)
+    }
```
andrews-moc, line 75: NIT. From what I see, the names like `reportSiteQueuesError`, `onFetchingQueuesError()` might reflect the function responsibility better.

widgetssdk/src/main/java/com/glia/widgets/core/queue/QueueRepository.kt
```diff
@@ -32,57 +40,65 @@ internal class QueueRepositoryImpl(private val gliaCore: GliaCore, private val c
...
+
+    private fun getIds(queues: List<Queue>): List<String> = queues.map { it.id }
```
andrews-moc, line 77: NIT. It's a tiny function, I think `queues.map { it.id }` can just be put without having a function.

andrews-moc:
> > > in some circumstances, we're requesting relevant queues before we receive the site queues from the server
> > 
> > 
> > Could you please describe these cases? Maybe we will need to include them to the test cases for exploratory testing.
> 
> Sure! I mistakenly thought it was impossible to request an engagement right after initialization. So we can get IDs directly. But, when you tap the chat button without initialization, it will immediately initialize SDK and open the chat screen, which will start enqueueing with a (still) empty list. I also somehow skipped the case when the Entry widget was requested without initializing the SDK, and I fixed this case as well.

Got it, thank you.

> I also somehow skipped the case when the Entry widget was requested without initializing the SDK, and I fixed this case as well.

This case has been handled correctly, I think. Entry Widget with an error state was shown before SDK is initialized. I've just re-tested. Or you mean something else?

DavDo:
> > Sure! I mistakenly thought it was impossible to request an engagement right after initialization. So we can get IDs directly. But, when you tap the chat button without initialization, it will immediately initialize SDK and open the chat screen, which will start enqueueing with a (still) empty list. I also somehow skipped the case when the Entry widget was requested without initializing the SDK, and I fixed this case as well.
> 
> Got it, thank you.
> 
> > I also somehow skipped the case when the Entry widget was requested without initializing the SDK, and I fixed this case as well.
> 
> This case has been handled correctly, I think. Entry Widget with an error state was shown before SDK is initialized. I've just re-tested. Or you mean something else?

Correct! That was because we were trying to fetch queues and were getting the `SDK is not initialized` error, but since I added a check for initialization before fetching queues, this logic got broken.
It was based on the `fetchQueues` error instead of handling initialization itself.

andrews-moc (approved):

AndriiHorishniiMOC (approved):

DavDo:
!squash

PR #1123, opened 2024-10-18
Upgrade all dependencies.

DavDo:
**What was solved?**
I have upgraded all dependencies, and regenerated snapshots to align them with a new version of the paparazzi.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,8 @@ allprojects {
 
     // Needed for some automated and manual testing (e.g: acceptance tests)
     mavenLocal()
+    //TODO switch to the release version before releasing core SDK
+    maven { url = "https://s01.oss.sonatype.org/content/repositories/comglia-1316" }
   }
 }
 
@@ -66,6 +68,26 @@ subprojects {
   // For some reason artifact ID mentioned in docs does not exist, probably a typo. This one has slightly different order of words
   dependencies {
     dokkaPlugin(libs.java.dokka)
+  }
+}
+// This is known bug in paparazzi 1.3.3 version.
+subprojects {
+  plugins.withId("app.cash.paparazzi") {
+    // Defer until afterEvaluate so that testImplementation is created by Android plugin.
+    afterEvaluate {
+      dependencies.constraints {
+        add("testImplementation", "com.google.guava:guava") {
+          attributes {
+            attribute(
+              TargetJvmEnvironment.TARGET_JVM_ENVIRONMENT_ATTRIBUTE,
+              objects.named(TargetJvmEnvironment, TargetJvmEnvironment.STANDARD_JVM)
+            )
+          }
+          because("LayoutLib and sdk-common depend on Guava's -jre published variant." +
+            "See https://github.com/cashapp/paparazzi/issues/906.")
+        }
+      }
+    }
   }
 }
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -3,42 +3,43 @@ javaVersion = "17"
 
 kotlinVersion = "1.9.25" # Kotlin and Dokka versions should have the same minor version to support the same kotlin version
 dokkaVersion = "1.9.20" # Kotlin and Dokka versions should have the same minor version to support the same kotlin version
-agpVersion = "8.6.0"
+agpVersion = "8.6.1"
 nexusPublishVersion = "2.0.0"
 ktlintVersion = "11.5.0"
 
 # be carefully with updating this plugin, it may upgrade kotlin version for the entire project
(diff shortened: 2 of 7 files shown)


gradle/libs.versions.toml
```diff
@@ -3,42 +3,43 @@ javaVersion = "17"
...
-mediaLottieVersion = "5.2.0"
+javaRxAndroid3Version = "3.0.2"
+javaRxJava3Version = "3.1.9"
+javaNavVersion = "2.8.3"
+dataGsonVersion = "2.11.0"
+firebaseBomVersion = "33.4.0"
+mediaAudioswitch = "1.2.0"
+mediaPicassoVersion = "2.71828"
+mediaLottieVersion = "6.5.2"
+mediaExifInterfaceVersion = "1.3.7"
```
andrews-moc, line 33: I expected to see only updated versions. However, `mediaExifInterfaceVersion` is a new one, it seems. Is this correct?

andrews-moc (commented):
Looks like these two screenshots show the real differences: 
Screenshot 2024-10-21 at 15 10 53
Screenshot 2024-10-21 at 15 12 27

DavDo:
@AndriiHorishniiMOC thanks for your research and fix! 👍

AndriiHorishniiMOC:
The issue was in Paparazzi 1.3.4. Unfortunately, the next version has not yet been released. There is 1.3.5-SNAPSHOT, so we should decide whether to update now or wait until a stable release.

DavDo:
> The issue was in Paparazzi `1.3.4`. Unfortunately, the next version has not yet been released. There is `1.3.5-SNAPSHOT`, so we should decide whether to update now or wait until a stable release.

Since it will stay in the feature branch and there are no visible problems with this snapshot build, I think we can go with the snapshot version.

DavDo:
!squash

gradle/libs.versions.toml
DavDo: Yes, because previously, the `android.core-ktx` library transitively contained this dependency, but in newer versions, they dropped it.

DavDo:
> The issue was in Paparazzi `1.3.4`. Unfortunately, the next version has not yet been released. There is `1.3.5-SNAPSHOT`, so we should decide whether to update now or wait until a stable release.

Seems we will be unable to use this version. Sadly, it forces the Kotlin 2.0.20 version, and there is no way to change it back to 1.9.25.

DavDo:
Paparazzi version 1.3.3 works well, so I'll leave this version.

AndriiHorishniiMOC:
@DavDo, can we apply this plugin only to the snapshot build type?

DavDo:
!squash

DavDo:
> @DavDo, can we apply this plugin only to the snapshot build type?

Interesting Idea, let's try it.

DavDo:
!squash

DavDo:
> @DavDo, can we apply this plugin only to the snapshot build type?

This also breaks the Kotlin version of the project, so we will go with 1.3.3

build.gradle
```diff
@@ -25,6 +25,7 @@ allprojects {
 
     // Needed for some automated and manual testing (e.g: acceptance tests)
     mavenLocal()
+    maven { url = "https://s01.oss.sonatype.org/content/repositories/comglia-1315" }
```
AndriiHorishniiMOC, line 28: Why do we need it here? Is it a template?

AndriiHorishniiMOC (approved):

build.gradle
DavDo: To use core sdk from the staging repository, which includes the latest SC V2 changes.

andrews-moc: Maybe it makes sense to add a TODO here to remove this line before release.

andrews-moc (approved):

widgetssdk/build.gradle
```diff
@@ -80,6 +80,11 @@ tasks.register('publishToLocalMaven') {
   dependsOn ':widgetssdk:publishSnapshotPublicationToMavenLocal'
 }
 
+tasks.withType(Test).configureEach {
+  // Increase the maximum heap size to 2 GB
```
gugalo, line 84: Please comment 'why' this configuration is needed.

gugalo (approved):

widgetssdk/build.gradle
DavDo: This was needed with Paparazzi 1.3.4 to not fail while running on the local machine; I will recheck for the 1.3.3 version.

PR #1124, opened 2024-10-18
Implement Entry Widget item click navigation

andrews-moc:
**What was solved?**
- [Implement Entry Widget item click navigation](https://glia.atlassian.net/browse/MOB-3702)
- [Make changes to the getView() interface based on the discussions](https://glia.atlassian.net/browse/MOB-3694)

**Release notes:**

 - [x] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

https://github.com/user-attachments/assets/005915c1-ab30-414b-b4fc-200b65c15335

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
@@ -21,9 +21,9 @@ interface EntryWidget {
     /**
      * Creates the entry widget view that might be easily embedded in a custom UI.
      *
-     * @return The entry widget view.
+     * @return New instance of an entry widget view.
      */
-    fun createEntryWidgetView(context: Context): View
+    fun getView(context: Context): View
 
     /**
      * Hides the entry widget.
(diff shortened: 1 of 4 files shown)


widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
```diff
@@ -23,7 +23,7 @@ interface EntryWidget {
      *
      * @return The entry widget view.
      */
-    fun createEntryWidgetView(context: Context): View
+    fun getView(context: Context): View
```
gugalo, line 26: I think with the new method name we should clarify that every time this method is called it creates a new instance of the view. This is a public method, right?

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
andrews-moc: @gugalo Could you please have a look if it's clear now cbaca298b6004c4efa9fc80e4988360a7cf65cc5 ?

gugalo (approved):

PR #1131, opened 2024-10-25
Handle edge cases

andrews-moc:
**What was solved?**
- [Entry Widget should show an empty state if only 'messaging' engagement is available but visitor is not authenticated](https://glia.atlassian.net/browse/MOB-3712)
- [Hide the 'Try Again' button if SDK is not initialized](https://glia.atlassian.net/browse/MOB-3728)
- Added snapshot tests for the case when Glia is not initialized and we should hide the 'Try again' button

**Release notes:**

 - [x] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

https://github.com/user-attachments/assets/1b9e9b02-a0c6-49ac-897c-b09f91a0b73a

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
@@ -40,7 +40,10 @@ internal class EntryWidgetImpl(
     override fun show(activity: Activity) = activityLauncher.launchEntryWidget(activity)
 
     override fun getView(context: Context): View {
-        val adapter = EntryWidgetAdapter(EntryWidgetContract.ViewType.EMBEDDED_VIEW, themeManager.theme?.entryWidgetTheme)
+        val adapter = EntryWidgetAdapter(
+            EntryWidgetContract.ViewType.EMBEDDED_VIEW,
+            themeManager.theme?.entryWidgetTheme
+        )
         return EntryWidgetView(context, adapter)
     }
 
--- a/widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
@@ -29,12 +29,12 @@ internal class EntryWidgetController(
         if (core.isInitialized) {
             subscribeToQueueState()
         } else {
-            initErrorState()
+            showSdkNotInitializedState()
         }
     }
 
-    private fun initErrorState() {
-        view.showItems(listOf(EntryWidgetContract.ItemType.ERROR_STATE))
+    private fun showSdkNotInitializedState() {
+        view.showItems(listOf(EntryWidgetContract.ItemType.SDK_NOT_INITIALIZED_STATE))
     }
 
     private fun subscribeToQueueState() {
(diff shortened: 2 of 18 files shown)


widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
```diff
@@ -40,7 +41,10 @@ internal class EntryWidgetImpl(
     override fun show(activity: Activity) = activityLauncher.launchEntryWidget(activity)
 
     override fun getView(context: Context): View {
-        val adapter = EntryWidgetAdapter(EntryWidgetContract.ViewType.EMBEDDED_VIEW, themeManager.theme?.entryWidgetTheme)
+        val adapter = EntryWidgetAdapter(
+            Glia.isInitialized(),
```
AndriiHorishniiMOC, line 45: I don't like passing this state directly to the Adapter. What if the Controller handles this?

DavDo: Defining a new `ViewType.UNAUTHENTICATED_ITEM` type may resolve this issue so we can move all the business logic to the controller.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
```diff
@@ -72,7 +72,7 @@ internal class EntryWidgetController(
                     if (allMedias.contains(Engagement.MediaType.VIDEO)) add(EntryWidgetContract.ItemType.VIDEO_CALL)
                     if (allMedias.contains(Engagement.MediaType.MESSAGING) && isAuthenticatedUseCase()) add(EntryWidgetContract.ItemType.SECURE_MESSAGE)
 
-                    if (allMedias.isEmpty()) {
+                    if (this.isEmpty()) {
```
DavDo, line 75: This solution looks great, but I think it might be more reliable to check the existing media items and authentication in combination because adding some extra items before this step that might not be related to the media types may break this logic.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
AndriiHorishniiMOC: @DavDo Yes, we can handle everything [here](https://github.com/salemove/android-sdk-widgets/blob/feature/entry-widget-and-secure-conversations-v2/widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt#L32).

DavDo: Great!

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
andrews-moc: Okay. This concern should be covered by this fixup: 6426daffd45c8e4f647b2c9c0b14c0f2406bdf14.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
andrews-moc: @DavDo @AndriiHorishniiMOC Okay. Thank you for the feedback. I introduced a new `ItemType` called `SDK_NOT_INITIALIZED_STATE`. Please have a look now.

DavDo (approved):
Looks Good! 👍

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
```diff
@@ -40,7 +40,9 @@ internal class EntryWidgetImpl(
     override fun show(activity: Activity) = activityLauncher.launchEntryWidget(activity)
 
     override fun getView(context: Context): View {
-        val adapter = EntryWidgetAdapter(EntryWidgetContract.ViewType.EMBEDDED_VIEW, themeManager.theme?.entryWidgetTheme)
+        val adapter = EntryWidgetAdapter(
+            EntryWidgetContract.ViewType.EMBEDDED_VIEW,
+            themeManager.theme?.entryWidgetTheme)
```
AndriiHorishniiMOC, line 45:
```suggestion
        val adapter = EntryWidgetAdapter(
            EntryWidgetContract.ViewType.EMBEDDED_VIEW,
            themeManager.theme?.entryWidgetTheme
        )
```
HINT: I like that style.

AndriiHorishniiMOC (approved):
LGTM!

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
andrews-moc: Thank you. Changed. ✅

PR #1139, opened 2024-10-31
Add SC business logic into Entry Widget public functions

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3650

**What was solved?**
Add secure conversation business logic into the entry widget public functions

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/SecureConversationsRepository.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/SecureConversationsRepository.kt
@@ -8,10 +8,14 @@ import com.glia.androidsdk.secureconversations.SecureConversations
 import com.glia.widgets.chat.data.GliaChatRepository
 import com.glia.widgets.chat.domain.GliaSendMessageUseCase
 import com.glia.widgets.core.queue.QueueRepository
+import com.glia.widgets.entrywidget.EntryWidget
 import com.glia.widgets.helper.unSafeSubscribe
+import com.glia.widgets.launcher.EngagementLauncher
 import io.reactivex.rxjava3.core.Observable
+import io.reactivex.rxjava3.core.Single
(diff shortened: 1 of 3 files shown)


widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/SecureConversationsRepository.kt
```diff
@@ -65,6 +69,21 @@ internal class SecureConversationsRepository(private val secureConversations: Se
...
+            if (!emitter.isDisposed) {
+                emitter.onSuccess(hasPendingSecureConversations ?: false)
+            }
+        }
+    }
+
+    /**
+     * Since we're using this function in [EntryWidget] and [EngagementLauncher] where we need the immediate UI response to the user interaction,
+     * it is important to not make the user wait.
+     */
```
andrews-moc, line 85: Let me clarify6 for my understanding. So, when visitor taps 'Contact us' which should show the bottom sheet Entry Widget, if there is ongoing SC, we show the messaging screen after a programmatic 1-sec delay, right? Why is it better than showing it right immediately?

DavDo: No, it will reply with a default value(false) **if the request takes more than 1 second**.

We will have this value right after initialization, so this scenario is a corner case when the internet connection is slow, and EntryWidget is requested right after initialization.

andrews-moc: Possible improvement: "to not make the user wait" => “to ensure a seamless experience for the user” or “to avoid delays for the user”.
Up to you. 😉

AndriiHorishniiMOC (approved):
LGTM!

andrews-moc (approved):

PR #1141, opened 2024-11-01
Add ability to pass visitor context id through Engagement Launcher

andrews-moc:
[MOB-3752](https://glia.atlassian.net/browse/MOB-3752)

**What was solved?**
- Add ability to pass visitor context id through Engagement Launcher

**Release notes:**

 - [x] Feature
 - [x] Ignore

**Additional info:**

[MOB-3752]: https://glia.atlassian.net/browse/MOB-3752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/di/RepositoryFactory.java
+++ b/widgetssdk/src/main/java/com/glia/widgets/di/RepositoryFactory.java
@@ -149,7 +149,12 @@ public class RepositoryFactory {
 
     public EngagementRepository getEngagementRepository() {
         if (engagementRepository == null) {
-            engagementRepository = new EngagementRepositoryImpl(gliaCore, getOperatorRepository(), getQueueRepository());
+            engagementRepository = new EngagementRepositoryImpl(
+                gliaCore,
+                getOperatorRepository(),
+                getQueueRepository(),
+                configurationManager
+            );
         }
         return engagementRepository;
     }
--- a/widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
@@ -37,6 +37,7 @@ import com.glia.widgets.helper.Logger
 import com.glia.widgets.helper.isAudioOrVideo
 import com.glia.widgets.helper.isQueueUnavailable
 import com.glia.widgets.helper.unSafeSubscribe
+import com.glia.widgets.launcher.ConfigurationManager
 import io.reactivex.rxjava3.core.Completable
 import io.reactivex.rxjava3.core.Flowable
 import io.reactivex.rxjava3.disposables.CompositeDisposable
@@ -55,7 +56,8 @@ internal const val SKIP_ASKING_SCREEN_SHARING_PERMISSION_RESULT_CODE = 0x1995
 internal class EngagementRepositoryImpl(
     private val core: GliaCore,
     private val operatorRepository: GliaOperatorRepository,
-    private val queueRepository: QueueRepository
+    private val queueRepository: QueueRepository,
+    private val configurationManager: ConfigurationManager
 ) : EngagementRepository {
     private val engagementRequestCallback = Consumer<IncomingEngagementRequest>(::handleIncomingEngagementRequest)
     private val engagementOutcomeCallback = Consumer<Outcome>(::handleEngagementOutcome)
@@ -213,7 +215,12 @@ internal class EngagementRepositoryImpl(
         queueIngDisposable.add(
             queueRepository.relevantQueueIds.subscribe { ids ->
                 if (ids.isNotEmpty()) {
-                    core.queueForEngagement(ids, mediaType, null, null, MEDIA_PERMISSION_REQUEST_CODE) {
+                    core.queueForEngagement(
+                        ids,
+                        mediaType,
+                        configurationManager.visitorContextAssetId,
+                        null,
+                        MEDIA_PERMISSION_REQUEST_CODE) {
                         handleQueueingResponse(it)
                     }
                 } else {
--- a/widgetssdk/src/main/java/com/glia/widgets/launcher/ConfigurationManager.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/launcher/ConfigurationManager.kt
@@ -14,17 +14,26 @@ internal interface ConfigurationManager {
 
     val queueIdsObservable: Flowable<List<String>>
 
+    val visitorContextAssetId: String?
+
     /**
      * Applies [GliaWidgetsConfig] from the [GliaWidgets.init] configuration step
      */
     fun applyConfiguration(config: GliaWidgetsConfig)
 
     /**
-     * Retrives queue IDs
+     * Sets queue IDs
      * [queueIds] A list of queue IDs to be used for the engagement launcher.
      * When empty or invalid, the default queues will be used.
      */
     fun setQueueIds(queueIds: List<String>)
+
+    /**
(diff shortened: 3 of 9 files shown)


AndriiHorishniiMOC (approved):
LGTM!

DavDo (changes requested):
I would highly recommend avoiding passing this attribute through activities because this way, we're increasing complexity and risks of losing this value during this chain. For example, when navigating from SC to Live Chat/Audi/Video screens. We can store it in ConfigurationManager and update the value each time the `EngagementLauncher.start-smth()` is called(like we do with queues), so we can use up-to-date visitorContextAssetId each time we need it, from one source.

andrews-moc:
> I would highly recommend avoiding passing this attribute through activities because this way, we're increasing complexity and risks of losing this value during this chain. For example, when navigating from SC to Live Chat/Audi/Video screens. We can store it in ConfigurationManager and update the value each time the `EngagementLauncher.start-smth()` is called(like we do with queues), so we can use up-to-date visitorContextAssetId each time we need it, from one source.

@DavDo Agree with you. I was also thinking about it, however wasn't sure that it's possible after the refactoring. Eventually, I was able to implement the visitor context through the `ConfigurationManager` and added unit tests. Please have a look now.

widgetssdk/src/main/java/com/glia/widgets/di/RepositoryFactory.java
```diff
@@ -149,7 +149,11 @@ public PermissionsRequestRepository getPermissionsRequestRepository() {
 
     public EngagementRepository getEngagementRepository() {
         if (engagementRepository == null) {
-            engagementRepository = new EngagementRepositoryImpl(gliaCore, getOperatorRepository(), getQueueRepository());
+            engagementRepository = new EngagementRepositoryImpl(
+                gliaCore,
+                getOperatorRepository(),
+                getQueueRepository(),
+                configurationManager);
```
DavDo, line 156: Could you please move the bracket to the next line?

widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
```diff
@@ -213,7 +215,12 @@ internal class EngagementRepositoryImpl(
         queueIngDisposable.add(
             queueRepository.relevantQueueIds.subscribe { ids ->
                 if (ids.isNotEmpty()) {
-                    core.queueForEngagement(ids, mediaType, null, null, MEDIA_PERMISSION_REQUEST_CODE) {
+                    core.queueForEngagement(
```
DavDo, line 218: 👍

widgetssdk/src/main/java/com/glia/widgets/launcher/ConfigurationManager.kt
```diff
@@ -14,17 +14,26 @@ internal interface ConfigurationManager {
...
+    val visitorContextAssetId: String?
+
     /**
      * Applies [GliaWidgetsConfig] from the [GliaWidgets.init] configuration step
      */
     fun applyConfiguration(config: GliaWidgetsConfig)
 
     /**
-     * Retrives queue IDs
+     * Sets queue IDs
```
DavDo, line 25: 👍

widgetssdk/src/main/java/com/glia/widgets/launcher/ConfigurationManager.kt
```diff
@@ -14,17 +14,26 @@ internal interface ConfigurationManager {
...
      * When empty or invalid, the default queues will be used.
      */
     fun setQueueIds(queueIds: List<String>)
+
+    /**
+     * Sets the visitor context for the engagement.
+     *
+     * @param visitorContextAssetId a visitor context id from Glia Hub.
+     */
+    fun setVisitorContext(visitorContextAssetId: String)
```
DavDo, line 36: It would be nice to rename the function to `setVisitorContextAssetId()`

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
```diff
@@ -13,45 +13,69 @@ interface EngagementLauncher {
      *
      * @param activity The Activity used to launch the chat screen
      */
-    fun startChat(activity: Activity)
+    fun startChat(activity: Activity, visitorContext: String? = null)
```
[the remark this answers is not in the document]
DavDo: And in remaining functions as well

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
```diff
@@ -13,45 +13,69 @@ interface EngagementLauncher {
...
-    override fun startChat(activity: Activity) {
-        activityLauncher.launchChat(activity)
+    /**
+     * Starts a chat engagement.
+     *
+     * @param activity The Activity used to launch the chat screen
+     * @param visitorContext Optional visitor context id from Glia Hub
+     */
+    override fun startChat(activity: Activity, visitorContext: String?) {
+        activityLauncher.launchChat(activity, visitorContext)
```
DavDo, line 49: Why are we not updating the visitor context here instead of passing it to the activity launcher layer? This is the place where we receive it from the integrators, so the sooner we get rid of it, the better.

DavDo (commented):
Overall looks good! Added several suggestions.

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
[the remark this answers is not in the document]
andrews-moc: Fixed bd8af4284515b5561f26567935970b63416b8e5a

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
andrews-moc: Good point 👍 
Changed: 5f83329c2ff1558fcd77f7a520c209e27571beef

DavDo (approved):
Good Job! 👍

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/launcher/ConfigurationManager.kt
```diff
@@ -14,17 +14,26 @@ internal interface ConfigurationManager {
...
      * [queueIds] A list of queue IDs to be used for the engagement launcher.
      * When empty or invalid, the default queues will be used.
      */
     fun setQueueIds(queueIds: List<String>)
+
+    /**
+     * Sets the visitor context for the engagement.
+     *
+     * @param visitorContextAssetId a visitor context id from Glia Hub.
+     */
```
gugalo, line 35: This is an internal class and the method name is self explanatory. IMHO think no javadoc is needed here

gugalo: I see this in many other places. It is not a problem but I just don't see the point.

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/launcher/ConfigurationManager.kt
andrews-moc: You are right. It's for our internal use. I just saw that other functions are documented in the `ConfigurationManager` interface and decided to follow this approach. But I agree with you.

PR #1142, opened 2024-11-01
Introduce Intention Enum for opening Chat screen

DavDo:
This is the first review in this chain, so don't pay attention to `TODO`s, I added them for the next PRs in the chain.

**Jira issue:**
https://glia.atlassian.net/browse/MOB-3757

**What was solved?**
I introduced a new enum type - `Intention`, that will reflect all the intentions we open the `Chat Screen` and will fully replace the old `ChatType`. I also removed the `ChatType` and all the related use cases, so currently, all the business logic is broken. Each piece of business logic will be handled with the screen I work on in the future PRs in this chain.

**Release notes:**

 - [x] Feature(part of SC V2 project)
 - [ ] Ignore

**Additional info:**

**Screenshots:**


widgetssdk/src/main/java/com/glia/widgets/chat/Intention.kt
```diff
@@ -0,0 +1,24 @@
+package com.glia.widgets.chat
+
+internal enum class Intention {
+    RESTORE_CHAT,
+    SC_DIALOG_START_AUDIO,
+    SC_DIALOG_START_VIDEO,
```
andrews-moc, line 6: Just an idea. `SC_DIALOG_START_AUDIO` and `SC_DIALOG_START_VIDEO` sound more like the place from where the call originated. To sound more like an intention, maybe these names might be better: `CHAT_FROM_SC_AUDIO` and `CHAT_FROM_SC_VIDEO`?

andrews-moc (approved):
The overall idea looks good to me.

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/Intention.kt
DavDo: I think it sounds exactly like I wanted.
`SC_DIALOG_START_AUDIO` intention is - to open SC chat with the `leave sc` dialog and then, based on user input, stay in SC chat or start the audio call.

Maybe it makes sense to add a description for each intention?

andrews-moc: Ah, okay, agree. Yes, a description is definitely needed here.

DavDo: added

widgetssdk/src/main/java/com/glia/widgets/chat/Intention.kt
```diff
@@ -0,0 +1,37 @@
...
+    RESTORE_CHAT,
+
+    /* Open Secure Conversation chat with Leave dialog, and if the visitor chooses to leave, open chat screen and start enqueueing for audio */
+    SC_DIALOG_START_AUDIO,
+
+    /* Open Secure Conversation chat with Leave dialog, and if the visitor chooses to leave, open chat screen and  start enqueueing for video */
+    SC_DIALOG_START_VIDEO,
+
+    /* Open Secure Conversation chat with Leave dialog, and if the visitor chooses to leave, start enqueueing for chat */
+    SC_DIALOG_ENQUEUE,
```
andrews-moc, line 14: @DavDo Not critical. Will it be better to rename `SC_DIALOG_ENQUEUE` to `SC_DIALOG_START_CHAT` to align with similar flows `SC_DIALOG_START_AUDIO` and `SC_DIALOG_START_VIDEO`?

DavDo: Good point!
I also came to that conclusion and renamed it in my ongoing task to `SC_DIALOG_START_ENQUEUEING`, this will be a bit clearer since we will be in the chat screen, but when the visitor chooses to leave, we must start enqueueing.

andrews-moc: Yes. The matter is that ENQUEUE might mean any type of engagement: enqueue to chat/audio/video, isn't it? That's why I prefer `SC_DIALOG_START_CHAT` or `SC_DIALOG_ENQUEUE_CHAT`. And I would suggest to align the names for all these cases: for chat, audio, video.

DavDo: But from the chat screen point, these cases are different.

- SC_DIALOG_START_AUDIO, SC_DIALOG_START_VIDEO → will open the call screen
- SC_DIALOG_START_ENQUEUEING → will start enqueueing

Having this said, I'd prefer `SC_DIALOG_ENQUEUE_FOR_TEXT` since the engagement type is `Engagement.MediaType.TEXT`

andrews-moc: Okay, I think I got your point now. Agree.

PR #1143, opened 2024-11-04
Make SC APIs work seamlessly when sdk is not initialized

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3761

**What was solved?**
Added initialization and authentication checks for the SC `getUnreadMessagesCound()` and `hasSecureConversations()` inside their use cases to prevent crashes in the Entry Widget.

**Release notes:**

 - [x] Feature(Part of SC V2 feature)
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/GetUnreadMessagesCountWithTimeoutUseCase.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/GetUnreadMessagesCountWithTimeoutUseCase.kt
@@ -1,7 +1,9 @@
 package com.glia.widgets.core.secureconversations.domain
 
 import androidx.annotation.VisibleForTesting
+import com.glia.widgets.chat.domain.IsAuthenticatedUseCase
 import com.glia.widgets.core.secureconversations.SecureConversationsRepository
+import com.glia.widgets.di.GliaCore
 import com.glia.widgets.helper.Logger
 import io.reactivex.rxjava3.core.Single
 import java.util.concurrent.TimeUnit
@@ -20,9 +22,16 @@ internal const val TIMEOUT_SEC = 3L
 
 internal const val NO_UNREAD_MESSAGES = 0
 
-internal class GetUnreadMessagesCountWithTimeoutUseCase(private val repository: SecureConversationsRepository) {
+internal class GetUnreadMessagesCountWithTimeoutUseCase(
+    private val repository: SecureConversationsRepository,
+    private val isAuthenticatedUseCase: IsAuthenticatedUseCase,
+    private val core: GliaCore
+) {
 
     /**
+     * This function provides a default value instead of an error.
+     * If error handling is needed, use the repository function directly.
+     *
      * @return [NO_UNREAD_MESSAGES] if the current socket doesn't signal a success value within the specified [TIMEOUT_SEC] window.
      *
      * This is combined with the chat transcript result to avoid "Jumping UI" when adding new messages'
@@ -30,7 +39,13 @@ internal class GetUnreadMessagesCountWithTimeoutUseCase(private val repository: 
      * there is no warranty that it will return anything so timeout is added to make sure that it will return [NO_UNREAD_MESSAGES]
      * after the timeout if there is no answer from socket yet.
      */
-    operator fun invoke(): Single<Int> = Single.create {
+    operator fun invoke(): Single<Int> = when {
+        core.isInitialized.not() -> Single.just(NO_UNREAD_MESSAGES)
+        isAuthenticatedUseCase().not() -> Single.just(NO_UNREAD_MESSAGES)
+        else -> getUnreadMessagesCountWithTimeout()
+    }
+
+    private fun getUnreadMessagesCountWithTimeout() = Single.create {
         repository.getUnreadMessagesCount { count, exception ->
             if (it.isDisposed) return@getUnreadMessagesCount
             if (exception != null) {
--- a/widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/HasPendingSecureConversationsWithTimeoutUseCase.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/HasPendingSecureConversationsWithTimeoutUseCase.kt
@@ -1,0 +1,33 @@
+package com.glia.widgets.core.secureconversations.domain
+
+import com.glia.widgets.chat.domain.IsAuthenticatedUseCase
+import com.glia.widgets.core.secureconversations.SecureConversationsRepository
+import com.glia.widgets.di.GliaCore
+import com.glia.widgets.entrywidget.EntryWidget
+import com.glia.widgets.launcher.EngagementLauncher
+import io.reactivex.rxjava3.core.Single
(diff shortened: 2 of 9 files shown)


widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
```diff
@@ -37,11 +37,11 @@ internal class EntryWidgetImpl(
     private val activityLauncher: ActivityLauncher,
     private val themeManager: UnifiedThemeManager,
     private val entryWidgetHideController: EntryWidgetHideController,
-    private val secureConversationsRepository: SecureConversationsRepository
+    private val hasPendingSecureConversationsWithTimeoutUseCase: HasPendingSecureConversationsWithTimeoutUseCase
 ) : EntryWidget {
 
     override fun show(activity: Activity) {
-        secureConversationsRepository.getHasPendingSecureConversationsWithTimeout().unSafeSubscribe {
+        hasPendingSecureConversationsWithTimeoutUseCase().unSafeSubscribe {
```
AndriiHorishniiMOC, line 44: Does it mean the visitor can see a delay when they click some button and see the result?

DavDo: Since this property should be present right after the initialization in the `visitor_state,` we will not have a delay. But if it is somehow missing, the default `false` value will be returned after 1 second of waiting.

AndriiHorishniiMOC: Why can't we display the Entry Widget loading state and, after everything is prepared, go to the SC or display the available media types?

DavDo: I think that will not be user-friendly. So, the visitors will see the entry widget with the loading state, and then the entry widget will disappear, and the chat screen will appear automatically. That might be not very clear.
Besides that, this will happen in very, very corner cases when the system is healthy, but this attribute is somehow missing from the `visitor_state` or, for some reason, the socket disconnected, so it is safe to just open the entry widget.

widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/GetUnreadMessagesCountWithTimeoutUseCase.kt
```diff
@@ -30,7 +36,13 @@ internal class GetUnreadMessagesCountWithTimeoutUseCase(private val repository:
      * there is no warranty that it will return anything so timeout is added to make sure that it will return [NO_UNREAD_MESSAGES]
      * after the timeout if there is no answer from socket yet.
      */
-    operator fun invoke(): Single<Int> = Single.create {
+    operator fun invoke(): Single<Int> = when {
+        core.isInitialized.not() -> Single.just(NO_UNREAD_MESSAGES)
+        isAuthenticatedUseCase().not() -> Single.just(NO_UNREAD_MESSAGES)
+        else -> getUnreadMessagesCountWithTimeout()
```
andrews-moc, line 42: Although it can also work well, I think that it's not perfect to return `NO_UNREAD_MESSAGES` for Glia not initialized and visitor not authenticated cases because, in fact, we don't know this rather than there are no unread messages. Ideally, `GetUnreadMessagesCountWithTimeoutUseCase` should return an error. This would allow the caller to distinguish when to show "messaging unavailable" vs "no unread messages".

But I'm not strongly against the solution that you implemented. I just want to bring my concern up to know if you took this into account.

widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/HasPendingSecureConversationsWithTimeoutUseCase.kt
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * @return `false` if the SDk is not initialized, visitor is not authenticated or
+     * if the current socket doesn't signal a success value within the specified 1 sec window.
+     *
+     * Since we're using this function in [EntryWidget] and [EngagementLauncher] where we need the immediate UI response to the user interaction,
+     *  * it is important to to avoid delays for the user.
+     */
+    operator fun invoke(): Single<Boolean> = when {
+        core.isInitialized.not() -> Single.just(false)
+        isAuthenticatedUseCase().not() -> Single.just(false)
```
andrews-moc, line 25: The same as I commented for `NO_UNREAD_MESSAGES`. 😉

widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/GetUnreadMessagesCountWithTimeoutUseCase.kt
DavDo: What will give it to us?
We will have to map the error, regardless of the type, to the default value in all the places we use these use cases.
I mean, these functions are designed for the use cases where we always need a value regardless of the request result. So, if we want to deal with errors, we can still use the repository functions.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
AndriiHorishniiMOC: /Turning on the QA mode

It doesn't matter to me how often it happens, as long as it's possible. Displaying the loading screen is way more user-friendly than nothing to display. Moreover, we do not even provide some callback interfaces for integrators.

/Turning off the QA mode

Nice solution! :)

widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/GetUnreadMessagesCountWithTimeoutUseCase.kt
andrews-moc: Okay, maybe this is worth adding as a comment: 
> if we want to deal with errors, we can still use the repository functions.

and add references to the functions

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidget.kt
DavDo: > It doesn't matter to me how often it happens, as long as it's possible. Displaying the loading screen is way more user-friendly than nothing to display

I'm sorry Andrii's QA mode 🙂, but I do not agree with this statement.
The use case is to open the chat directly, instead of the entry widget, if there is pending SC. This should be very smooth for the users. I think a 1-second timeout is not enough to make visible delays for users, but if you insist, we can decrease that time or return the default value immediately if there is no value.


> Moreover, we do not even provide some callback interfaces for integrators

But in which cases integrators will need to know about pending SC?


> Nice solution! :)

Thanks 💟

widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/GetUnreadMessagesCountWithTimeoutUseCase.kt
DavDo: added

andrews-moc (approved):

PR #1162, opened 2024-11-14
Separate Delivered and Tap To Retry views

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3799

>[!NOTE]
> Tests will be in separate PR.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

https://github.com/user-attachments/assets/f8ea825b-cda6-4d39-8ed3-7c96956ba0d9


AndriiHorishniiMOC (commented):
Only one question: is Tap To Retry view at least 48 dp height? We must consider the minimum touch size to be compatible with the ADA.

DavDo:
> Only one question: is Tap To Retry view at least 48 dp height? We must consider the minimum touch size to be compatible with the ADA.

I made some improvements and attached a recording to the description.

andrews-moc (approved):
I don't have any comments, looks good to me.

PR #1173, opened 2024-11-27
Fix RX timeout issue

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3839

**What was solved?**

In some places the Observable#timeout() had unwanted results. Specifically:
- It terminated observable breaking delivery of events after the timeout triggered
- Plan was for it to fire timeout exception only if there was no previous event but it also triggered even when there was valid values.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


widgetssdk/src/main/java/com/glia/widgets/helper/rx/RxExtensions.kt
```diff
@@ -0,0 +1,15 @@
...
+fun <T: Any> Observable<T>.timeoutFirstWithDefaultUntilChanged(timeout: Long, unit: TimeUnit, defaultValue: T): Observable<T> {
+    return this
+        .firstOrError()
+        .timeout(timeout, unit)
+        .onErrorReturnItem(defaultValue)
+        .flatMapObservable { firstValue ->
+            this.startWithItem(firstValue)
+        }
+        .distinctUntilChanged()
+}
```
gugalo, line 15: Special thanks for @DavDo for providing this solution 🙇‍♂️

DavDo (approved):
LGTM

andrews-moc (approved):

PR #1169, opened 2024-11-21
Add 'Need Live Support?' banner to SC chat

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3639

**What was solved?**

To maintain a smaller PR size and have clearer progress a couple of known issues will be resolved in separate PR(s):
- New snapshot tests for secure conversations chat with the new banner
- Fix business logic for detecting pending secure conversation engagement
- Restore 'Need live support' open drawer on screen rotation
- ✅ Fix missing dividers

**Release notes:**

 - [x] Feature: **part of SCv2 project**
 - [x] Ignore
 - [ ] Release notes (Is it clear from the description here?)
 - [ ] Migration guide (If changes are needed for integrator already using the SDK - what needs to be communicated? Add underneath please)

**Additional info:**

 - [ ] Is the feature sufficiently tested? All tests fixed? Necessary unit, acceptance, snapshots added? Check that at least new public classes & methods are covered with unit tests
 - [ ] Did you add logging beneficial for troubleshooting of customer issues?
 - [ ] **Did you add new logging?** We would like the logging between platforms to be similar. Refer to [**Logging from Android SDKs** → **Things to consider for newly added logs**](https://glia.atlassian.net/wiki/spaces/ENG/pages/3568861468/Logging+from+Android+SDKs#Things-to-consider-for-newly-added-logs) in Confluence for more information.

**Screenshots:**

https://github.com/user-attachments/assets/e0cd8ce4-fd7f-4a54-b991-85f720a0542b

PR #1174, opened 2024-11-27
Prevent recursive dismiss() call

andrews-moc:
**Jira issue:**
[[Android] SDK crashes on Entry Widget item tap](https://glia.atlassian.net/browse/MOB-3845)

**What was solved?**
Crash reason is [here](https://github.com/salemove/android-sdk-widgets/pull/1169/files#diff-4f29a0397380e2017630aae072bd897cc65374955c607c104dfddcff4bfb6cadR63-R69).
`dismiss()` was called recursively.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**


DavDo (approved):
Good Catch!

gugalo (approved):

PR #1175, opened 2024-11-28
Consider the ongoing sc when checking for secure messaging availability

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3652

**What was solved?**

**Release notes:**

 - [X] Feature(Part of SC V2)
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetContract.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetContract.kt
@@ -4,6 +4,9 @@ import android.app.Activity
 import com.glia.widgets.base.BaseController
 import com.glia.widgets.base.BaseView
 
+private const val LOWEST_ORDER = Int.MAX_VALUE
+private const val HIGHEST_ORDER = Int.MIN_VALUE
+
 internal interface EntryWidgetContract {
 
     enum class ViewType {
@@ -12,16 +15,16 @@ internal interface EntryWidgetContract {
         MESSAGING_LIVE_SUPPORT
     }
 
-    sealed class ItemType(private val order: Int) : Comparable<ItemType> {
-        data object VideoCall : ItemType(0)
-        data object AudioCall : ItemType(1)
-        data object Chat : ItemType(2)
-        data class Messaging(val value: Int) : ItemType(3)
-        data object LoadingState : ItemType(10)
-        data object EmptyState : ItemType(10)
-        data object SdkNotInitializedState : ItemType(10)
-        data object ErrorState : ItemType(10)
-        data object ProvidedBy : ItemType(11)
+    sealed class ItemType(private val order: Int = HIGHEST_ORDER) : Comparable<ItemType> {
+        data object VideoCall : ItemType(order = 0)
+        data object AudioCall : ItemType(order = 1)
+        data object Chat : ItemType(order = 2)
+        data class Messaging(val value: Int) : ItemType(order = 3)
+        data object LoadingState : ItemType()
+        data object EmptyState : ItemType()
+        data object SdkNotInitializedState : ItemType()
+        data object ErrorState : ItemType()
+        data object ProvidedBy : ItemType(order = LOWEST_ORDER)
 
         override fun compareTo(other: ItemType): Int {
             return this.order.compareTo(other.order)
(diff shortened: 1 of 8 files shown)


widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetContract.kt
```diff
@@ -4,6 +4,8 @@ import android.app.Activity
 import com.glia.widgets.base.BaseController
 import com.glia.widgets.base.BaseView
 
+private const val NO_ORDER = Int.MIN_VALUE
```
andrews-moc, line 7: MIN_VALUE will actually show an item at the very top. So, it like the highest order in Entry Widget. Maybe it makes sense to write `private const val LOWEST_ORDER = Int.MAX_VALUE` and use `LOWEST_ORDER` for `ProvidedBy` as well?

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
```diff
@@ -1,109 +1,149 @@
...
+    }
+
+    private val errorState: List<EntryWidgetContract.ItemType.EmptyState> by lazy {
+        listOf(EntryWidgetContract.ItemType.EmptyState)
+    }
+
+    private val messagingChatObservableItemType: Observable<List<EntryWidgetContract.ItemType>>
+        get() = queueStateObservable.map(::mapMessagingQueueState)
+
+    /** Actually the Bottom Sheet must be skipped when there is ongoing SC, so technically no need for ongoing SC check in this case.
```
andrews-moc, line 54: This sentence sounds contradictory to me because logically
ongoing SC check should be executed to know that "there is ongoing SC". That's why it sounds confusing to me that Bottom Sheet could be skipped without a check.

andrews-moc: Hm, maybe I misunderstood the comment. 🤔

DavDo: I will try to improve this comment, or maybe remove it because it was written particularly for the bottom sheet, but now the bottom sheet and embedded view share business logic.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetContract.kt
DavDo: I know it will not happen, but in the case the `provided by` and the `error` items meet, the error item should have higher priority, so the order is right, but the naming doesn't reflect this. so maybe it makes sense to rename the `NO_ORDER` to `HIGHER_ORDER` or `DEFAULT_ORDER` or just put `Int.MIN_VALUE`?

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
DavDo: I decided to remove it. It is not relevant anymore.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetContract.kt
DavDo: changed to `private const val HIGHEST_ORDER = Int.MIN_VALUE`

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetContract.kt
```diff
@@ -12,16 +14,16 @@ internal interface EntryWidgetContract {
...
+    sealed class ItemType(private val order: Int = HIGHEST_ORDER) : Comparable<ItemType> {
+        data object VideoCall : ItemType(order = 0)
+        data object AudioCall : ItemType(order = 1)
+        data object Chat : ItemType(order = 2)
+        data class Messaging(val value: Int) : ItemType(order = 3)
+        data object LoadingState : ItemType()
+        data object EmptyState : ItemType()
+        data object SdkNotInitializedState : ItemType()
+        data object ErrorState : ItemType()
+        data object ProvidedBy : ItemType(order = Int.MAX_VALUE)
```
andrews-moc, line 26: NIT. Maybe it might improve the readability and consistency if you extract this value to a constant as well `private const val LOWEST_ORDER = Int.MAX_VALUE`

andrews-moc (approved):

DavDo:
!squash

PR #1168, opened 2024-11-20
Adjust Entry Widget UI based on designer review

andrews-moc:
**Jira issue:**
[MOB-3754](https://glia.atlassian.net/browse/MOB-3754)

**What was solved?**
Adjust Entry Widget UI based on designer review:
- Adjust font for engagement type title and description
- Adjust bottom margin for "Powered by" and align it with the Glia logo baseline

**Release notes:**

 - [x] Feature
 - [x] Ignore
 - [ ] Release notes (Is it clear from the description here?)
 - [ ] Migration guide (If changes are needed for integrator already using the SDK - what needs to be communicated? Add underneath please)

**Additional info:**

 - [x] Is the feature sufficiently tested? All tests fixed? Necessary unit, acceptance, snapshots added? Check that at least new public classes & methods are covered with unit tests
 - [ ] Did you add logging beneficial for troubleshooting of customer issues?
 - [ ] **Did you add new logging?** We would like the logging between platforms to be similar. Refer to [**Logging from Android SDKs** → **Things to consider for newly added logs**](https://glia.atlassian.net/wiki/spaces/ENG/pages/3568861468/Logging+from+Android+SDKs#Things-to-consider-for-newly-added-logs) in Confluence for more information.

**Screenshots:**

[MOB-3754]: https://glia.atlassian.net/browse/MOB-3754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1177, opened 2024-11-29
Fix Entry Widget watermark visibility is not configurable

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3853

**What was solved?**
> [!IMPORTANT]
> I decided to hide text and icons inside the 'Powered By' item instead of omitting this item entirely to maintain a small margin after the last divider. I also adjusted some margins related to this item, which is why the snapshots differ for the entry widget. However, I have no idea why the snapshots for the call screen have changed.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


DavDo:
Converted to draft since we have a different design for this case, and also, all the error cases contain the `Powered By` item.

DavDo:
> Converted to draft since we have a different design for this case, and also, all the error cases contain the `Powered By` item.

It seems I broke it in my previous PR(

DavDo:
Fixed everything now

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
```diff
@@ -86,17 +90,23 @@ internal class EntryWidgetController @JvmOverloads constructor(
...
             QueuesState.Loading -> messagingOrDefault(loadingState)
             is QueuesState.Error -> messagingOrDefault(errorState)
             is QueuesState.Queues -> mapEntryWidgetQueues(queuesState.queues, unreadMessagesCount, hasOngoingSC)
         }
+
+        if (view.whiteLabel) {
+            return items.sorted()
+        }
+
+        return items.plus(EntryWidgetContract.ItemType.PoweredBy).sorted()
```
andrews-moc, line 109: I expected to see vice-a-versa:
```
if (white label enabled) {
   add PoweredBy item to the bottom and return
}

otherwise - just return
```
Am I missing anything?

widgetssdk/src/test/snapshots/images/com.glia.widgets.entrywidget_EntryWidgetBottomSheetTest_contactsDefaultTheme.png
andrews-moc, line 1: The Glia logo moved up and is not not aligned with text now, unfortunately. See the changes to `entry_widget_powered_by_item.xml` in [this](https://github.com/salemove/android-sdk-widgets/pull/1168/files) PR. 
Screenshot 2024-12-02 at 09 57 59

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
DavDo: image
I just copied the behavior from the other parts of SDK.

andrews-moc (commented):
Have you tried reverting the Call screen screenshots, are snapshot tests failing? I mean that if you run 'record snapshots' for the whole entry widget package, it will create new files for all Entry Widget tests. However, only screenshots with 'Powered by' are affected and all other files can be reverted.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
andrews-moc: Hm, maybe 'whiteLabel true' means 'hide while label. And 'whiteLabel false' means 'do not hide it'? 🤔

widgetssdk/src/test/snapshots/images/com.glia.widgets.entrywidget_EntryWidgetBottomSheetTest_contactsDefaultTheme.png
DavDo: image
image
This is how it looks in design.
Figma has no paddings for the logo and text separately, but they look aligned vertically in the center.

DavDo:
> Have you tried reverting the Call screen screenshots, are snapshot tests failing? I mean that if you run 'record snapshots' for the whole entry widget package, it will create new files for all Entry Widget tests. However, only screenshots with 'Powered by' are affected and all other files can be reverted.

I haven't tried.
It's better to have fresh ones for the call screen if they differ.
What do you think?

widgetssdk/src/test/snapshots/images/com.glia.widgets.entrywidget_EntryWidgetBottomSheetTest_contactsDefaultTheme.png
andrews-moc: Looks like Risto hasn't updated the design because on design review he pointed out that it doesn't look good if 'glia' logo in not aligned with 'Powered by'.

andrews-moc:
> > Have you tried reverting the Call screen screenshots, are snapshot tests failing? I mean that if you run 'record snapshots' for the whole entry widget package, it will create new files for all Entry Widget tests. However, only screenshots with 'Powered by' are affected and all other files can be reverted.
> 
> I haven't tried. It's better to have fresh ones for the call screen if they differ. What do you think?

I think we need to update only those tests that are affected. All the rest should stay unchanged. Otherwise we can miss some unexpected changes (test indicates a change, that is difficult to notice and human eye doesn't recognise the change by a mistake).

DavDo:
> It's worth adding snapshot tests for the case when whiteLabel is disabled, to test the margins and paddings. If you don't have time, I can add it once I recover.

I think it makes sense to do it in a separate ticket because we need to create a separate theme for that.
I will create a ticket for that, but I will work on my ongoing tasks at this moment.

DavDo:
> > It's worth adding snapshot tests for the case when whiteLabel is disabled, to test the margins and paddings. If you don't have time, I can add it once I recover.
> 
> I think it makes sense to do it in a separate ticket because we need to create a separate theme for that. I will create a ticket for that, but I will work on my ongoing tasks at this moment.

I created a [ticket](https://glia.atlassian.net/browse/MOB-3854) as follows.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
```diff
@@ -35,8 +35,7 @@ internal class EntryWidgetController @JvmOverloads constructor(
             EntryWidgetContract.ItemType.LoadingState,
             EntryWidgetContract.ItemType.LoadingState,
             EntryWidgetContract.ItemType.LoadingState,
-            EntryWidgetContract.ItemType.LoadingState,
-            EntryWidgetContract.ItemType.ProvidedBy
+            EntryWidgetContract.ItemType.LoadingState
```
andrews-moc, line 38: Does it makes sense to add the white label item logic here, to keep it together with all loadingState items?

DavDo:
> > > Have you tried reverting the Call screen screenshots, are snapshot tests failing? I mean that if you run 'record snapshots' for the whole entry widget package, it will create new files for all Entry Widget tests. However, only screenshots with 'Powered by' are affected and all other files can be reverted.
> > 
> > 
> > I haven't tried. It's better to have fresh ones for the call screen if they differ. What do you think?
> 
> I think we need to update only those tests that are affected. All the rest should stay unchanged. Otherwise we can miss some unexpected changes (test indicates a change that is difficult to notice and human eye doesn't recognise the change by a mistake).

I agree, but since we are still in the development phase, I'd prefer to release with the latest snapshots.

andrews-moc:
> > > > Have you tried reverting the Call screen screenshots, are snapshot tests failing? I mean that if you run 'record snapshots' for the whole entry widget package, it will create new files for all Entry Widget tests. However, only screenshots with 'Powered by' are affected and all other files can be reverted.
> > > 
> > > 
> > > I haven't tried. It's better to have fresh ones for the call screen if they differ. What do you think?
> > 
> > 
> > I think we need to update only those tests that are affected. All the rest should stay unchanged. Otherwise we can miss some unexpected changes (test indicates a change that is difficult to notice and human eye doesn't recognise the change by a mistake).
> 
> I agree, but since we are still in the development phase, I'd prefer to release with the latest snapshots.

The matter is that we have passed design review for Entry Widget already. That is why I would not consider that design is in development, I would rather treat it as 'set in stone'.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
gugalo: ☝️ Thats true. This confused me at first so much that I had to make it supper visual int [this docs](https://glia.atlassian.net/wiki/spaces/ENG/pages/3390898324/White+label+configuration+in+remote+configurations)

gugalo (approved):

widgetssdk/src/test/snapshots/images/com.glia.widgets.call_CallViewAudioSnapshotTest_audioCallStarted.png
gugalo, line 1: > However, I have no idea why the snapshots for the call screen have changed.

Can you share the paparazzi diff image that highlights the differences?

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
DavDo: I decided to keep this logic in one place in a sense of clarity so that it can be applied to all possible cases.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
DavDo: But from the integrator's perspective, it sounds correct. So, white labeling from their side means that the feature will be present with their own branding.
`White label products are sold by retailers with their own branding and logo but the products themselves are manufactured by a third party`

DavDo:
> > > > > Have you tried reverting the Call screen screenshots, are snapshot tests failing? I mean that if you run 'record snapshots' for the whole entry widget package, it will create new files for all Entry Widget tests. However, only screenshots with 'Powered by' are affected and all other files can be reverted.
> > > > 
> > > > 
> > > > I haven't tried. It's better to have fresh ones for the call screen if they differ. What do you think?
> > > 
> > > 
> > > I think we need to update only those tests that are affected. All the rest should stay unchanged. Otherwise we can miss some unexpected changes (test indicates a change that is difficult to notice and human eye doesn't recognise the change by a mistake).
> > 
> > 
> > I agree, but since we are still in the development phase, I'd prefer to release with the latest snapshots.
> 
> The matter is that we have passed design review for Entry Widget already. That is why I would not consider that design is in development, I would rather treat it as 'set in stone'.

Correct, but other changes can affect the generated snapshots as well(environment, paparazzi version, etc.).
So, the snapshots, in my opinion, should reflect the current state of the feature branch.
My concerns are about the fact that paparazzi generate different snapshots for the Call screen, which means the circumstances for that particular screen are changed, so we should have fresh examples.

andrews-moc:
> Correct, but other changes can affect the generated snapshots as well(environment, paparazzi version, etc.). So, the snapshots, in my opinion, should reflect the current state of the feature branch. My concerns are about the fact that paparazzi generate different snapshots for the Call screen, which means the circumstances for that particular screen are changed, so we should have fresh examples.

In my understanding, each time I run 'record snapshot tests', Paparazzi generates new files even though there were no any changes. If so, then there is no sense to replace the images for tests that are not affected. In other words, I don't see a reason for replacing the images for the tests that are succeeding.

DavDo:
> > Correct, but other changes can affect the generated snapshots as well(environment, paparazzi version, etc.). So, the snapshots, in my opinion, should reflect the current state of the feature branch. My concerns are about the fact that paparazzi generate different snapshots for the Call screen, which means the circumstances for that particular screen are changed, so we should have fresh examples.
> 
> In my understanding, each time I run 'record snapshot tests', Paparazzi generates new files even though there were no any changes. If so, then there is no sense to replace the images for tests that are not affected. In other words, I don't see a reason for replacing the images for the tests that are succeeding.

Yes, but those new files should be the same for all cases if there are no changes.

widgetssdk/src/test/snapshots/images/com.glia.widgets.entrywidget_EntryWidgetBottomSheetTest_contactsDefaultTheme.png
andrews-moc: > I think it makes sense to ask Risto to update the design ASAP since, for us, the design is the source of truth.

For sure, agree. I have asked Risto in a private message.

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
andrews-moc: Oh, now I understand. It's correct then. The best we can do is to add a comment to this code for our grandchildren who will also look into this code...

widgetssdk/src/test/snapshots/images/com.glia.widgets.call_CallViewAudioSnapshotTest_audioCallStarted.png
DavDo: There are no diff images generated for Call views.

gugalo: Oh.. Got it

widgetssdk/src/test/snapshots/images/com.glia.widgets.entrywidget_EntryWidgetBottomSheetTest_contactsDefaultTheme.png
DavDo: I rolled back the `XML` changes.

DavDo:
!squash

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
AndriiHorishniiMOC: Why not?
```suggestion
        if (!view.whiteLabel) {
            items.plus(EntryWidgetContract.ItemType.PoweredBy)
        }

        return items.sorted()
```

AndriiHorishniiMOC (approved):

widgetssdk/src/main/java/com/glia/widgets/entrywidget/EntryWidgetController.kt
DavDo: makes sense

PR #1178, opened 2024-12-02
Fix send and attachment buttons overlap

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3844

**What was solved?**
Fix the send and attachment buttons overlaping.
A side effect of making the layout flat is that now those buttons have slight animation.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**
[Screen_recording_20241202_131317.webm](https://github.com/user-attachments/assets/a974c86b-c4fe-48be-a4a8-7ccdbb81550c)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ allprojects {
     // Needed for some automated and manual testing (e.g: acceptance tests)
     mavenLocal()
     //TODO switch to the release version before releasing core SDK
-    maven { url = "https://s01.oss.sonatype.org/content/repositories/comglia-1329" }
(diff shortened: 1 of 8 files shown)


DavDo (approved):

andrews-moc (commented):
Could you please add an overview of the changes because I'm not sure I understand what exactly was causing the issue among these changes.

Also, what do you mean by "making the layout flat"? It makes me think about the elevation, but I don't see the changes in it.

PR #1184, opened 2024-12-11
Implement Leave Current Conversation dialog

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3637

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/res/values/new_strings.xml
+++ b/widgetssdk/src/main/res/values/new_strings.xml
@@ -170,6 +170,10 @@
     <string name="secure_messaging_chat_banner_bottom">@string/message_center_bottom_banner</string>
     <string name="secure_messaging_chat_unavailable_message">@string/message_center_error_unavailable_message</string>
     <string name="secure_messaging_chat_banner_top">@string/need_live_support_label</string>
+    <string name="secure_messaging_chat_leave_current_conversation_title">@string/message_center_leaving_conversation_title</string>
+    <string name="secure_messaging_chat_leave_current_conversation_message">@string/message_center_leaving_conversation_message</string>
+    <string name="secure_messaging_chat_leave_current_conversation_button_positive">@string/message_center_leaving_conversation_positive_button</string>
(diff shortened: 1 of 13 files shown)


widgetssdk/src/main/res/values/new_strings.xml
```diff
@@ -170,6 +170,10 @@
     <string name="secure_messaging_chat_banner_bottom">@string/message_center_bottom_banner</string>
     <string name="secure_messaging_chat_unavailable_message">@string/message_center_error_unavailable_message</string>
     <string name="secure_messaging_chat_banner_top">@string/need_live_support_label</string>
+    <string name="secure_messaging_chat_leave_current_conversation_title">@string/message_center_leaving_conversation_title</string>
```
AndriiHorishniiMOC, line 173: Do we use references for the new strings?
As I remember, we need to put there just a value.

AndriiHorishniiMOC (approved):

widgetssdk/src/main/res/values/new_strings.xml
gugalo: I have no idea what is the right approach.
Is there some discussion that has been done regarding this before?

andrews-moc: I know only the [String resources key namingagreements](https://glia.atlassian.net/wiki/spaces/ENG/pages/3433726090/String+resources+key+naming+agreements) doc. However, it doesn't say anything about this case. Looks like something that needs clarification, indeed.

andrews-moc (approved):

widgetssdk/src/main/res/values/new_strings.xml
gugalo: I will close this PR for now as is. But it would be great to bring this to discussion on tomorrow's standup

PR #1206, opened 2025-01-10
Implement EngagementLauncher business logic for ongoing live engagement

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3932

**What was solved?**

**Release notes:**

 - [x] Feature(SC + Entry Widget)
 - [ ] Ignore

**Additional info:**

**Screenshots:**


widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
```diff
@@ -2,50 +2,96 @@ package com.glia.widgets.launcher
...
+    fun startChat(context: Context) = startChat(context, null)
+
     /**
      * Starts a chat engagement.
      *
      * @param context Activity or Context used to launch the chat screen
      * @param visitorContextAssetId Optional visitor context id from Glia Hub
      */
-    fun startChat(context: Context, visitorContextAssetId: String? = null)
+    @FeatureUnderConsiderationForRemoval
```
DavDo, line 45: What do you think about this solution?
As the `@JvmOverloads` annotation is not supported for interfaces, integrators using Java must explicitly pass null for `visitorContextAssetId`. For this reason, I have defined a new function instead.
Later I added this annotation to prevent integrators from using the functions with `visitorContextAssetId` until they need it. To do that, they have to mark their functions with the `@OptIn(FeatureUnderConsiderationForRemoval::class)` annotation, otherwise it won't compile.

widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/HasOngoingSecureConversationUseCase.kt
```diff
@@ -22,12 +20,19 @@ internal class HasOngoingSecureConversationUseCase(
             secureConversationsRepository.unreadMessagesCountObservable,
             engagementStateUseCase()
         ) { pendingSecureConversations, unreadMessagesCount, state ->
-            isAuthenticatedUseCase() &&
-                (pendingSecureConversations || unreadMessagesCount > 0 || state is State.TransferredToSecureConversation)
+            pendingSecureConversations || unreadMessagesCount > 0 || state is State.TransferredToSecureConversation
```
DavDo, line 23: Removed the `isAuthenticatedUseCase()` check, as it was included in the `unread_messages_count` observable on the core SDK side.

DavDo:
closed in favor of #1210

PR #1210, opened 2025-01-13
Implement EngagementLauncher business logic for ongoing live engagement

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3932

**What was solved?**

**Release notes:**

 - [x] Feature(SC v2)
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerActivityWatcher.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerActivityWatcher.kt
@@ -1,6 +1,7 @@
 package com.glia.widgets.callvisualizer
 
 import android.app.Activity
+import androidx.annotation.StringRes
 import com.glia.widgets.R
 import com.glia.widgets.base.BaseSingleActivityWatcher
 import com.glia.widgets.callvisualizer.controller.CallVisualizerContract
@@ -14,7 +15,6 @@ import com.glia.widgets.launcher.ActivityLauncher
 import com.glia.widgets.locale.LocaleProvider
 import com.glia.widgets.locale.LocaleString
 import com.glia.widgets.view.Dialogs
-import com.glia.widgets.view.snackbar.SnackBarDelegate
 import com.glia.widgets.view.snackbar.SnackBarDelegateFactory
 import com.glia.widgets.view.unifiedui.theme.UnifiedThemeManager
 import com.glia.widgets.webbrowser.WebBrowserActivity
@@ -45,7 +45,8 @@ internal class CallVisualizerActivityWatcher(
             activity == null || activity.isFinishing -> Logger.d(TAG, "skipping.. activity is null or finishing")
             activity is WebBrowserActivity && state is ControllerState.DisplayConfirmationDialog -> Logger.d(TAG, "skipping.. WebBrowser is open")
             activity is WebBrowserActivity && state is ControllerState.OpenWebBrowserScreen -> event.consume { controller.onWebBrowserOpened() }
-            state is ControllerState.ShowTimeoutSnackBar -> event.consume { showSnackBar(activity) }
+            state is ControllerState.ShowTimeoutSnackBar -> event.consume { showTimedOutSnackBar(activity) }
+            state is ControllerState.ShowAlreadyInCvSnackBar -> event.consume { showAlreadyInCvSnackBar(activity) }
             //Ensure this state remains unconsumed until the opening of the WebBrowserActivity.
             state is ControllerState.OpenWebBrowserScreen -> openWebBrowser(activity, state.title, state.url)
             state is ControllerState.DisplayVisitorCodeDialog -> displayVisitorCodeDialog(activity)
@@ -94,15 +95,15 @@ internal class CallVisualizerActivityWatcher(
         activityLauncher.launchWebBrowser(activity, title, url)
     }
 
-    private fun showSnackBar(activity: Activity) {
-        makeSnackBar(activity).show()
-    }
+    private fun showTimedOutSnackBar(activity: Activity) = showSnackBar(activity, R.string.engagement_incoming_request_timed_out_message)
 
-    private fun makeSnackBar(activity: Activity): SnackBarDelegate =
-        SnackBarDelegateFactory(
-            activity,
-            R.string.engagement_incoming_request_timed_out_message,
-            localeProvider,
-            themeManager.theme
-        ).createDelegate()
+    private fun showAlreadyInCvSnackBar(activity: Activity) = showSnackBar(activity, R.string.entry_widget_call_visualizer_description)
+
+    private fun showSnackBar(activity: Activity, @StringRes messageRes: Int) = SnackBarDelegateFactory(
+        activity,
+        messageRes,
+        localeProvider,
+        themeManager.theme
+    ).createDelegate().show()
+
 }
--- a/widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/HasOngoingSecureConversationUseCase.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/HasOngoingSecureConversationUseCase.kt
@@ -1,6 +1,5 @@
 package com.glia.widgets.core.secureconversations.domain
 
-import com.glia.widgets.chat.domain.IsAuthenticatedUseCase
 import com.glia.widgets.core.secureconversations.SecureConversationsRepository
 import com.glia.widgets.engagement.State
 import com.glia.widgets.engagement.domain.EngagementStateUseCase
@@ -10,7 +9,6 @@ import io.reactivex.rxjava3.core.Flowable
 
 internal class HasOngoingSecureConversationUseCase(
     private val secureConversationsRepository: SecureConversationsRepository,
-    private val isAuthenticatedUseCase: IsAuthenticatedUseCase,
     private val engagementStateUseCase: EngagementStateUseCase
 ) {
     /**
@@ -22,12 +20,19 @@ internal class HasOngoingSecureConversationUseCase(
             secureConversationsRepository.unreadMessagesCountObservable,
             engagementStateUseCase()
         ) { pendingSecureConversations, unreadMessagesCount, state ->
-            isAuthenticatedUseCase() &&
-                (pendingSecureConversations || unreadMessagesCount > 0 || state is State.TransferredToSecureConversation)
+            !state.isLiveEngagement && !state.isQueueing && (pendingSecureConversations || unreadMessagesCount > 0 || state is State.TransferredToSecureConversation)
         }
 
     operator fun invoke(): Flowable<Boolean> = hasOngoingInteraction.distinctUntilChanged().observeOn(AndroidSchedulers.mainThread())
 
-    operator fun invoke(callback: (Boolean) -> Unit) = invoke().firstOrError().unSafeSubscribe(callback)
+    operator fun invoke(onHasOngoingSecureConversation: () -> Unit, onNoOngoingSecureConversation: () -> Unit = {}) {
+        invoke().firstOrError().unSafeSubscribe { hasOngoing ->
+            if (hasOngoing) {
+                onHasOngoingSecureConversation()
+            } else {
+                onNoOngoingSecureConversation()
+            }
+        }
+    }
 
 }
--- a/widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
@@ -2,10 +2,11 @@ package com.glia.widgets.launcher
 
 import android.content.Context
 import com.glia.androidsdk.Engagement
+import com.glia.widgets.callvisualizer.controller.CallVisualizerContract
 import com.glia.widgets.chat.Intention
 import com.glia.widgets.core.secureconversations.domain.HasOngoingSecureConversationUseCase
-import com.glia.widgets.di.ControllerFactory
 import com.glia.widgets.engagement.domain.EndEngagementUseCase
+import com.glia.widgets.engagement.domain.EngagementTypeUseCase
 import com.glia.widgets.engagement.domain.IsQueueingOrLiveEngagementUseCase
 import com.glia.widgets.helper.Logger
 import com.glia.widgets.helper.TAG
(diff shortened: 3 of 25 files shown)


widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
```diff
@@ -2,50 +2,96 @@ package com.glia.widgets.launcher
...
+    fun startChat(context: Context) = startChat(context, null)
+
     /**
      * Starts a chat engagement.
      *
      * @param context Activity or Context used to launch the chat screen
      * @param visitorContextAssetId Optional visitor context id from Glia Hub
      */
-    fun startChat(context: Context, visitorContextAssetId: String? = null)
+    @FeatureUnderConsiderationForRemoval
```
DavDo, line 45: What do you think about this solution?
As the `@JvmOverloads` annotation is not supported for interfaces, integrators using Java must explicitly pass null for `visitorContextAssetId`. For this reason, I have defined a new function instead.
Later I added this annotation to prevent integrators from using the functions with `visitorContextAssetId` until they need it. To do that, they have to mark their functions with the `@OptIn(FeatureUnderConsiderationForRemoval::class)` annotation, otherwise it won't compile.

widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/HasOngoingSecureConversationUseCase.kt
```diff
@@ -22,12 +20,19 @@ internal class HasOngoingSecureConversationUseCase(
             secureConversationsRepository.unreadMessagesCountObservable,
             engagementStateUseCase()
         ) { pendingSecureConversations, unreadMessagesCount, state ->
-            isAuthenticatedUseCase() &&
-                (pendingSecureConversations || unreadMessagesCount > 0 || state is State.TransferredToSecureConversation)
+            pendingSecureConversations || unreadMessagesCount > 0 || state is State.TransferredToSecureConversation
```
DavDo, line 23: Removed the `isAuthenticatedUseCase()` check, as it was included in the unread_messages_count observable on the core SDK side.

gugalo: I don't see a problem if under the hood as I understood it is still taken into account.

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
gugalo: Since it would be part of 3.0.0 release I think this would be cool to have 👍

gugalo (approved):

Andrii-Horishnii-Glia (approved):

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerActivityWatcher.kt
```diff
@@ -94,15 +95,15 @@ internal class CallVisualizerActivityWatcher(
         activityLauncher.launchWebBrowser(activity, title, url)
     }
 
-    private fun showSnackBar(activity: Activity) {
-        makeSnackBar(activity).show()
-    }
+    private fun showTimedOutSnackBar(activity: Activity) = showSnackBar(activity, R.string.engagement_incoming_request_timed_out_message)
+
+    private fun showAlreadyInCallSnackBar(activity: Activity) = showSnackBar(activity, R.string.entry_widget_call_visualizer_description)
```
andrews-moc, line 100: At the moment this localization fits our needs here. I'm thinking if an integrator might need different wording for an ongoing Call Visualizer and any ongoing engagement when using an Engagement Launcher? 🤔

widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/HasOngoingSecureConversationUseCase.kt
andrews-moc: In my opinion we can keep `isAuthenticatedUseCase()` here because it's directly related to HasOngoing**Secure**ConversationUseCase. Unlike the `unread_messages_count` observable - it's name doesn't say that it's related to Secure Conversations.

widgetssdk/src/main/java/com/glia/widgets/callvisualizer/CallVisualizerActivityWatcher.kt
DavDo: This snack bar is only used in cases of ongoing CVs, so the use case is pretty much the same.
I'll rename `showAlreadyInCallSnackBar(activity: Activity)` -> `showAlreadyInCVSnackBar(activity: Activity)`

widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/HasOngoingSecureConversationUseCase.kt
DavDo: So, in general, we don't need to check for authentication here.

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
andrews-moc: @DavDo I like the idea to have a two separate functions. I was also considering it, but at that time decided to keep a list of interfaces for integrator as simple as possible. I'm just not getting - are we planning to remove the functions that you marked with `FeatureUnderConsiderationForRemoval`? I haven't heard about the plans to remove them and don't see it in the Entry Widget project plan. If so, we should also add the `Deprecated` annotations, shouldn't we?

widgetssdk/src/main/java/com/glia/widgets/view/snackbar/SnackBarDelegate.kt
```diff
@@ -50,10 +51,10 @@ internal abstract class SnackBarDelegate(
 
     fun show() = snackBar.show()
 
-    private fun makeSnackBar(view: View, titleStringKey: Int, stringProvider: LocaleProvider, snackBarTheme: SnackBarTheme?): Snackbar {
+    private fun makeSnackBar(view: View, @StringRes titleStringKey: Int, stringProvider: LocaleProvider, snackBarTheme: SnackBarTheme?): Snackbar {
         val bgColor = snackBarTheme?.backgroundColorTheme?.primaryColor ?: view.getColorCompat(fallbackBackgroundColor)
         val textColor = snackBarTheme?.textColorTheme?.primaryColor ?: view.getColorCompat(fallbackTextColor)
-        val message = stringProvider.getRemoteString(titleStringKey)
+        val message = stringProvider.getString(titleStringKey)
```
andrews-moc, line 57: I think we should even use `localeProvider` from constructor instead of `stringProvider`. 🤔

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
DavDo: >  I was also considering it, but at that time decided to keep a list of interfaces for integrator as simple as possible

All the integrators using v 3.0 must use **EngagementLauncher** or the **EntryWidget**.
Since the `@JvmOverloads` is not supported for interfaces, all the integrators using the **EngagementLauncher** have to explicitly pass `null` for `visitorContextAssetId`, which makes things not so clear for the integrators that don't need this property.
With this solution integrators who will try to use function with the `visitorContextAssetId` must explicitly put @OptIn(FeatureUnderConsiderationForRemoval::class) annotation. Otherwise, it won't compile. So it is almost impossible to use it by mistake.

>  I'm just not getting - are we planning to remove the functions that you marked with FeatureUnderConsiderationForRemoval? I haven't heard about the plans to remove them and don't see it in the Entry Widget project plan. If so, we should also add the Deprecated annotations, shouldn't we?

We haven't decided to deprecate it for removal yet, but we don't encourage the use of it because it's poorly implemented, and we'll remove it at some point. So I wanted to have some way for preventing these functions usage, without deprecating them, as much as it is possible.

I understand that `FeatureUnderConsiderationForRemoval` is not the best name for this annotation, but it’s the best I could come up with. Therefore, I would really appreciate a couple of new name suggestions.

widgetssdk/src/main/java/com/glia/widgets/view/snackbar/SnackBarDelegate.kt
DavDo: image

will rename to `localeProvider`

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
andrews-moc: > We haven't decided to deprecate it for removal yet, but we don't encourage the use of it because it's poorly implemented, and we'll remove it at some point. So I wanted to have some way for preventing these functions usage, without deprecating them, as much as it is possible.

> I understand that FeatureUnderConsiderationForRemoval is not the best name for this annotation, but it’s the best I could come up with. Therefore, I would really appreciate a couple of new name suggestions.

Hm, let me think.

andrews-moc: So, what options do we have overall:
A) Provide only `fun startChat(context: Context, visitorContextAssetId: String? = null)`. Most integrators would need to pass `null` for `visitorContextAssetId`.
B) Provide `fun startChat(context: Context, visitorContextAssetId: String? = null)` and `fun startChat(context: Context)`. Most integrators would use the second one. Only those who need `visitorContextAssetId` will find the another one and use.
C) The same as B + mark with `FeatureUnderConsiderationForRemoval`. Integrators would need to mark with `FeatureUnderConsiderationForRemoval` if they would like to use `visitorContextAssetId`.
D) The same as B + mark function with `visitorContextAssetId` as Deprecated. This is a straightforward and common practice.

I'm not completely sure if we will deprecate `visitorContextAssetId`. So, I doubt if we need to go with C or D.

Anyway, my personal preference (from most to least preferable) is: B, A, D, C.

If you would still prefer to go with C, I can suggest the following names for the annotation: `DeprecatedButUsable`, `RestrictedUsage`, `DiscouragedMethod`, `LegacyFunctionWarning`.

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
```diff
@@ -2,50 +2,96 @@ package com.glia.widgets.launcher
...
+@RequiresOptIn(
+    level = RequiresOptIn.Level.ERROR,
+    message = "This feature is under consideration for removal in a future releases. Use with caution."
+)
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY)
+/**
+ * Marks the feature as under consideration for removal in a future releases.
+ */
+annotation class FeatureUnderConsiderationForRemoval
```
andrews-moc, line 23: If we do decide to use this approach, I would suggest to move this class out of `EngagementLauncher`. It looks like a standalone class. Also, it doesn't look good to ship the brand new `EngagementLauncher` with anything non-recommended or even deprecated, I think.

DavDo: > If we do decide to use this approach, I would suggest to move this class out of EngagementLauncher. It looks like a standalone class. 

Agree!

DavDo: > Also, it doesn't look good to ship the brand new EngagementLauncher with anything non-recommended or even deprecated, I think.

But the `visitorContextAssetId` is actually non-recommended, isn't it?

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
DavDo: 🤔 Huh, you convinced me!
I choose B), but I will modify it a bit to make `visitorContextAssetId` not optional `fun startChat(context: Context, visitorContextAssetId: String)`.

widgetssdk/src/main/java/com/glia/widgets/view/snackbar/SnackBarDelegate.kt
DavDo: fixed - [c170b19](https://github.com/salemove/android-sdk-widgets/pull/1210/commits/c170b190292d0e7ac7895f1dda11e8146baa7b55)

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
DavDo: done - [c170b19](https://github.com/salemove/android-sdk-widgets/pull/1210/commits/c170b190292d0e7ac7895f1dda11e8146baa7b55)

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
andrews-moc: 'Is a poorly-thought feature'. 🙂

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
```diff
@@ -15,37 +16,66 @@ import com.glia.widgets.helper.TAG
...
+     * @param context Activity or Context used to launch the chat screen
+     */
+    fun startChat(context: Context)
+
     /**
      * Starts a chat engagement.
      *
      * @param context Activity or Context used to launch the chat screen
-     * @param visitorContextAssetId Optional visitor context id from Glia Hub
+     * @param visitorContextAssetId Visitor context id from Glia Hub
```
andrews-moc, line 31: NIT. I remember one day tech writer advised to use visitor from a lowercase letter because it's not a name.

andrews-moc: Here and in other places below. 😉

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
andrews-moc: I like how it looks like now. If you still have any concerns regarding deprecating or limiting the `visitorContextAssetId` usage - please reach out to me, we can bring this topic to the product manager. 😉

andrews-moc (approved):
Great job!

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
DavDo: I'm wondering why the `Optional` was in uppercase

DavDo: `Activity` as well

DavDo: image
Maybe this is some kind of beginning of a sentence?
So we should:
a) change both
b) keep both

andrews-moc: Ah, indeed, Optional should have been with lowercase because it's not a name as well. However, I don't remember regarding Activity. If I'm not mistaken, if it's used as an Android component or class name then it should be from the capital letter.

andrews-moc: In my opinion optional should start from lowercase (it's just a regular word) and Activity or Context from a capital letter (because these are the names of Android components) here. IMHO. Not critical, anyway.

DavDo: Hmm, I checked the `AppCompatActivity.java` class from Android library.
image
It seems that the Word next to the `@param parameter-name` is considered the start of a sentence.

andrews-moc: Good to know, thank you. Sorry for confusion with `Visitor`.

PR #1215, opened 2025-01-20
Improve the mark messages as read logic

Andrii-Horishnii-Glia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3954

**What was solved?**
New messages should be marked as read when:

- We have new messages when the history loaded after a visitor opened the chat (transcript) screen or a visitor gets a new SC or Live Engagement message from an Operator; and
- There are no ongoing engagements or an ongoing engagement on_end equals retain; and
- The visitor has stayed on the chat (transcript) screen for at least 6 seconds after the new message; and
- The leave current SC dialog has not shown for at least 6 seconds after the new message.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/ShouldMarkMessagesReadUseCase.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/ShouldMarkMessagesReadUseCase.kt
@@ -1,0 +1,22 @@
+package com.glia.widgets.core.secureconversations.domain
+
+import com.glia.widgets.engagement.EngagementRepository
+
+/**
+ * This use case is responsible for determining if the messages should be marked as read.
+ * The messages should be marked as read if there is no ongoing live engagement or if the current
+ * engagement action on end is retain.
+ *
+ * If result is true, [MarkMessagesReadWithDelayUseCase] should be used to mark the messages as read.
+ * Additional logic is implemented in [MarkMessagesReadWithDelayUseCase] to delay the marking of
+ * messages as read if the chat screen is open or the leave dialog is visible.
+ */
+internal class ShouldMarkMessagesReadUseCase(
+    private val engagementRepository: EngagementRepository
+) {
+
+    operator fun invoke(): Boolean {
+        return !engagementRepository.hasOngoingLiveEngagement
+            || engagementRepository.isRetainAfterEnd
+    }
+}
--- a/widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
@@ -171,6 +171,9 @@ internal class EngagementRepositoryImpl(
     override val isSecureMessagingRequested: Boolean
         get() = _isSecureMessagingRequested
 
+    override val isRetainAfterEnd: Boolean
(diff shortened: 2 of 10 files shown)


DavDo (approved):
LGTM

widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
```diff
@@ -67,7 +67,7 @@ internal class EngagementRepositoryImpl(
     private val engagementStateCallback: Consumer<EngagementState> = Consumer(::handleEngagementState)
     private val engagementEndCallback: Runnable = Runnable(::handleEngagementEnd)
 
-    private var currentEngagement: Engagement? = null
+    override var currentEngagement: Engagement? = null
```
andrews-moc, line 70: I'm a bit concerned if it's okay to extend the visibility because we have access to `currentEngagement` from `GliaCore`. So I would think twice if we really have strong reasons for doing this and cannot avoid.

DavDo: I’d suggest defining a property `EngagementRepository.isRetain: Boolean get() = currentEngagement?.state?.actionOnEnd == Engagement.ActionOnEnd.RETAIN`

Andrii-Horishnii-Glia: fixed 1e0c5714022c9182c9839ab1aa8152dc720d2eb2

widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/ShouldMarkMessagesReadUseCase.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import com.glia.widgets.engagement.EngagementRepository
+
+internal class ShouldMarkMessagesReadUseCase(
+    private val engagementRepository: EngagementRepository
+) {
+
+    operator fun invoke(): Boolean {
+        return !engagementRepository.hasOngoingLiveEngagement
+            || engagementRepository.isRetainAfterEnd
```
andrews-moc, line 11: Discussed with @Andrii-Horishnii-Glia that it would be great to aggregate all the decision logic described in PR description in a single place rather than having two conditions here and the remaining two somewhere else. @Andrii-Horishnii-Glia will try to do it.

DavDo:
@Andrii-Horishnii-Glia, what is the current status of this PR? Should we merge it or close it?

widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/ShouldMarkMessagesReadUseCase.kt
Andrii-Horishnii-Glia: Unfortunately, the logic in one place broke some cases.
The current implementation separates logic into two places: when the mark messages as read should be called and the delay logic. I added comments about it to the UseCase classes.

Andrii-Horishnii-Glia:
@DavDo PR is ready to be merged.

andrews-moc (approved):

DavDo:
> @DavDo PR is ready to be merged.

😎

Andrii-Horishnii-Glia:
!squash

PR #1219, opened 2025-01-23
MOB 3942: Implement engagement 'on_end' functionality

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3942

**What was solved?**

**Release notes:**

 - [x] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
@@ -572,13 +572,16 @@ internal class ChatController(
 
     private fun onEngagementStateChanged(state: State) {
         when (state) {
-            State.FinishedCallVisualizer, State.FinishedOmniCore -> {
+            is State.EngagementEnded -> {
                 if (!isQueueingOrOngoingEngagement) {
                     dialogController.dismissDialogs()
                 }
+                if (state.onEnd == Engagement.ActionOnEnd.RETAIN) {
+                    onTransferredToSecureConversation()
+                }
             }
 
-            State.StartedOmniCore -> newEngagementLoaded()
+            is State.EngagementStarted -> if(!state.isCallVisualizer) newEngagementLoaded()
             is State.Update -> handleEngagementStateUpdate(state.updateState)
             is State.PreQueuing, is State.Queuing -> queueForEngagementStarted()
             is State.QueueUnstaffed, is State.UnexpectedErrorHappened, is State.QueueingCanceled -> emitViewState { chatState.chatUnavailableState() }
--- a/widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
@@ -5,6 +5,7 @@ import android.app.Activity
 import android.content.Intent
 import androidx.annotation.VisibleForTesting
 import com.glia.androidsdk.Engagement
+import com.glia.androidsdk.Engagement.ActionOnEnd
 import com.glia.androidsdk.Engagement.MediaType
 import com.glia.androidsdk.EngagementRequest
 import com.glia.androidsdk.EngagementRequest.Outcome
@@ -196,8 +197,8 @@ internal class EngagementRepositoryImpl(
         cancelQueuing()
     }
 
-    private fun resetState() {
-        _isSecureMessagingRequested = false
+    private fun resetState(retainSecureMessagingState: Boolean = false) {
+        _isSecureMessagingRequested = retainSecureMessagingState
         _operatorMediaState.onNext(Data.Empty)
         _visitorMediaState.onNext(Data.Empty)
         _currentOperator.onNext(Data.Empty)
@@ -213,12 +214,12 @@ internal class EngagementRepositoryImpl(
             currentEngagement = null
             Logger.i(TAG, "Engagement ended locally, silently:$silently")
             unsubscribeFromEvents(it)
-            notifyEngagementEnded(it)
+            notifyEngagementEnded(it, false)
             it.end { ex -> ex?.also { Logger.d(TAG, "Ending engagement failed") } }
             if (silently || it is OmnibrowseEngagement) {
                 _survey.onNext(SurveyState.Empty)
             } else {
-                fetchSurvey(it, false)
+                fetchSurvey(it)
             }
             resetState()
         }
@@ -331,15 +332,13 @@ internal class EngagementRepositoryImpl(
         }
     }
 
-    private fun fetchSurvey(engagement: Engagement, isOperator: Boolean) {
+    private fun fetchSurvey(engagement: Engagement) {
         engagement.getSurvey { survey, _ ->
             when {
                 survey != null -> {
                     Logger.i(TAG, "Survey loaded")
                     _survey.onNext(SurveyState.Value(survey))
                 }
-
-                isOperator -> _survey.onNext(SurveyState.EmptyFromOperatorRequest)
                 else -> _survey.onNext(SurveyState.Empty)
             }
         }
@@ -408,7 +407,7 @@ internal class EngagementRepositoryImpl(
 
             engagement.state.isLiveEngagementTransferredToSecureConversation -> _engagementState.onNext(State.TransferredToSecureConversation)
 
-            else -> _engagementState.onNext(State.StartedOmniCore)
+            else -> _engagementState.onNext(State.EngagementStarted(EngagementType.OmniCore))
         }
 
         currentEngagement = engagement
@@ -428,7 +427,7 @@ internal class EngagementRepositoryImpl(
             unsubscribeFromEvents(it)
             currentEngagement = null
             resetState()
-        } ?: _engagementState.onNext(State.StartedCallVisualizer)
+        } ?: _engagementState.onNext(State.EngagementStarted(EngagementType.CallVisualizer))
 
         currentEngagement = engagement
         handleEngagementState(engagement.state)
@@ -531,23 +530,25 @@ internal class EngagementRepositoryImpl(
         currentEngagement?.also {
             currentEngagement = null
             unsubscribeFromEvents(it)
-            notifyEngagementEnded(it)
-            resetState()
+            notifyEngagementEnded(it, true)
 
-            if (it is OmnicoreEngagement) {
-                fetchSurvey(it, true)
+            resetState(it.state.actionOnEnd == ActionOnEnd.RETAIN)
+
+            if (it is OmnicoreEngagement && it.state.actionOnEnd == ActionOnEnd.SHOW_SURVEY) {
+                fetchSurvey(it)
             } else {
                 _survey.onNext(SurveyState.Empty)
             }
         }
     }
 
-    private fun notifyEngagementEnded(engagement: Engagement) {
-        val state = if (engagement is OmnibrowseEngagement)
-            State.FinishedCallVisualizer
+    private fun notifyEngagementEnded(engagement: Engagement, isEndedByOperator: Boolean) {
+        val engagementType = if (engagement is OmnibrowseEngagement)
+            EngagementType.CallVisualizer
         else
-            State.FinishedOmniCore
+            EngagementType.OmniCore
 
+        val state = State.EngagementEnded(engagementType, !isEndedByOperator, engagement.state.actionOnEnd)
         _engagementState.onNext(state)
     }
 
--- a/widgetssdk/src/main/java/com/glia/widgets/engagement/States.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/engagement/States.kt
@@ -1,5 +1,6 @@
 package com.glia.widgets.engagement
 
+import com.glia.androidsdk.Engagement.ActionOnEnd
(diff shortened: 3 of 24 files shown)


igorkravchenko (approved):

gugalo:
!fixup

gugalo:
!squash

gugalo:
CI is failing because the last Core SDK staging build expired again.

Andrii-Horishnii-Glia (approved):
LGTM

widgetssdk/src/main/java/com/glia/widgets/chat/controller/ChatController.kt
```diff
@@ -572,13 +572,16 @@ internal class ChatController(
...
                 if (!isQueueingOrOngoingEngagement) {
                     dialogController.dismissDialogs()
                 }
+                if (state.onEnd == Engagement.ActionOnEnd.RETAIN) {
+                    onTransferredToSecureConversation()
+                }
             }
 
-            State.StartedOmniCore -> newEngagementLoaded()
+            is State.EngagementStarted -> newEngagementLoaded()
```
andrews-moc, line 584: To double check, previously `newEngagementLoaded()` was called only for OmniCore. Now it will be called for Call Visualizer as well. Is this an intended change?

widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
```diff
@@ -531,23 +530,25 @@ internal class EngagementRepositoryImpl(
...
-        val state = if (engagement is OmnibrowseEngagement)
-            State.FinishedCallVisualizer
+    private fun notifyEngagementEnded(engagement: Engagement, isEndedByOperator: Boolean) {
+        val engagementType = if (engagement is OmnibrowseEngagement)
+            EngagementType.CallVisualizer
         else
-            State.FinishedOmniCore
+            EngagementType.OmniCore
 
+        val state = State.EngagementEnded(engagementType, !isEndedByOperator, engagement.state.actionOnEnd)
```
andrews-moc, line 551: I'd like to suggest changing the `isEndedByVisitor` param in `State.EngagementEnded` to `isEndedByOperator`. By doing this, we will deal with the same logical entity tight to operator. I think it will improve readability, simplify the code and drop the negation in the `!isEndedByOperator` like here we see.

widgetssdk/src/main/java/com/glia/widgets/engagement/States.kt
```diff
@@ -12,19 +12,17 @@ internal sealed interface State {
...
     data object QueueingCanceled : State
-    data object StartedOmniCore : State
-    data object StartedCallVisualizer : State
-    data object FinishedOmniCore : State
-    data object FinishedCallVisualizer : State
+    data class EngagementStarted(val type: EngagementType) : EngagementState(type)
+    data class EngagementEnded(val type: EngagementType, val isEndedByVisitor: Boolean, val onEnd: ActionOnEnd) : EngagementState(type)
     data object TransferredToSecureConversation : State
-    data class Update(val state: EngagementState, val updateState: EngagementUpdateState) : State
+    data class Update(val state: com.glia.androidsdk.engagement.EngagementState, val updateState: EngagementUpdateState) : State
```
andrews-moc, line 18: Have you dropped the `import com.glia.androidsdk.engagement.EngagementState` and added `com.glia.androidsdk.engagement.` here to the code on purpose?

widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionActivityWatcher.kt
```diff
@@ -28,16 +29,13 @@ internal class EngagementCompletionActivityWatcher(
...
-
-            /*
-            * This state finishes all SDK activities even if the application is in the background(activity == null),
-            * to be sure that everything related to the queueing or engagement is released.
-            * */
-            state is EngagementCompletionState.QueuingOrEngagementEnded -> event.consume { finishActivities() }
-            activity == null || activity.isFinishing -> Logger.d(TAG, "skipping.. activity is null or finishing")
+            activity == null || activity.isFinishing -> {
+                Logger.d(TAG, "skipping.. activity is null or finishing")
+                if (state is EngagementCompletionState.EngagementEnded) event.consume { finishActivities() }
```
andrews-moc, line 34: Let me also double check here because I don't know the reason for removing the comment and making changes here. Before the change we were checking for QueuingOrEngagementEnded. After the change we check for Engagement only (not interested in queueing anymore). Is this intended?

widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionActivityWatcher.kt
```diff
@@ -73,4 +71,23 @@ internal class EngagementCompletionActivityWatcher(
...
+                showOperatorEndedEngagementDialog(activity, consumeCallback)
+            }
+            Engagement.ActionOnEnd.RETAIN -> {
+                consumeCallback()
+            }
+            Engagement.ActionOnEnd.SHOW_SURVEY -> { // The survey showing is controlled from separate event
+                consumeCallback()
+            }
+        }
+    }
```
andrews-moc, line 92: It seems to me that this block can be refactored.
1. I'm a bit confused by the `consumeCallback()` name. Can you think of a more meaningful name based on the code that is expected in a callback? Do I understand correctly that it's only for marking as consumed? Maybe `markConsumedCallback` or `markConsumed` name will be more meaningful?
2. consumeCallback() is called from almost all cases. Maybe it makes sense to unite these cases or even check only the condition when it shouldn't be called and call consumeCallback() for all other cases?
3. I still don't understand why survey is mentioned in the comments several times, there is even a SHOW_SURVEY event but there is no code that is triggering the survey here. Maybe you can think of a more specific name for the function instead of a generic `handleEngagementEndedEvent()`, for example `showOperatorEndedEngagementDialogOrMarkConsumed` or something shorted?

widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionController.kt
```diff
@@ -37,20 +35,18 @@ internal class EngagementCompletionController(
 
     private fun handleEngagementState(state: State) {
         when (state) {
-            State.FinishedCallVisualizer, State.FinishedOmniCore -> {
-                releaseResourcesUseCase()
-                _state.onNext(EngagementCompletionState.QueuingOrEngagementEnded)
+            is State.EngagementEnded -> {
+                if (state.onEnd != Engagement.ActionOnEnd.RETAIN) releaseResourcesUseCase()
```
andrews-moc, line 39: It looks a bit confusing to me. I would expect to see that we keep resources for `RETAIN` event, based on it's name.

widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionController.kt
```diff
@@ -37,20 +35,18 @@ internal class EngagementCompletionController(
...
+                _state.onNext(EngagementCompletionState.EngagementEnded(state.isEndedByVisitor, state.onEnd))
```
andrews-moc, line 40: As I already mentioned, ideally, it might be better to use identical logical entity everywhere - either `isEndedByVisitor` or `isEndedByOperator`, but not both.

andrews-moc (commented):
Also, taking into account that new business logic is introduced, I think it makes sense to make sure that new logic is covered by existing tests. Maybe a separate ticket is needed.

widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
gugalo: Every data model should correctly represent data that matters. And in this class, we only care whether the engagement has been ended by the visitor or not. For example, we don't care if it was ended by the operator or ended because of some error. We only care if it was an explicit action by the visitor to end the engagement because in this case, we don't need to show any alert dialogs.

Overall I agree that it is subjective depending on which way you look at it. And your argument also makes sense to me but I would prefer to leave it this way unless you have another argument )

widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionActivityWatcher.kt
gugalo: > I don't know the reason for removing the comment
I have re-read it multiple times only to understand that it is not really useful here. It literally explained what is obvious from one line code with some additional details that are not really helpful

gugalo: > Before the change we were checking for QueuingOrEngagementEnded. After the change we check for Engagement only (not interested in queueing anymore). Is this intended?

This is a bit harder to explain as `QueuingOrEngagementEnded` is removed from the list and variation of `event.consume { finishActivities() }` is called in many different cases. I have tested this logic quite a lot and have not detected any problems with new solution

widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
andrews-moc: Okay. I will not insist in this case.

widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionActivityWatcher.kt
gugalo: > 1. I'm a bit confused by the consumeCallback() name. Can you think of a more meaningful name based on the code that is expected in a callback?

👌 

> 2. consumeCallback() is called from almost all cases. Maybe it makes sense to unite these cases or even check only the condition when it shouldn't be called and call consumeCallback() for all other cases?

If it had been just my new logic that you see in the lines above then I would have done just that. But the problem is the `showOperatorEndedEngagementDialog` which triggered the callback only after an alert dialog button click is performed. I need to pass the callback there. Or maybe I am missing what you exactly suggest. If you still have some ideas please share code snippets. 

> 3. I still don't understand why survey is mentioned in the comments several times, there is even a SHOW_SURVEY event but there is no code that is triggering the survey here. Maybe you can think of a more specific name for the function instead of a generic handleEngagementEndedEvent(), for example showOperatorEndedEngagementDialogOrMarkConsumed or something shorted?

I am confused regarding this suggestion. I would like to discuss it live

widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionController.kt
gugalo: Good point. I will investigate this a bit more

widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionActivityWatcher.kt
andrews-moc: I should have suggested refactored code instead of just commenting, sorry. So, how about this:
```
private fun EngagementCompletionState.EngagementEnded.handleEngagementEndedEvent(
        activity: Activity,
        onEventHandled: () -> Unit
    ) {
        if (isEndedByVisitor || actionOnEnd == Engagement.ActionOnEnd.RETAIN || actionOnEnd == Engagement.ActionOnEnd.SHOW_SURVEY) {
            onEventHandled()
        } else if (actionOnEnd == Engagement.ActionOnEnd.END_NOTIFICATION || actionOnEnd == Engagement.ActionOnEnd.UNKNOWN) {
            showOperatorEndedEngagementDialog(activity, onEventHandled)
        }
    }
```
I tried to collect all conditions for a `onEventHandled()` call and all conditions for `showOperatorEndedEngagementDialog(activity, onEventHandled)`. What do you think?

We can discuss it live, sure, just let me know.

widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionController.kt
gugalo: I double checked the `releaseResourcesUseCase` being called on `ActionOnEnd.RETAIN` and yes, it is expected. We want to reset the Glia chat view and related controllers to letter setup them again into new correct states.

widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionActivityWatcher.kt
gugalo: Sounds good 👍

widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionController.kt
andrews-moc: I see. Thank you for checking. I think it might be useful to add this info as a comment to the code. IMHO.

andrews-moc (approved):

gugalo:
!squash

PR #1220, opened 2025-01-27
Add chat input disabled state

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3891

**What was solved?**
After messing up with different `Drawable`s and finding out that it is really hard to copy the current state from the `StateListDrawable`, I decided to go with a safer solution and used the `StatefulWidgetAdapter` that I had previously created for the SC Welcome screen. Combining it with the default state, defined with the `UnifiedUi`, gives us the flexibility to use the default state theme when the theme for that state is absent in clients UnifiedUi.

**Release notes:**

 - [x] Feature
 - [x] Ignore(Part of SC V2)

**Additional info:**

**Screenshots:**


widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/UnifiedUiExtensions.kt
```diff
@@ -51,58 +51,71 @@ internal fun createBackgroundFromTheme(color: ColorTheme): Drawable = color.run
...
+        else -> GradientDrawable()
+    }
+
 /**
  * will update whole background without keeping an old background or stroke
  * in case when the old background differs from [GradientDrawable]
  */
-internal fun View.applyLayerTheme(defTheme: LayerTheme?, disabledTheme: LayerTheme? = null) {
-    if (defTheme?.fill == null && defTheme?.stroke == null) return
+internal fun View.applyLayerTheme(layerTheme: LayerTheme?) {
```
DavDo, line 68: This change resulted in the 'snapshots' modifications, as we don't need `StateListDrawable' for views that have only one default state.

widgetssdk/src/test/snapshots/images/com.glia.widgets.chat_ChatViewSnapshotTest_secureMessagingUnavailableWithGlobalColors.png
andrews-moc, line 1: Hm, I'm wondering why input background color has been changed here, but stayed the same for other cases?

andrews-moc: UPD. Has been changed not only here. In some cases it has been changed, in others - no. I would expect the input background color to be the same as SC bottom banner background color in all cases, am I wrong?

andrews-moc (commented):
Isn't it better to replace snapshot files only for the cases that contain changes to avoid missing some undesired changes?

widgetssdk/src/main/java/com/glia/widgets/view/unifiedui/UnifiedUiExtensions.kt
andrews-moc: Isn't it better to replace snapshot files only for the cases that contain changes to avoid missing some undesired changes? Or you mean that it's not possible in this PR?

widgetssdk/src/test/snapshots/images/com.glia.widgets.chat_ChatViewSnapshotTest_secureMessagingUnavailableWithGlobalColors.png
DavDo, line 1: This is expected because the input is disabled when the SC is unavailable.

DavDo: Actually, it should change for the cases when SC is unavailable since we added that state.
Input color should be the same with:

- Chat background color - Enabled
- SC Bottom Banner background color - Disabled

Unless it is not unified ui customization without global colors.

widgetssdk/src/main/java/com/glia/widgets/chat/StatefulChatInputBackground.kt
```diff
@@ -0,0 +1,62 @@
...
+import com.glia.widgets.helper.getAttr
+import com.glia.widgets.helper.getColorCompat
+import com.glia.widgets.messagecenter.SimpleStatefulWidgetAdapter
+import com.glia.widgets.messagecenter.StatefulWidgetAdapter
+import com.glia.widgets.messagecenter.StatefulWidgetAdapterCallback
+import com.glia.widgets.view.unifiedui.applyLayerTheme
+import com.glia.widgets.view.unifiedui.theme.base.ColorTheme
+import com.glia.widgets.view.unifiedui.theme.base.LayerTheme
+
+internal class StatefulChatInputBackground @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null) : View(context, attrs) {
```
gugalo, line 18: I would appreciate some comment on why this was added for future reference or you may shortly summarize what this class tries to achive

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/StatefulChatInputBackground.kt
```diff
@@ -0,0 +1,62 @@
...
+        StatefulWidgetAdapterCallback { applyLayerTheme(it) }
+
+    private val statefulWidgetAdapter: StatefulWidgetAdapter<State, LayerTheme> by lazy {
+        SimpleStatefulWidgetAdapter(State.ENABLED, adapterCallback)
+    }
+
+    @get:ColorInt
+    private val gliaSystemAgentBubbleColor: Int by lazy {
+        getColorCompat(getAttr(R.attr.gliaSystemAgentBubbleColor, R.color.glia_neutral_color))
+    }
```
gugalo, line 30: Could you maybe make `gliaSystemAgentBubbleColor` as a `val disabledColor` value assign at class `init { }`. I mean this is a question of preference so it is up to you if you want to leave it as is.

gugalo (approved):

gugalo:
Not to block this PR you can resolve both of my comments when you want even if you decide to leave it as is.

Andrii-Horishnii-Glia (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/StatefulChatInputBackground.kt
DavDo: fixed - eccc1faaebbf293da3ecb4927c4fdf3a78aac252

widgetssdk/src/main/java/com/glia/widgets/chat/StatefulChatInputBackground.kt
DavDo: fixed - eccc1faaebbf293da3ecb4927c4fdf3a78aac252

DavDo:
!squash

PR #1221, opened 2025-01-27
Feature/mob 3969 adjust engagement launcher logic as per diagram

andrews-moc:
**Jira issue:**
[[Android] Devs want to adjust the EngagementLauncher logic according to the diagram](https://glia.atlassian.net/browse/MOB-3969)

**What was solved?**
1. Adjusted the EngagementLauncher logic according to the [diagram](https://www.figma.com/board/XAwLSPaSzcaIv6BLTdgzXc/Mobile-Secure-Conversations-Flows?node-id=496-1078&t=PJqOvL8a8xQA6R5t-1).
2. Extracted Snackbar functionality from Call Visualizer to an app level. Decided to leave current Snackbar usages intact.
3. Moved Live Observation snackbar classes to a separate sub-directory of `snackbar` package (separate commit for easier review).
4. Revisited the unit tests (separate commit for easier review).

**Release notes:**

 - [x] Feature
 - [x] Ignore

**Additional info:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
@@ -2,14 +2,12 @@ package com.glia.widgets.launcher
 
 import android.content.Context
 import com.glia.androidsdk.Engagement
-import com.glia.widgets.callvisualizer.controller.CallVisualizerContract
+import com.glia.widgets.R
 import com.glia.widgets.chat.Intention
 import com.glia.widgets.core.secureconversations.domain.HasOngoingSecureConversationUseCase
-import com.glia.widgets.engagement.domain.EndEngagementUseCase
 import com.glia.widgets.engagement.domain.EngagementTypeUseCase
 import com.glia.widgets.engagement.domain.IsQueueingOrLiveEngagementUseCase
-import com.glia.widgets.helper.Logger
-import com.glia.widgets.helper.TAG
+import com.glia.widgets.view.snackbar.SnackbarContract
 
 /**
  * An interface for launching different types of engagements, such as chat,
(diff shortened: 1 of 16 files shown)


widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
```diff
@@ -80,115 +79,142 @@ interface EngagementLauncher {
...
+    private val isChatEngagementOrChatQueueing: Boolean
+        get() = engagementTypeUseCase.isChatEngagement || isQueueingOrLiveEngagementUseCase.isQueueingForLiveChat
+
+    override fun startChat(context: Context) {
         when {
             engagementTypeUseCase.isCallVisualizer -> callVisualizerController.showAlreadyInCvSnackBar()
-            isQueueingOrLiveEngagementUseCase.hasOngoingLiveEngagement || isQueueingOrLiveEngagementUseCase.isQueueingForLiveChat ->
-                activityLauncher.launchChat(context, Intention.RETURN_TO_CHAT)
-
+            isMediaEngagementOrMediaQueueing -> snackbarController.showSnackBar(R.string.entry_widget_call_visualizer_description)
```
DavDo, line 111: I think we should have different texts for `ongoing engagement` and `enqueueing` states.
This is true for each function inside `EngagementLauncher`.

andrews-moc: For the record, PM approved going with the current text.

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
```diff
@@ -80,115 +79,142 @@ interface EngagementLauncher {
...
+
+    /**
+     * @return true if there is an ongoing engagement or enqueueing for live chat
+     */
+    private val isChatEngagementOrChatQueueing: Boolean
+        get() = engagementTypeUseCase.isChatEngagement || isQueueingOrLiveEngagementUseCase.isQueueingForLiveChat
+
+    override fun startChat(context: Context) {
         when {
             engagementTypeUseCase.isCallVisualizer -> callVisualizerController.showAlreadyInCvSnackBar()
```
DavDo, line 110: `callVisualizerController.showAlreadyInCvSnackBar()` uses the same `R.string.entry_widget_call_visualizer_description` string resource under the hood, maybe it makes sense then just remove this function, and use `snackbarController.showSnackBar(R.string.entry_widget_call_visualizer_description)` everywhere.

andrews-moc: Good point. I also thought about it but actually thought it would be much more difficult, but it's not. Fixed: 48c14f0b9b99dad7d0b652dde9fbd8cfeef52b83

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
```diff
@@ -80,115 +78,141 @@ interface EngagementLauncher {
...
 
     override fun startChat(context: Context, visitorContextAssetId: String) {
         configurationManager.setVisitorContextAssetId(visitorContextAssetId)
-
         startChat(context)
     }
 
     override fun startAudioCall(context: Context) {
-        if (isQueueingOrLiveEngagementUseCase.isQueueingForLiveChat || isQueueingOrLiveEngagementUseCase.isQueueingForVideo) {
-            Logger.i(TAG, "Canceling ongoing queue ticket to create a new one")
```
gugalo, line 120: This log and a similar one in `fun startChat()` & `fun startVideoCall()` methods have been completely removed. Is this intended? If yes then please check whether [this doc](https://docs.google.com/spreadsheets/d/1ty9bhrbX3VTXBThI5f4GFzy5X-RbEHOCPWfXd7nGzbo/edit?gid=0#gid=0) needs updating.

gugalo (approved):

widgetssdk/src/test/java/com/glia/widgets/view/snackbar/LiveObservationPopupUseCaseTest.kt
```diff
@@ -4,6 +4,7 @@ import com.glia.androidsdk.GliaException
 import com.glia.androidsdk.RequestCallback
 import com.glia.androidsdk.site.SiteInfo
 import com.glia.widgets.chat.domain.SiteInfoUseCase
+import com.glia.widgets.view.snackbar.liveobservation.LiveObservationPopupUseCase
```
Andrii-Horishnii-Glia, line 7: Better move the file to the same package.

Andrii-Horishnii-Glia (approved):
LGTM

widgetssdk/src/main/java/com/glia/widgets/launcher/EngagementLauncher.kt
andrews-moc: Good point. Thank you for a good catch. Yes, it's intended. I have asked the iOS team to confirm that they also don't have this log now. Will remove it from the doc once they confirm.

widgetssdk/src/test/java/com/glia/widgets/view/snackbar/LiveObservationPopupUseCaseTest.kt
andrews-moc: Sorry, I'm not following. Could you please explain what do you mean? What file to which package are you suggesting to move?

Andrii-Horishnii-Glia: You moved `LiveObservationPopupUseCase.kt` from `com.glia.widgets.view.snackbar` to `com.glia.widgets.view.snackbar.liveobservation`. So, do the same for `LiveObservationPopupUseCaseTest.kt`.

andrews-moc: Ah, got it now. Good point. I moved `LiveObservationPopupUseCaseTest.kt` to `com.glia.widgets.view.snackbar.liveobservation` now.

PR #1223, opened 2025-01-29
Fix chat screen is getting stuck after Visitor ends engagement

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3975

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionActivityWatcher.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionActivityWatcher.kt
@@ -76,7 +76,10 @@ internal class EngagementCompletionActivityWatcher(
         if (actionOnEnd == Engagement.ActionOnEnd.UNKNOWN) {
             Logger.w(TAG, "Engagement ended with unknown case.")
         }
-        if (isEndedByVisitor || actionOnEnd == Engagement.ActionOnEnd.RETAIN || actionOnEnd == Engagement.ActionOnEnd.SHOW_SURVEY) {
+        if (isEndedByVisitor) {
+            finishActivities()
+            onEventHandled()
+        } else if (actionOnEnd == Engagement.ActionOnEnd.RETAIN || actionOnEnd == Engagement.ActionOnEnd.SHOW_SURVEY) {
             onEventHandled()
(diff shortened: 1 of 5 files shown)


widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionActivityWatcher.kt
```diff
@@ -76,7 +76,10 @@ internal class EngagementCompletionActivityWatcher(
         if (actionOnEnd == Engagement.ActionOnEnd.UNKNOWN) {
             Logger.w(TAG, "Engagement ended with unknown case.")
         }
-        if (isEndedByVisitor || actionOnEnd == Engagement.ActionOnEnd.RETAIN || actionOnEnd == Engagement.ActionOnEnd.SHOW_SURVEY) {
+        if (isEndedByVisitor) {
```
DavDo, line 79: Could you please replace this `if-else` chain with `when`

DavDo (approved):

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionActivityWatcher.kt
gugalo: Is there a difference in readability or performance?

DavDo: I think `when` is more readable.

DavDo (approved):

andrews-moc (approved):

widgetssdk/src/main/java/com/glia/widgets/engagement/completion/EngagementCompletionActivityWatcher.kt
gugalo: Converted to `when` but I think the way I did it it is hard to say for sure, at least for me, if readability improved. Could you check this change and see if there are any improvements that I missed? Otherwise, right now I think it is a question of preference.

DavDo: Looks better to me. Thanks 👍.

PR #1226, opened 2025-02-03
Fix Application chat bubble remains after CV engagement end

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-4009

**What was solved?**

**Release notes:**

 - [x] Feature(SC v2)
 - [x] Ignore

**Additional info:**

**Screenshots:**


DavDo:
> So, there was something wrong with removing `ChatHeadLayout`? Out of interest, what has changed so that this part stopped working?

Seems so!
So the problem was only happening with the chat head layout of the application when we add the chat bubble to different parent views.
While debugging, I found out that when we call `gliaRootView.removeView(chatHeadLayout)`, it doesn't remove the bubble, which means that this view is actually not the parent of the bubble.
It turns out that `fetchGliaOrRootView()` returns an inappropriate view in this case.

PR #1254, opened 2025-03-21
Remove duplicate Authentication creation from the example app

andrews-moc:
**What was solved?**
- A tiny thing that I noticed - there's an extra `Authentication` creation right below the same code in the example. I believe it's a mistake.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**


Andrii-Horishnii-Glia (approved):

app/src/main/java/com/glia/exampleapp/MainFragment.kt
```diff
@@ -515,8 +515,6 @@ class MainFragment : Fragment() {
     private fun prepareAuthentication() {
         val sharedPreferences = PreferenceManager.getDefaultSharedPreferences(requireContext())
         authentication = GliaWidgets.getAuthentication(getAuthenticationBehaviorFromPrefs(sharedPreferences, resources))
-        authentication =
```
Andrii-Horishnii-Glia, line 518: The second time to be sure :)

PR #1255, opened 2025-03-31
Add Widgets public classes for Core SDK classes: GliaException and RequestCallback

andrews-moc:
**Jira issue:**
[Add public class for com.glia.androidsdk.GliaException to Android Widgets SDK](https://glia.atlassian.net/browse/MOB-4213)

**What was solved?**
- Add public class for `com.glia.androidsdk.GliaException`
- Add public class for `com.glia.androidsdk.RequestCallback`

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**


Andrii-Horishnii-Glia (commented):
As I see, it means a lot of changes on the integrators' side. Can we just use com.glia.androidsdk.GliaException package and class name in the Widgets SDK?

andrews-moc:
> As I see, it means a lot of changes on the integrators' side. Can we just use com.glia.androidsdk.GliaException package and class name in the Widgets SDK?

@Andrii-Horishnii-Glia Thank you for having a look. Could you please explain in more details what exactly do you mean?

DavDo (changes requested):
I don't think we need to map `GliaException` to `GliaWidgetsException` everywhere.
As far as I understood from the PR, we're mapping them only for internal usage, but in reality we need this mapping only in public interfaces most probably only for `GliaWidgets.onAppCreate()`(if we keep it after @Andrii-Horishnii-Glia's changes, and also for `GliaWidgets.init` function.
For `GliaWidgets.init` function I didn't get it how is the mapping happening, most probably we need to wrap core `Glia.init` call with try/catch and rethrow mapped exception.
I'd suggest doing the same for `RequestCallback`, I'd suggest introducing more client friendly solution for the public functions that require callback like fetching queues and updating visitor info. We're free to introduce new callback and map the core's one inside Widgets.

DavDo:
I would suggest making Core internal in Widgets, and then follow the errors in testing app.

andrews-moc:
> I would suggest making Core internal in Widgets, and then follow the errors in testing app.

Yes, that's what I'm doing locally. However, it's too early to push this (making Core internal in Widgets) to a branch because a lot of places will be broken and won't compile.

andrews-moc:
@DavDo @Andrii-Horishnii-Glia I have taken into account your reasonable feadback and made appropriate changes.
So now I have:
- removed the mapping of `GliaException` to `GliaWidgetsException` in the internal classes
- wrapped the Core SDK calls with try/catch and rethrew mapped `GliaWidgetsException`
- changed the `RequestCallback` interface

Looking forward to seeing your feedback.

widgetssdk/src/main/java/com/glia/widgets/RequestCallback.kt
```diff
@@ -0,0 +1,24 @@
...
+     * @param response result returned if the request succeeded
+     */
+    fun onSuccess(response: T?)
+
+    /**
+     * Function that is fired if request fails
+     *
+     * @param exception error returned if the request failed
+     */
+    fun onError(exception: GliaWidgetsException?)
```
DavDo, line 22: Why are the `response` and `exception` parameters are nullable?
Do we have cases where the successful response is empty?
If so I'd prefer having separate onSuccess Callback for thouse particular cases.

Andrii-Horishnii-Glia (approved):
Looks good

widgetssdk/src/main/java/com/glia/widgets/RequestCallback.kt
DavDo: I think we can have separate callbacks for success and error. For example
```kotlin
typealias OnSuccess<T> = (T) -> Unit
typealias OnError = (Throwable?) -> Unit
typealias OnSuccessCallback = () -> Unit
```

DavDo (approved):
LGTM
Added one suggestion.

widgetssdk/src/main/java/com/glia/widgets/RequestCallback.kt
andrews-moc: @DavDo Maybe I'm missing something, but `typealias` name can directly be referenced only from Kotlin file, isn't it? We need it to work from Java classes as well. 

So I have created a separate `callbacks` package and put there three interfaces:
1. `OnWidgetsResult<T>`
2. `OnWidgetsSuccess`
3. `OnWidgetsError`

You can see how it looks like for get and update visitor info [here](https://github.com/salemove/android-sdk-widgets/pull/1257/files).

Looking forward to seeing your feedback.

DavDo: You're right!
TypeAliases are not directly accessible from java code, and if we pass it as a function parameter it will be converted to `Function0` or smth similar type, so having separate interfaces will look a way readable.

gugalo (commented):
The PR title says "Add public class for GliaException and RequestCallback" however it seems confusing to me. 

Do I understand that instead of `RequestCallback` we now have these three interfaces:
```
OnWidgetsError
OnWidgetsResult
OnWidgetsSuccess
``` 

P.S: I don't have anything against it. Just clarifying that this is intented

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -264,6 +305,11 @@ private static void setupRxErrorHandler() {
         });
     }
 
+    private static RuntimeException mapCoreExceptionToWidgets(GliaException gliaException) {
+        GliaWidgetsException widgetsException = GliaWidgetsException.Companion.from(gliaException);
```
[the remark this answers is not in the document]
andrews-moc: Yes, indeed. Thank you. Fixed. 68067f226a6e4e6544df17402f4f42af47e2836c

andrews-moc:
> The PR title says "Add public class for GliaException and RequestCallback" however it seems confusing to me.
> 
> Do I understand that instead of `RequestCallback` we now have these three interfaces:
> 
> ```
> OnWidgetsError
> OnWidgetsResult
> OnWidgetsSuccess
> ```
> 
> P.S: I don't have anything against it. Just clarifying that this is intented

Yes, that's correct.
To avoid the confusion, I've just renamed PR to 'Add Widgets public classes for Core SDK classes: GliaException and RequestCallback'.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsException.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.glia.androidsdk.GliaException as GliaCoreException
+
+/**
+ * Glia Widgets SDK exception that contains a message for debugging and
+ * a {@link Cause} enum value for handling some of the errors programmatically.
+ */
+class GliaWidgetsException(
+    val debugMessage: String?,
+    val gliaCause: Cause?
+) : RuntimeException(
```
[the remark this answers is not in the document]
andrews-moc: Done a6ee8fe588448cf46ad28a26957e357155732974

widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsException.kt
```diff
@@ -0,0 +1,52 @@
...
+    debugMessage
+) {
+
+    /**
+     * Possible exception causes
+     */
+    enum class Cause(val sdkCause: GliaCoreException.Cause?) {
```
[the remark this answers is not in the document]
andrews-moc: @gugalo I agree with this point.
Changed, please have a look: 62017229c64c966917f0cebfc1fdc4f1e5044404
I've also added unit tests.

gugalo (approved):

PR #1256, opened 2025-04-01
Do not call Core SDK onAppCreate() method

Andrii-Horishnii-Glia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-4198

**What was solved?**
Widgets SDK doesn't call Core SDK `.onAppCreate()` method.
Added the `InitializationProvider`. The provider can be used to call `GliaWidgets.onAppCreate()`. A custom application class in the application or library is not necessary anymore.

**Release notes:**

 - [x] Feature
 - [?] Ignore
 - [?] Release notes (Is it clear from the description here?)
 - [?] Migration guide (If changes are needed for integrator already using the SDK - what needs to be communicated? Add underneath please)

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsConfig.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsConfig.kt
@@ -47,9 +47,14 @@ class GliaWidgetsConfig private constructor(builder: Builder) {
     val manualLocaleOverride: String?
 
     init {
+        // The Context can be an Activity that has its life cycle.
+        // To prevent storing the link to an Activity that can be destroyed, we need to take the Application Context.
+        // An Application object lives all the time when the application is active, so it is safe to store it.
+        val applicationContext = builder.context?.applicationContext
+
         siteApiKey = builder.siteApiKey
         siteId = builder.siteId
-        context = builder.context
+        context = applicationContext
         region = builder.region
         baseDomain = builder.baseDomain
         requestCode = builder.requestCode
--- a/widgetssdk/src/main/java/com/glia/widgets/InitializationProvider.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/InitializationProvider.kt
@@ -1,0 +1,56 @@
+package com.glia.widgets
+
+import android.app.Application
+import android.content.ContentProvider
+import android.content.ContentValues
+import android.database.Cursor
+import android.net.Uri
+
+/**
+ * @hide
+ */
+open class InitializationProvider : ContentProvider() {
+    override fun onCreate(): Boolean {
+        (context as? Application)?.let { GliaWidgets.onAppCreate(it) }
+        return true
+    }
+
+    override fun query(
+        uri: Uri,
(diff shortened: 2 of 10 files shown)


widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsConfig.kt
```diff
@@ -49,7 +49,7 @@ class GliaWidgetsConfig private constructor(builder: Builder) {
     init {
         siteApiKey = builder.siteApiKey
         siteId = builder.siteId
-        context = builder.context
+        context = builder.context?.applicationContext // remove link to the possibly lifecycle-aware component
```
andrews-moc, line 52: Could you please explain the "remove link to the possibly lifecycle-aware component" comment? What does this comment mean? What "lifecycle-aware component" do you mean?

widgetssdk/src/main/java/com/glia/widgets/InitializationProvider.kt
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    override fun delete(p0: Uri, p1: String?, p2: Array<out String>?): Int {
+        return 0
+    }
+
+    override fun update(p0: Uri, p1: ContentValues?, p2: String?, p3: Array<out String>?): Int {
+        return 0
+    }
+}
```
andrews-moc, line 43: Not critical, but it might still be better to have descriptive param names instead of `p0`, `p1`, etc.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsConfig.kt
Andrii-Horishnii-Glia: Even if we ask integrators to use an application context, 
```
        /**
         * @param context - your application context
         * @return Builder instance
         */
        @NonNull
        public Builder setContext(@Nullable Context context) { ... }
```

the context can appear as an activity that has its life cycle. We don't want to store the link to an activity that has already been destroyed. In this case, we can get the application context by ourselves. An application object lives all the time when the application is active, so it is safety to store it.

Andrii-Horishnii-Glia: Fixed 2ad858b77e2ba3d6db45ffda8ba68104f4715341

DavDo:
What was the final descision regarding this solution?

Andrii-Horishnii-Glia:
@DavDo I suggest testing the `InitializationProvider` well first and only then marking the `GliaWidgets.onAppCreate()` as deprecated.

DavDo:
> @DavDo I suggest testing the `InitializationProvider` well first and only then marking the `GliaWidgets.onAppCreate()` as deprecated.

Okay, so we will go with this solution right? And at leas we'll have it on ATs.

Andrii-Horishnii-Glia:
@DavDo Yes, we will try this solution.

Andrii-Horishnii-Glia:
!squash

PR #1257, opened 2025-04-01
Add getVisitorInfo() and updateVisitorInfo() to Android Widgets SDK

andrews-moc:
**Jira issue:**
[Add getVisitorInfo() to Android Widgets SDK](https://glia.atlassian.net/browse/MOB-4203)
[Add updateVisitorInfo() to Android Widgets SDK](https://glia.atlassian.net/browse/MOB-4204)

**What was solved?**

1. Added public `GliaWidgets.getVisitorInfo(callback)`
2. Added public `GliaWidgets.updateVisitorInfo(visitorInfo)`
3. Added public `VisitorInfo` class
4. Added public `VisitorInfoUpdateRequest` class
5. Integrators who use `Glia.getVisitorInfo()` and `Glia.updateVisitorInfo()` Core SDK interfaces can still use them and do not need to change anything until we hide Core SDK.

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/VisitorInfoFragment.kt
+++ b/app/src/main/java/com/glia/exampleapp/VisitorInfoFragment.kt
@@ -13,10 +13,13 @@ import android.widget.Toast
 import androidx.appcompat.widget.SwitchCompat
 import androidx.fragment.app.Fragment
 import androidx.recyclerview.widget.RecyclerView
-import com.glia.androidsdk.Glia
-import com.glia.androidsdk.GliaException
-import com.glia.androidsdk.visitor.VisitorInfo
-import com.glia.androidsdk.visitor.VisitorInfoUpdateRequest
+import com.glia.widgets.GliaWidgets
+import com.glia.widgets.GliaWidgetsException
+import com.glia.widgets.callbacks.OnError
+import com.glia.widgets.callbacks.OnSuccess
+import com.glia.widgets.callbacks.OnComplete
+import com.glia.widgets.core.visitor.VisitorInfo
+import com.glia.widgets.core.visitor.VisitorInfoUpdateRequest
 import java.util.UUID
 
 class VisitorInfoFragment : Fragment() {
@@ -77,20 +80,20 @@ class VisitorInfoFragment : Fragment() {
         customAttributesAdapter.setAttributes(visitorInfo.customAttributesMap)
     }
 
-    private fun showError(exception: GliaException) {
+    private fun showError(exception: GliaWidgetsException) {
         Toast.makeText(context, exception.message ?: exception.debugMessage, Toast.LENGTH_LONG).show()
     }
 
     private fun obtainVisitorInfoUpdateRequest(): VisitorInfoUpdateRequest {
-        return VisitorInfoUpdateRequest.Builder()
-            .setName(nameEditText.text.toString())
-            .setEmail(emailEditText.text.toString())
-            .setPhone(phoneEditText.text.toString())
-            .setNote(noteEditText.text.toString())
-            .setCustomAttributes(obtainCustomAttributes())
-            .setCustomAttrsUpdateMethod(obtainCustomAttributesUpdateMethod())
-            .setNoteUpdateMethod(obtainNoteUpdateMethod())
-            .build()
+        return VisitorInfoUpdateRequest(
+            name = nameEditText.text.toString(),
+            email = emailEditText.text.toString(),
+            phone = phoneEditText.text.toString(),
+            note = noteEditText.text.toString(),
+            customAttributes = obtainCustomAttributes(),
+            customAttrsUpdateMethod = obtainCustomAttributesUpdateMethod(),
+            noteUpdateMethod = obtainNoteUpdateMethod(),
+        )
     }
 
     private fun obtainNoteUpdateMethod() = if (noteModeSwitch.isChecked) {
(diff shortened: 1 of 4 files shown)


DavDo (commented):
Overall looks good, but taking into account my suggestion in #1255 I'd suggest having more client friendly approach and not stick to core's RequestCallback.
For example having separate callbacks for onSuccess and onFailure cases will be way better, and integrators don't have to check for response nullability.

widgetssdk/src/main/java/com/glia/widgets/core/visitor/VisitorInfoUpdateRequest.kt
```diff
@@ -0,0 +1,75 @@
...
+ * @see {@link GliaWidgets#getVisitorInfo(RequestCallback)}
+ */
+data class VisitorInfoUpdateRequest(
+    val name: String,
+    val email: String,
+    val phone: String,
+    val note: String,
+    val noteUpdateMethod: NoteUpdateMethod,
+    val customAttributes: Map<String, String>,
+    val customAttrsUpdateMethod: CustomAttributesUpdateMethod
```
Andrii-Horishnii-Glia, line 19: These parameters should be optional.

widgetssdk/src/main/java/com/glia/widgets/core/visitor/VisitorInfo.kt
```diff
@@ -0,0 +1,32 @@
...
+    val name: String? = null,
+    val email: String? = null,
+    val phone: String? = null,
+    val note: String? = null,
+    val customAttributesMap: Map<String, String>,
+    val generatedName: String? = null,
+    val banned: Boolean? = null,
+    val href: String? = null,
+    val id: String? = null,
+    val externalId: String? = null,
```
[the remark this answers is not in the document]
Andrii-Horishnii-Glia: Yes, we need to align the platforms.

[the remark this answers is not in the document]
andrews-moc: I've asked Rasmus to consider this change in iOS SDK. ✅

widgetssdk/src/main/java/com/glia/widgets/core/visitor/VisitorInfoUpdateRequest.kt
andrews-moc: It's an excellent point.
Thank you.
Fixed [here](https://github.com/salemove/android-sdk-widgets/pull/1257/commits/6e753ab1ae26487cad22f4508b18602c0159861c).

andrews-moc:
> Overall looks good, but taking into account my suggestion in #1255 I'd suggest having more client friendly approach and not stick to core's RequestCallback. For example having separate callbacks for onSuccess and onFailure cases will be way better, and integrators don't have to check for response nullability.

I made appropriate changes and would appreciate your feedback.

DavDo (approved):
LGTM

app/src/main/java/com/glia/exampleapp/VisitorInfoFragment.kt
```diff
@@ -77,20 +80,20 @@ class VisitorInfoFragment : Fragment() {
...
-        return VisitorInfoUpdateRequest.Builder()
-            .setName(nameEditText.text.toString())
-            .setEmail(emailEditText.text.toString())
-            .setPhone(phoneEditText.text.toString())
-            .setNote(noteEditText.text.toString())
-            .setCustomAttributes(obtainCustomAttributes())
-            .setCustomAttrsUpdateMethod(obtainCustomAttributesUpdateMethod())
-            .setNoteUpdateMethod(obtainNoteUpdateMethod())
-            .build()
+        return VisitorInfoUpdateRequest(
```
Andrii-Horishnii-Glia, line 88: Did we decide to get rid of builders?

Andrii-Horishnii-Glia (approved):

app/src/main/java/com/glia/exampleapp/VisitorInfoFragment.kt
andrews-moc: I don't think we discussed it. Do you think we should keep the builder pattern? Honestly, I don't see much benefits of using it in Kotlin.

andrews-moc: As Andrii said in a private message, iOS SDK doesn't have builders, so we can look at it as a unification between platforms.

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.java
```diff
@@ -171,6 +178,37 @@ public static void onActivityResult(int requestCode, int resultCode, Intent data
         }
     }
 
+    /**
+     * Fetches the visitor's information
+     */
+    public static void getVisitorInfo(OnWidgetsResult<VisitorInfo> onWidgetsResult,
+                                      OnWidgetsError onWidgetsError) {
```
gugalo, line 185: I personally don't like when you have to pass multiple callbacks/listeners in same method. What is you preference regarding this. For example we could introduce one more interface that will extend both of these callbacks. For example:
```
OnWidgetsCallback<T>: OnWidgetsResult<T>, OnWidgetsError
```
And use it in here instead.

It is a question of each personal preference so I am up for a discussion. But I do want to discuss this topic since once it is in a public interface it might be hard to change it later.

DavDo: I would personally prefer a) because:

- From the integrator side it is easy and compact to use it, instead of having one callback with 2 or more functions(In my view it looks awful).
- This way we can make the result `non-null` which will prevent visitors checking result on being null each time receiving it.
- I'd not prefer b) because in my view it adds more flexibility.

Andrii-Horishnii-Glia: a) it is way Kotlin/functional programming style.
It is better to use:
```
foo({
  // result
}, {
  // error
})
```

than:
```
foo(object: WidgetsRequestCallback {
  override fun onResult(result: Bar?) {
    // result
  }
  
  override fun onError(exception: GliaWidgetsException?) {
    // error
  }
})
```

DavDo:
> @gugalo @DavDo @Andrii-Horishnii-Glia I have changed callbacks as per our discussion. Please see [this commit.](https://github.com/salemove/android-sdk-widgets/pull/1255/commits/24fa34f5adf53e36fe635e11810964af486e2ca9).
> 
> And on top of that, [here](https://github.com/salemove/android-sdk-widgets/pull/1257/commits/a68be1a423ac6b1391d0b5b5159ffbf1a877835a) is how it looks like for get and update visitor info now.
> 
> Please let me know if you have any concerns/suggestions. Thank you!

Looks Good!

gugalo (approved):

PR #1260, opened 2025-04-09
Convert GliaWidgets.java class to Kotlin

andrews-moc:
**Jira issue:**
[Convert GliaWidgets.java to Kotlin](https://glia.atlassian.net/browse/MOB-4268)

 - [x] Feature
 - [x] Ignore

**Additional info:**


gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.kt
```diff
@@ -0,0 +1,399 @@
...
+            if ((error is NullPointerException) || (error is IllegalArgumentException)) {
+                // that's likely a bug in the application
+                throwUncaughtException(error)
+                return@setErrorHandler
+            }
+            if (error is IllegalStateException) {
+                // that's a bug in RxJava or in a custom operator
+                throwUncaughtException(error)
+                return@setErrorHandler
+            }
```
Andrii-Horishnii-Glia, line 376: What about using `when` expressions here?

Andrii-Horishnii-Glia (approved):

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.kt
andrews-moc: Good point. I've improved it here c15d1c7a207fb78ed33418197a775134f2f6e763.

PR #1261, opened 2025-04-10
Revert "Quick fix for onAppCreate()"

Andrii-Horishnii-Glia:
**What was solved?**
Just revert the quick fix changes.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/Application.java
+++ b/app/src/main/java/com/glia/exampleapp/Application.java
@@ -3,7 +3,6 @@ package com.glia.exampleapp;
 import android.text.TextUtils;
 import android.util.Log;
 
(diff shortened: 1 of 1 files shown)


andrews-moc (commented):
I see that you use a new approach for naming commits ("chore: ") here and in React Native repository. Do we have it described somewhere in Confluence?

andrews-moc (approved):

andrews-moc:
> @andrews-moc No, we don't. You don't like it? I can remove it here.

I would just like to know/agree on the list of all possible tags and their descriptions so that everyone was on the same page.

DavDo (approved):

PR #1267, opened 2025-04-11
Fix - Bubble not visible on Android after screen is rotated

DavDo:
…ndling during configuration changes

**Jira issue:**
https://glia.atlassian.net/browse/MOB-4274

**What was solved?**
- Prevent Bubble disappearing after screen rotation
- Prevent Bubble head from going off-screen

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**


app/src/main/AndroidManifest.xml
```diff
@@ -12,6 +12,8 @@
         android:networkSecurityConfig="@xml/network_security_config"
         android:theme="@style/Theme.GliaAndroidSDKWidgetsExample">
 
+<!--        Add this parameter to activity to prevent the activity from restarting when the device is rotated-->
+<!--        android:configChanges="orientation|screenSize"-->
```
andrews-moc, line 16: Is it for testing purposes? If so, I think it makes sense to mention it as well as describe when it's needed, what scenario is it for.

widgetssdk/src/main/java/com/glia/widgets/view/head/ChatHeadLayout.kt
```diff
@@ -39,6 +39,7 @@ internal class ChatHeadLayout @JvmOverloads constructor(
 
     private val chatHeadSize: Float by lazy { resources.getDimension(R.dimen.glia_chat_head_size) }
     private val chatHeadMargin: Float by lazy { resources.getDimension(R.dimen.glia_chat_head_content_padding) }
+    private val dragMargin by lazy { resources.getDimension(R.dimen.glia_small) }
```
andrews-moc, line 42: As far as I understood, it's a minimum value for the `coerceIn()` function. What do you think about naming it `minDraggingValue`, `minimumCoordinate` or whatnot? I'm suggesting it because `dragMargin` is associated with padding, space between views to me.

andrews-moc (approved):

app/src/main/AndroidManifest.xml
Andrii-Horishnii-Glia: What about adding this to the specific activity for testing the bubble?

Andrii-Horishnii-Glia (approved):

app/src/main/AndroidManifest.xml
DavDo: Great ideas!
We have `Legacy Activity` that we need to test dialogs, I guess it is the best place to add this configuration.

widgetssdk/src/main/java/com/glia/widgets/view/head/ChatHeadLayout.kt
DavDo: This is the bubble margins from the screen borders.
I'll think how to rename it to sound more specific.

app/src/main/AndroidManifest.xml
DavDo: fixed

widgetssdk/src/main/java/com/glia/widgets/view/head/ChatHeadLayout.kt
DavDo: fixed!

widgetssdk/src/main/java/com/glia/widgets/view/head/ChatHeadLayout.kt
```diff
@@ -39,6 +39,7 @@ internal class ChatHeadLayout @JvmOverloads constructor(
 
     private val chatHeadSize: Float by lazy { resources.getDimension(R.dimen.glia_chat_head_size) }
     private val chatHeadMargin: Float by lazy { resources.getDimension(R.dimen.glia_chat_head_content_padding) }
+    private val bubbleDragSpaceMargin by lazy { resources.getDimension(R.dimen.glia_small) }
```
andrews-moc, line 42: NIT NIT. Maybe just `bubbleMargin`? 🙂 
'Space' duplicates the 'Margin', per se. 'Drag' doesn't seem to be an important detail. 🙂

DavDo: `chatHeadMargin` is about the same, but only applied to the right and bottom borders(when the bubble is initially placed and when dragging). The `bubbleDragSpaceMargin` is applied to the left and top borders only when user moves the bubble.
I was unable to use the initial `chatHeadMargin` because, the bubble itself has some asymmetry in internal paddings.
Maybe just name them `chatHeadBottomRightMargin`, `chatHeadTopRight` margin?

andrews-moc: Oh, I didn't notice `chatHeadMargin`. Yes, I like the idea of `chatHeadBottomRightMargin`, `chatHeadTopRight`. It reflects the actual meaning and is clear.

DavDo: Changed

PR #1268, opened 2025-04-14
Add callback to GliaWidgets.init() so that integrators could know when SDK init finishes

andrews-moc:
**What was solved?**
- [Add callback to GliaWidgets.init() so that integrators could know when SDK init finishes](https://glia.atlassian.net/browse/MOB-4273)
- [Add GliaWidgets.isInitialized() to Android Widgets SDK](https://glia.atlassian.net/browse/MOB-4211)
- [In progress] I will review the unit tests

These changes rely on [this Core SDK PR](https://github.com/salemove/android-sdk/pull/813).

**Release notes:**
Add callback to GliaWidgets.init() so that integrators could know when SDK init finishes

 - [x] Feature
 - [ ] Ignore

**Additional info:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/di/Dependencies.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/di/Dependencies.kt
@@ -7,6 +7,7 @@ import androidx.core.content.ContextCompat
 import androidx.lifecycle.Lifecycle
 import com.glia.androidsdk.Glia
 import com.glia.androidsdk.GliaConfig
+import com.glia.androidsdk.RequestCallback
 import com.glia.widgets.GliaWidgetsConfig
 import com.glia.widgets.StringProvider
 import com.glia.widgets.callvisualizer.CallVisualizerActivityWatcher
@@ -223,6 +224,20 @@ internal object Dependencies {
         repositoryFactory.initialize()
         configurationManager.applyConfiguration(gliaWidgetsConfig)
         localeProvider.setCompanyName(gliaWidgetsConfig.companyName)
+    }
+
+    @JvmStatic
+    fun onSdkInit(gliaWidgetsConfig: GliaWidgetsConfig, callback: RequestCallback<Boolean?>? = null) {
+        val gliaConfig = createGliaConfig(gliaWidgetsConfig)
+        gliaCore.init(gliaConfig) { success, error ->
+            if (error == null) {
+                controllerFactory.init()
+                repositoryFactory.initialize()
+                configurationManager.applyConfiguration(gliaWidgetsConfig)
+                localeProvider.setCompanyName(gliaWidgetsConfig.companyName)
+            }
+            callback?.onResult(success, error)
+        }
     }
 
     private fun createGliaConfig(gliaWidgetsConfig: GliaWidgetsConfig): GliaConfig {
--- a/app/src/main/java/com/glia/exampleapp/MainFragment.kt
+++ b/app/src/main/java/com/glia/exampleapp/MainFragment.kt
@@ -492,7 +492,7 @@ class MainFragment : Fragment() {
     }
 
     private fun initGliaWidgets() {
-        if (Glia.isInitialized()) {
+        if (GliaWidgets.isInitialized()) {
             setupAuthButtonsVisibility()
             listenForCallVisualizerEngagements()
             return
@@ -503,13 +503,18 @@ class MainFragment : Fragment() {
                 context = requireActivity().applicationContext,
 //                uiJsonRemoteConfig = UnifiedUiConfigurationLoader.fetchLocalConfigSample(requireContext()),
 //                region = "us"
-            )
-        )
-        prepareAuthentication()
-        setupAuthButtonsVisibility()
-        listenForCallVisualizerEngagements()
+            ),
+            onComplete = {
+                prepareAuthentication()
+                setupAuthButtonsVisibility()
+                listenForCallVisualizerEngagements()
 
-        view?.post { initMenu() }
+                view?.post { initMenu() }
+            },
+            onError = { error ->
+                error?.let { showToast(it.message.toString()) }
+            }
+        )
     }
 
     private fun prepareAuthentication() {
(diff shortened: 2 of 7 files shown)


widgetssdk/src/main/java/com/glia/widgets/di/Dependencies.kt
```diff
@@ -225,6 +226,20 @@ internal object Dependencies {
         localeProvider.setCompanyName(gliaWidgetsConfig.companyName)
     }
 
+    @JvmStatic
+    fun onSdkInit(gliaWidgetsConfig: GliaWidgetsConfig, callback: RequestCallback<Boolean?>? = null) {
+        val gliaConfig = createGliaConfig(gliaWidgetsConfig)
+        gliaCore.init(gliaConfig) { success, error ->
+            if (error == null) {
+                controllerFactory.init()
```
Andrii-Horishnii-Glia, line 234: I'm not sure if we need to do it only after the core SDK is initiated.
This means the SDK will crash if an integrator immediately calls something else after `.init()`.

andrews-moc: Yes, exactly, the SDK will crash if an integrator immediately calls something else after `init()`. I think it's expected if `onComplete` and `onError` callbacks are passed to `init()`. 

Otherwise, it looks logically inconsistent to me, if I leave it the way it was before (SDK is considered initialized even if init failed). In this case init callback will receive an error only on a first attempt. All subsequent retry attempts will call onComplete callback even if the SDK is not still functioning in fact.

DavDo: This is a little bit controversial to me.
We implemented Entry Widget in a way to support it event when the SDK is not initialized, but SDK will crash if something is called before it is initialized.

Andrii-Horishnii-Glia: Can we leave initializing process as we have? Can we just add the success and error callbacks?

andrews-moc: > Can we leave initializing process as we have? Can we just add the success and error callbacks?

@Andrii-Horishnii-Glia 
Okay, let me describe a scenario:
- Integrator has specified incorrect credentials.
- Calls `GliaWidgets.init()`.
- Receives a `Failed to initialise Glia Widgets SDK. Please check credentials., cause: INVALID_INPUT` error in `onError` callback. That's fine.
- However, if they call `GliaWidgets.init()` for the second time, the behavior differs - `onError` callback is called with `Glia SDK is already initialized, cause: INVALID_INPUT`.

I see two issues here:
1. SDK is not initialized correctly but isInitialized() returns `true`.
2. Errors should be consistent for a first and subsequent `GliaWidgets.init()` calls.
3. iOS SDK sends an error "💥 Core SDK configuration is not valid. Unexpected error=\'siteSettingsRetrieval\'." on **each** attempt.

@DavDo @Andrii-Horishnii-Glia Do you agree that with these issues?

In my opinion, for the new `GliaWidgets.init(config, onComplete, onError)`:
1. onComplete() should be called if credentials are correct. We can consider that a successful `setupSiteInfoLoading()` as an indicator for this.
2. `onError` should be called with a consistent errors.
4. `isInitialized()` should return `true` only if credentials are correct.
5. iOS and Android should be aligned as much as possible

@DavDo @Andrii-Horishnii-Glia What is your vision?

To solve them, I [moved](https://github.com/salemove/android-sdk/pull/813/files#diff-972253a1cd4a18eb13c28411fac59fa55a5ccc4e0f7532be88cbe3921e7ca62eR194) the `setInitialized(true)` to the stage when `setupSiteInfoLoading()` in Core SDK succeeds. If I left the `controllerFactory.init()`, `repositoryFactory.initialize()`, as they were before (not waiting for `setupSiteInfoLoading()` to succeed), then they are called **before** `setInitialized()` is set to `true`. `assertThatInit()` in Core SDK throws `GliaException("SDK is not initialized", GliaException.Cause.INVALID_INPUT)`.

andrews-moc: For the record, I've left the behavior for the current `GliaWidgets.init()` without callbacks intact, without any changes.

Andrii-Horishnii-Glia: I highly doubt anyone will call .init() multiple times if there is an error.

Andrii-Horishnii-Glia: I would like to see the same behavior from the SDK with `.init(config, onComplete, onError)` as it is with `.init(config)`. 
However, your description also makes sense.

Andrii-Horishnii-Glia: Maybe there is nothing critical. Unfortunately, I don't have much time today to test it.

DavDo: After your description, I agree with every point.
I support current solution.

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/di/Dependencies.kt
Pelkar: Having a callback that can provide additional info(error or completion) is a good approach in my opinion, but it comes down to the team, because you will be supporting it in the long run. If API does not break, then I will give my vote to adding callbacks.

andrews-moc: @Pelkar Thank you for the feedback.
The question raised here is more about the way how the callbacks will receive calls and the moment when `GliaWidgets.isInitialized()` returns `true`. You can check these details in my verbose comment above and share your opinion, agreement/disagreement. 😉

gugalo: I am also okay with current solution. But at the same time also think it would be nice to have both `isInitialised()` and `isinitialising()` statuses clearly.

gugalo (approved):

PR #1272, opened 2025-04-17
Add ChatMessage to Android Widgets SDK

andrews-moc:
**Jira issue:**
[Add public class for ChatMessage to Android Widgets SDK](https://glia.atlassian.net/browse/MOB-4209)

**What was solved?**
- Added a Widgets wrapper for `ChatMessage` class, including `Participant`
- Added a Widgets wrapper for `MessageAttachment` class
- Dropped `SingleChoiceAttachment` class usage from `ExampleCustomCardAdapter` (together with @Andrii-Horishnii-Glia decided that it doesn't make sense)
- Fixed and added unit tests

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**


widgetssdk/src/main/java/com/glia/widgets/chat/adapter/CustomCardAdapter.java
```diff
@@ -10,7 +10,6 @@
 import androidx.annotation.VisibleForTesting;
 import androidx.collection.SparseArrayCompat;
 
-import com.glia.androidsdk.chat.ChatMessage;
```
Andrii-Horishnii-Glia, line 13: This means integrators should make the changes right after the release. Should we support the ChatMessage from the Core SDK right now?

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/ChatMessage.kt
```diff
@@ -0,0 +1,53 @@
...
+class ChatMessage(
+    val id: String,
+    val content: String,
+    val timestamp: Long,
+    val deliveredAt: String?,
+    val attachment: MessageAttachment?,
+    val metadata: JSONObject?,
+    val senderType: Participant
+) {
+    constructor(chatMessage: CoreChatMessage) : this(
```
DavDo, line 28: Can we make this constructor internal? Or integrators need it?

widgetssdk/src/main/java/com/glia/widgets/chat/domain/CustomCardShouldShowUseCase.kt
```diff
@@ -0,0 +1,31 @@
+package com.glia.widgets.chat.domain
+
+import com.glia.androidsdk.chat.ChatMessage
+import com.glia.androidsdk.chat.SingleChoiceAttachment
+import com.glia.widgets.chat.adapter.CustomCardAdapter
+
+/**
+ * @hide
+ */
+class CustomCardShouldShowUseCase(private val adapter: CustomCardAdapter?) {
```
DavDo, line 10: in Kotlin it's enough to make the class internal, instead of using `@hide`

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/ChatMessage.kt
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @property id The unique identifier of the message.
+ * @property content The content of the message.
+ * @property timestamp The timestamp of when the message was received by the server, in milliseconds.
+ * @property deliveredAt The date when the message was marked as read, or `null` if not marked yet.
+ * @property attachment The attachment associated with the message, or `null` if none exists.
+ * @property metadata Custom payloads for tailoring the visitor experience, or `null` if none exists.
+ * @property senderType The type of sender (visitor or operator).
+ */
+class ChatMessage(
```
DavDo, line 19: Maybe it makes sense to make this class `data` class to have `toString` overriden by default?
I'm also unable to understand if integrators need need whole class or having `metadata` is enough.

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/ChatMessage.kt
andrews-moc: Fixed ef470376e3f759d92b1eba9be9cfdf9fba107080

widgetssdk/src/main/java/com/glia/widgets/chat/domain/CustomCardShouldShowUseCase.kt
andrews-moc: Fixed ef470376e3f759d92b1eba9be9cfdf9fba107080

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/ChatMessage.kt
andrews-moc: Made it a data class ef470376e3f759d92b1eba9be9cfdf9fba107080

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/CustomCardAdapter.java
andrews-moc: @Andrii-Horishnii-Glia  Yes. Exactly. The only way to support current `CustomCardAdapter` that I see is to leave it as is and add a new `WidgetsCustomCardAdapter` that uses `ChatMessage` from Widgets. However, my initial thought was that we can introduce a breaking change in a minor version (requiring a one line import change from integrators). Instead of supporting two `CustomCardAdapters`.

Do you think this breaking change is not acceptable?
Do you have better ideas for a solution?

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/ChatMessage.kt
DavDo: As far as I informed, Custom Cards are designed to use metadata.
Since we're going to introduce some breaking changes with closing Core SDK from the integrators, I guess it's okay to introduce one more and don't care about the integrators who use additional properties from the core's ChatMessage.

andrews-moc: @DavDo Aligned the `CustomCardMessage` structure with iOS `MessageRenderer` in [this](https://github.com/salemove/android-sdk-widgets/pull/1272/commits/c086431bdd4868996932d24e1025d3c2679fc24b#diff-608ce7c204e3c9cb11ccaf7ca8faf27a4b542540428c8fe6ea2c2fd1c2a34ae9) commit.

However, I'm not sure what shall I use instead of `.getContent()` [here in ExampleCustomCardAdapter](https://github.com/salemove/android-sdk-widgets/pull/1272/commits/c086431bdd4868996932d24e1025d3c2679fc24b#diff-b189e005acf6a0a8372478e87e6f4c0d542e9ca44b751a2e990ccac5f3208e77R106). Maybe @Andrii-Horishnii-Glia can help. 🤔

app/src/main/java/com/glia/exampleapp/ExampleCustomCardAdapter.java
```diff
@@ -101,19 +102,14 @@ public static boolean isNativeViewType(ChatMessage message) {
         }
 
         @Override
-        public void bind(@NonNull ChatMessage message, @NonNull ResponseCallback callback) {
-            messageTextView.setText(message.getContent());
+        public void bind(@NonNull CustomCardMessage message, @NonNull ResponseCallback callback) {
+//            messageTextView.setText(message.getContent()); // TODO: 23.04.2025 what shall I use instead??
```
Andrii-Horishnii-Glia, line 106: Good question

DavDo: I think something from the `metadata`, because that's the way integrators can pass some custom properties to the `CustomCardAdapter` and example should show how to handle it.

Andrii-Horishnii-Glia: Probably Davit is right. However, messages can contain metadata and content. In this case, integrators should probably put everything in the metadata section.

DavDo: Is there a way for integrators to change content?

andrews-moc: I don't think integrators can change a `content` val in `ChatMessage`.

andrews-moc: So the question is if I shall replicate the `content` field from Core SDK's `ChatMessage` to Widget's `CustomCardMessage`. 🤔  Despite the fact that iOS doesn't have `content` in `MessageRenderer` class. I'm not sure if `MessageRenderer` is an exact analogy of Android `CustomCardMessage`, though.

Andrii-Horishnii-Glia: Let's base on the MessageRenderer model. If there are any requests, we can always add it later.

DavDo: > I don't think integrators can change a `content` val in `ChatMessage`.

I mean Integrators can send custom message with custom metadata using some glia api, but I think they can not control the content in this case.
I agree aligning with `MessageRenderer` model.

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/CustomCardAdapter.java
```diff
@@ -102,14 +127,32 @@ abstract public CustomCardViewHolder onCreateViewHolder(@NonNull ViewGroup paren
...
+        return false;
+    }
+
     @Nullable
-    public final Integer getChatAdapterViewType(ChatMessage chatMessage) {
-        Integer customCardViewType = getItemViewType(chatMessage);
+    public final Integer getChatAdapterViewType(ChatMessage message) {
+        Integer customCardViewType = getItemViewType(new CustomCardMessage(message));
+        if (customCardViewType == null) {
+            customCardViewType = getItemViewType(message);
```
Andrii-Horishnii-Glia, line 154: Could we add some warning/logging here?

widgetssdk/src/main/java/com/glia/widgets/chat/domain/CustomCardShouldShowUseCase.kt
```diff
@@ -0,0 +1,25 @@
...
+
+internal class CustomCardShouldShowUseCase(private val adapter: CustomCardAdapter?) {
+    fun execute(message: ChatMessage, viewType: Int, shouldApplySelectedOption: Boolean): Boolean {
+        if (shouldApplySelectedOption && !hasSelectedOption(message)) {
+            return true
+        }
+        if (adapter == null || message.metadata == null) {
+            return false
+        }
+        return adapter.shouldShowCard(CustomCardMessage(message), viewType)
```
Andrii-Horishnii-Glia, line 16: This call only new method. It is not backwards-compatible.

app/src/main/java/com/glia/exampleapp/ExampleCustomCardAdapter.java
andrews-moc: Okay, thank you.

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/holder/CustomCardViewHolder.java
```diff
@@ -49,12 +49,12 @@ public CustomCardViewHolder(@NonNull View itemView) {
...
 
     /**
-     * Called to display the data for the specified {@link ChatMessage}. This method should
+     * Called to display the data for the specified {@link CustomCardMessage}. This method should
      * update the content of {@link #itemView} to reflect the item for the given message.
      * @param message a chat message with metadata.
      * @param callback can be used to send the selected card option.
      */
-    public abstract void bind(@NonNull ChatMessage message, @NonNull ResponseCallback callback);
+    public abstract void bind(@NonNull CustomCardMessage message, @NonNull ResponseCallback callback);
```
Andrii-Horishnii-Glia, line 57: CustomCardViewHolder supports only new message type. No backwards-compatibility.

Andrii-Horishnii-Glia (commented):
It is still not backward-compatibility.

Andrii-Horishnii-Glia (changes requested):
My suggestions:
- Convert `CustomCardAdapter` to Kotlin
- Use something like this for all methods that use "new"/"old" parameters:
```
    open fun shouldShowCard(message: ChatMessage, viewType: Int): Boolean {
        return shouldShowCard(CustomCardMessage(message), viewType)
    }
```
- From use cases and the chat adapter call only methods with the "old" parameters.

So, all logic for "new"/"old" parameters will be in the CustomCardAdapter.

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/holder/CustomCardViewHolder.java
andrews-moc: I missed it, indeed. It's a blocker now.
Unfortunately, I have not been able to find a fully backward-compatible solution so far.

If I change the `bind()` function to work with `CustomCardMessage` => it will require some migrations from integrators.

If I leave the `bind()` function to work with ChatMessage => integrators would not be able to use it after hiding Core SDK.

If I provide two `CustomCardViewHolder`s: one with `bind(ChatMessage)` and the other one with `bind(CustomCardMessage)` => then I face with incompatible types in `CustomCardAdapter: abstract public CustomCardViewHolder onCreateViewHolder()`.

I don't have other ideas than to refuse the idea of backwards compatibility for Custom Cards, implement with breaking changes, postpone the merge until 4.0 release.

andrews-moc: Once again, let me make a short recap:
- At first, I introduced a `ChatMessage` and `MessageAttachment` to Widgets.
- The idea was to release with a small change required from integrators: `ChatMessage` and `MessageAttachment` import change.
- We realised that Custom Cards do not need all the fields of `ChatMessage`. It needs only metadata. Like it's on iOS.
- Taking this into account, it would require much more migrations from integrators, so I decided to try to develop a backwards compatible solution as much as I can.
- However, now I came to a point where I can't find a fully backward-compatible solution.
- If wouldn't have any ideas, I'd suggest to refuse the idea of backwards compatibility for Custom Cards, implement with breaking changes, postpone the merge until 4.0 release.

andrews-moc:
> * Convert `CustomCardAdapter` to Kotlin

I didn't want to overblow the pull request, let me do this in a separate PR.

Andrii-Horishnii-Glia:
@andrews-moc Yes, it is indeed better to do the conversion separately.
But what prevents us from making the changes that I made in this commit?
c1143338ac79aad5c1e4283c105ae8131179feb4

In my view, we add a new class and save backward-compatibility until the 4.0.0 release.

andrews-moc:
> But what prevents us from making the changes that I made in this commit? [c114333](https://github.com/salemove/android-sdk-widgets/commit/c1143338ac79aad5c1e4283c105ae8131179feb4)
> 
> In my view, we add a new class and save backward-compatibility until the 4.0.0 release.

@Andrii-Horishnii-Glia I didn't consider removing the `abstract` keyword from the `bind()` function because it removes the foolproof from incorrect integrator's implementation. Seems it's the only way to keep backward-compatibility, though.

Andrii-Horishnii-Glia:
@andrews-moc I believe the `abstract` keyword is not a tool to facilitate the integrators' process for integrators. Hundreds of SDKs provide abstract classes where developers should override some open methods, and this is not an issue. We just need to describe this in the documentation.

Andrii-Horishnii-Glia (approved):

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/CustomCardAdapter.java
andrews-moc: Added an info level log d34f3098e6d6a46c5b7eddfea8b91162746a41f5.
As `getItemViewType()` should be overridden by an integrator, I decided to go with 'info' level rather than a warning. Let me know if you still insist on 'warning'.

andrews-moc: However, `Get item view type for CustomCardMessage returned null` log will be sent from integrators who continue using (basically override) the deprecated `getItemViewType(message: ChatMessage)`. @Andrii-Horishnii-Glia Is this what you suggested or I misunderstood you?

andrews-moc: Let's me also add to the previous message that it can be sent multiple times - each time a Custom Card message is being shown.

andrews-moc:
> My suggestions:
> 
> * Convert `CustomCardAdapter` to Kotlin

Pull request: [Convert Custom Card related classes to Kotlin](https://github.com/salemove/android-sdk-widgets/pull/1281)

widgetssdk/src/main/java/com/glia/widgets/chat/adapter/CustomCardAdapter.java
Andrii-Horishnii-Glia: I thought a bit. If integrators return `null` it is not an issue. But we can do warning if: `getItemViewType(CustomCardMessage) == null && getItemViewType(ChatMessage) != null`. It means they still use the old method.

PR #1288, opened 2025-04-30
Refactor Authentication interfaces

andrews-moc:
**Jira issue:**
[Add Authentication class to Android Widgets SDK](https://glia.atlassian.net/browse/MOB-4206)

**What was solved?**
- Remove Core SDK Authentication inheritance from Widgets Authentication class
- Make AuthenticationManager implement Widgets Authentication class
- Move public Widgets Authentication class to a supposedly public package

**Release notes:**

 - [x] Feature
 - [x] Ignore

**Additional info:**


Andrii-Horishnii-Glia (approved):
Looks good

widgetssdk/src/main/java/com/glia/widgets/authentication/Authentication.kt
```diff
@@ -0,0 +1,94 @@
...
+     * Exception may have one of following causes:
+     * <br></br> [GliaException.Cause.NETWORK_TIMEOUT] - when request times out due to connection issues
+     * <br></br> [GliaException.Cause.INTERNAL_ERROR] - when internal error occurs
+     * <br></br> [GliaException.Cause.AUTHENTICATION_ERROR] - when authentication fails
+     */
+    fun refresh(
+        jwtToken: String,
+        externalAccessToken: String?,
+        authCallback: RequestCallback<Void>?
+    )
```
gugalo, line 78: Just to be sure. 
What will happen if integrators who use Java call this method and pass `jwtToken` as `null`? Will it crash the app at runtime with a fatal exception, or will it throw an error during compilation?

Andrii-Horishnii-Glia: It will crash an application at runtime.
However, Android Studio should show the warning the same way as for `@NotNull`. So, I don't see an issue here.

gugalo (approved):

PR #1289, opened 2025-05-02
Add public class for SiteApiKey to Android Widgets SDK

Andrii-Horishnii-Glia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-4212

**What was solved?**
Add public class for SiteApiKey to Android Widgets SDK

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsConfig.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsConfig.kt
@@ -1,7 +1,6 @@
 package com.glia.widgets
 
 import android.content.Context
-import com.glia.androidsdk.SiteApiKey
 import com.glia.androidsdk.screensharing.ScreenSharing
 import com.glia.widgets.helper.Logger
 import com.glia.widgets.helper.TAG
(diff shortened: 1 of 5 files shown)


widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsConfig.kt
```diff
@@ -176,6 +175,17 @@ class GliaWidgetsConfig private constructor(builder: Builder) {
...
         }
 
+        @Deprecated(
+            "Use {@link #setSiteApiKey(SiteApiKey)}",
+            ReplaceWith(
+                "setSiteApiKey(SiteApiKey(id, secret))",
+                "com.glia.widgets.SiteApiKey"
+            )
+        )
+        fun setSiteApiKey(siteApiKey: com.glia.androidsdk.SiteApiKey?): Builder {
```
DavDo, line 185: This question is not related to this PR, but maybe this is the best place to discuss it.
I feel like this property should be non-null. WDYT?

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsConfig.kt
Andrii-Horishnii-Glia: It definitely makes a lot of sense to use non-null parameters here.

Andrii-Horishnii-Glia: Yes. However the new method can be non-null now.

Andrii-Horishnii-Glia:
!squash

PR #1293, opened 2025-05-05
Add support for suppressing push notification dialog with deep link

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-4325

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/ExampleAppConfigManager.kt
+++ b/app/src/main/java/com/glia/exampleapp/ExampleAppConfigManager.kt
@@ -7,6 +7,7 @@ import androidx.preference.PreferenceManager
 import com.glia.androidsdk.SiteApiKey
 import com.glia.androidsdk.screensharing.ScreenSharing
 import com.glia.widgets.GliaWidgetsConfig
+import androidx.core.content.edit
 
 /**
  * Helper class to obtain Glia Config params from deep-link or preferences.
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/glia/exampleapp/ExampleAppConfigManager.kt
```diff
@@ -26,13 +27,15 @@ object ExampleAppConfigManager {
     private const val REGION_ACCEPTANCE = "acceptance"
     private const val BASE_DOMAIN = "base_domain"
     private const val DEFAULT_BASE_DOMAIN = "at.samo.io"
+    private const val SUPPRESS_PN_DIALOG_KEY = "suppress_pn_permission_dialog"
```
Andrii-Horishnii-Glia, line 30: Does it align with a deep link parameter on the iOS testing applications?

Andrii-Horishnii-Glia (approved):

app/src/main/java/com/glia/exampleapp/ExampleAppConfigManager.kt
DavDo: Should we align them?
I don't know if iOS team needs it at all or not.
@salemove/tm-mobile-ios

DavDo: @salemove/tm-mobile-ios merging this PR, let me know if `suppress_pn_permission_dialog` query param does not work for you for deep links.

PR #1300, opened 2025-05-12
Replace RequestCallback with Widgets callbacks in Secure Conversation interfaces

andrews-moc:
**Jira issue:**
[MOB-4346](https://glia.atlassian.net/browse/MOB-4346)

**What was solved?**
- Replace RequestCallback with Widgets callbacks in Secure Conversation interfaces

**Release notes:**

 - [x] Feature
 - [x] Ignore

**Additional info:**

[MOB-4346]: https://glia.atlassian.net/browse/MOB-4346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
@@ -1,7 +1,10 @@
 package com.glia.widgets.secureconversations
 
-import com.glia.androidsdk.GliaException
 import com.glia.androidsdk.RequestCallback
+import com.glia.widgets.GliaWidgetsException
+import com.glia.widgets.callbacks.OnError
+import com.glia.widgets.callbacks.OnSuccess
+import com.glia.widgets.toWidgetsType
 
 /**
  * Secure Conversations offers the ability to asynchronously and securely communications for authenticated visitors.
@@ -15,24 +18,40 @@ interface SecureConversations {
      * This number will increase with each message sent by the operator
      * that the visitor has not yet marked as read.
      *
-     * @param callback a callback that returns [RequestCallback] with the number of unread
-     * secure messages on success, or [GliaException] on failure.
+     * @param onSuccess [OnSuccess] a callback that returns the number of unread
+     * secure messages on success.
+     * @param onError [OnError] a callback that returns [GliaWidgetsException] on failure.
      *
      * Exception may have one of the following causes:
-     * [GliaException.Cause.AUTHENTICATION_ERROR] -when a visitor is not authenticated
-     * [GliaException.Cause.INVALID_INPUT] - when SDK is not initialized
+     * [GliaWidgetsException.Cause.AUTHENTICATION_ERROR] -when a visitor is not authenticated
+     * [GliaWidgetsException.Cause.INVALID_INPUT] - when SDK is not initialized
      */
-    fun getUnreadMessageCount(callback: RequestCallback<Int?>)
+    fun getUnreadMessageCount(onSuccess: OnSuccess<Int>, onError: OnError? = null)
 
     /**
-     * The same as [SecureConversations.getUnreadMessageCount] but with the ability to subscribe to updates.
+     * Subscribes to updates of the unread message count.
+     *
+     * This method allows you to receive updates whenever the unread message count changes.
+     * The provided callback will be triggered with the updated count.
+     *
+     * @param callback [OnSuccess] A callback that will be invoked with the updated unread message count.
+     *
+     * Note: Ensure to unsubscribe using [unSubscribeFromUnreadMessageCount] when updates are no longer needed
+     * to avoid memory leaks or unnecessary updates.
      */
-    fun subscribeToUnreadMessageCount(callback: RequestCallback<Int?>)
+    fun subscribeToUnreadMessageCount(callback: OnSuccess<Int>)
 
     /**
-     * Unsubscribe from updates of the unread message count.
+     * Unsubscribes from updates of the unread message count.
+     *
+     * This method stops receiving updates for the unread message count for the provided callback.
+     *
+     * @param callback [OnSuccess] A callback that was previously subscribed to receive updates.
+     *
+     * Note: Ensure that the callback passed to this method is the same instance that was used
+     * during subscription to successfully unsubscribe.
      */
-    fun unSubscribeFromUnreadMessageCount(callback: RequestCallback<Int?>)
+    fun unSubscribeFromUnreadMessageCount(callback: OnSuccess<Int>)
 }
 
 /**
@@ -42,16 +61,34 @@ class SecureConversationsImpl(
     private val secureConversations: com.glia.androidsdk.secureconversations.SecureConversations
 ) : SecureConversations {
 
-    override fun getUnreadMessageCount(callback: RequestCallback<Int?>) {
-        secureConversations.getUnreadMessageCount(callback)
+    internal val subscribedCallbacks: MutableMap<Int, RequestCallback<Int>> = mutableMapOf()
+
+    override fun getUnreadMessageCount(onSuccess: OnSuccess<Int>, onError: OnError?) {
+        secureConversations.getUnreadMessageCount { count, gliaException ->
+            if (gliaException != null || count == null) {
+                onError?.onError(gliaException.toWidgetsType("Failed to get unread message count"))
+            } else {
+                onSuccess.onSuccess(count)
+            }
+        }
     }
 
-    override fun subscribeToUnreadMessageCount(callback: RequestCallback<Int?>) {
-        secureConversations.subscribeToUnreadMessageCount(callback)
+    override fun subscribeToUnreadMessageCount(callback: OnSuccess<Int>) {
+        if (subscribedCallbacks.containsKey(callback.hashCode())) {
+            // Already subscribed
+            return
+        }
+        val requestCallback: RequestCallback<Int> = RequestCallback { count, gliaException ->
+                if (gliaException == null && count != null) {
+                    callback.onSuccess(count)
+                }
+            }
+        subscribedCallbacks[callback.hashCode()] = requestCallback
+        secureConversations.subscribeToUnreadMessageCount(requestCallback)
     }
 
-    override fun unSubscribeFromUnreadMessageCount(callback: RequestCallback<Int?>) {
-        secureConversations.unSubscribeFromUnreadMessageCount(callback)
+    override fun unSubscribeFromUnreadMessageCount(callback: OnSuccess<Int>) {
+        subscribedCallbacks[callback.hashCode()]?.let { secureConversations.unSubscribeFromUnreadMessageCount(it) }
+        subscribedCallbacks.remove(callback.hashCode())
     }
 }
-
--- a/widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsException.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/GliaWidgetsException.kt
@@ -54,7 +54,7 @@ internal fun GliaCoreException?.toWidgetsType(
 ): GliaWidgetsException =
     this?.toWidgetsType() ?: let {
         val exception = GliaWidgetsException(defaultMessage, defaultCause)
-        Logger.e("GliaWidgetsException", "The Core GliaException is null, using the default message and cause.", exception)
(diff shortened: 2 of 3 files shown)


widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
```diff
@@ -42,16 +45,34 @@ class SecureConversationsImpl(
...
+            }
+        }
     }
 
-    override fun subscribeToUnreadMessageCount(callback: RequestCallback<Int?>) {
-        secureConversations.subscribeToUnreadMessageCount(callback)
+    override fun subscribeToUnreadMessageCount(onSuccess: OnSuccess<Int>?, onError: OnError?) {
+        secureConversations.subscribeToUnreadMessageCount { count, gliaException ->
+            if (gliaException != null || count == null) {
+                onError?.onError(gliaException.toWidgetsType("Failed to get unread message count"))
```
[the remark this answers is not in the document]
Andrii-Horishnii-Glia: Good Copilot :)

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
```diff
@@ -42,16 +45,34 @@ class SecureConversationsImpl(
     private val secureConversations: com.glia.androidsdk.secureconversations.SecureConversations
 ) : SecureConversations {
 
-    override fun getUnreadMessageCount(callback: RequestCallback<Int?>) {
-        secureConversations.getUnreadMessageCount(callback)
+    override fun getUnreadMessageCount(onSuccess: OnSuccess<Int>?, onError: OnError?) {
```
Andrii-Horishnii-Glia, line 48: What about adding the default values?
```suggestion
    override fun getUnreadMessageCount(onSuccess: OnSuccess<Int>? = null, onError: OnError? = null) {
```

Andrii-Horishnii-Glia (commented):
I don't think this solution will work. Correct me if I'm wrong.

Andrii-Horishnii-Glia (changes requested):
In my view, changes are needed.

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
```diff
@@ -42,16 +45,34 @@ class SecureConversationsImpl(
...
+            }
+        }
     }
 
-    override fun unSubscribeFromUnreadMessageCount(callback: RequestCallback<Int?>) {
-        secureConversations.unSubscribeFromUnreadMessageCount(callback)
+    override fun unSubscribeFromUnreadMessageCount(onSuccess: OnSuccess<Int>?, onError: OnError?) {
+        secureConversations.unSubscribeFromUnreadMessageCount { count, gliaException ->
+            if (gliaException != null || count == null) {
+                onError?.onError(gliaException.toWidgetsType("Failed to get unread message count"))
```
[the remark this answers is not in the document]
andrews-moc: Fixed 52f93e3463bf40d7c4153633e89cfef22ca3534b

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
```diff
@@ -42,16 +45,34 @@ class SecureConversationsImpl(
...
+            } else {
+                onSuccess?.onSuccess(count)
+            }
+        }
     }
 
-    override fun subscribeToUnreadMessageCount(callback: RequestCallback<Int?>) {
-        secureConversations.subscribeToUnreadMessageCount(callback)
+    override fun subscribeToUnreadMessageCount(onSuccess: OnSuccess<Int>?, onError: OnError?) {
```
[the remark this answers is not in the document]
Andrii-Horishnii-Glia: probably, yes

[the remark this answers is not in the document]
andrews-moc: I think it's needed. I can just add a default value (null) so that there is no need to pass it obligatorily.

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
andrews-moc: Agree with `onError`. Here and in other places. However, I don't think it makes sense to specify `null` as a default value for `onSuccess` because what's the point of calling this function then? Am I missing anything? Here is how I've changed it for now dbe3c9aa06b461e1298b1ead52ac26fb6cc2e0a6.

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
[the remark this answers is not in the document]
andrews-moc: Improved here dbe3c9aa06b461e1298b1ead52ac26fb6cc2e0a6

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
```diff
@@ -42,16 +45,34 @@ class SecureConversationsImpl(
...
+        secureConversations.subscribeToUnreadMessageCount { count, gliaException ->
```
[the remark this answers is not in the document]
andrews-moc: > A visitor object should be stored somewhere and used to unsubscribe.

Visitor object or callback object?

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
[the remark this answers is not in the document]
DavDo: I think this approach won't work here, unless we don't keep internal map of `OnSuccess` and `RequestCallback`, because unsubscribe is happening with the `RequestCallback` instance.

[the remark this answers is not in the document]
andrews-moc: Yes, I'm working on it, thank you.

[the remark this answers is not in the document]
andrews-moc: @Andrii-Horishnii-Glia @DavDo I've reworked the implementation. Please have a look now.

andrews-moc:
@Andrii-Horishnii-Glia  @DavDo I've reworked the implementation. Please have a look now.

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
andrews-moc: I've reworked the implementation. Please have a look now.

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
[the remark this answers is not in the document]
andrews-moc: I've reworked the implementation. Please have a look now.

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
```diff
@@ -15,24 +18,40 @@ interface SecureConversations {
...
+     * @param onError [OnError] a callback that returns [GliaWidgetsException] on failure.
      *
      * Exception may have one of the following causes:
-     * [GliaException.Cause.AUTHENTICATION_ERROR] -when a visitor is not authenticated
-     * [GliaException.Cause.INVALID_INPUT] - when SDK is not initialized
+     * [GliaWidgetsException.Cause.AUTHENTICATION_ERROR] -when a visitor is not authenticated
+     * [GliaWidgetsException.Cause.INVALID_INPUT] - when SDK is not initialized
      */
-    fun getUnreadMessageCount(callback: RequestCallback<Int?>)
+    fun getUnreadMessageCount(onSuccess: OnSuccess<Int>, onError: OnError? = null)
```
Andrii-Horishnii-Glia, line 29: Does it make sense to rename the `OnSuccess` to `OnResult`?

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
```diff
@@ -15,24 +18,40 @@ interface SecureConversations {
...
+     * This method allows you to receive updates whenever the unread message count changes.
+     * The provided callback will be triggered with the updated count.
+     *
+     * @param callback [OnSuccess] A callback that will be invoked with the updated unread message count.
+     *
+     * Note: Ensure to unsubscribe using [unSubscribeFromUnreadMessageCount] when updates are no longer needed
+     * to avoid memory leaks or unnecessary updates.
      */
-    fun subscribeToUnreadMessageCount(callback: RequestCallback<Int?>)
+    fun subscribeToUnreadMessageCount(callback: OnSuccess<Int>)
```
Andrii-Horishnii-Glia, line 42: Does it make sense to rename the `OnSuccess` to `OnResult`?

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
```diff
@@ -42,16 +61,34 @@ class SecureConversationsImpl(
     private val secureConversations: com.glia.androidsdk.secureconversations.SecureConversations
 ) : SecureConversations {
 
-    override fun getUnreadMessageCount(callback: RequestCallback<Int?>) {
-        secureConversations.getUnreadMessageCount(callback)
+    internal val subscribedCallbacks: MutableMap<Int, RequestCallback<Int>> = mutableMapOf()
```
Andrii-Horishnii-Glia, line 64: I'm not complaining, but why not:
```suggestion
    internal val subscribedCallbacks: MutableMap<OnSuccess<Int>, RequestCallback<Int>> = mutableMapOf()
```

Andrii-Horishnii-Glia (approved):
LGTM!

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
andrews-moc: The reasons for going with an Int hash code were:
1. Efficiency: Using hash codes as keys is generally more memory-efficient and faster for lookups, as hash codes are simple integers.
2. Simplicity in Storage: Storing an Int instead of the full object reduces the memory footprint of the map.

However, the expected number of `RequestCallback`  should be small and performance should not be a critical concern, I agree. So, I can change it to `MutableMap<OnSuccess<Int>, RequestCallback<Int>>` however, I don't see strong reasons for keeping the whole object as a key. 🤔

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
gugalo: I agree that in this case or any other where the callback might be fired more than once then the `OnResult` makes more sense

gugalo (approved):

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
Andrii-Horishnii-Glia: Yes, it can be memory-efficient. Also, we don't store a link to the object. However we store `RequestCallback`s.
This is a little more convenient to use, because it is difficult to confuse an object.

But, let it be `Int`.

widgetssdk/src/main/java/com/glia/widgets/secureconversations/SecureConversations.kt
andrews-moc: Okay, I will rename `OnSuccess` to `OnResult` in a separate PR.

Andrii-Horishnii-Glia: So, should we add new callback type?

andrews-moc: > So, should we add new callback type?

@Andrii-Horishnii-Glia I renamed `OnSuccess` callback to `OnResult` in [this PR](https://github.com/salemove/android-sdk-widgets/pull/1303). Are you thinking about something else except it?

PR #1306, opened 2025-05-14
Refactor callbacks

Andrii-Horishnii-Glia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-4355

**What was solved?**
The idea is to make the interface more convenient for integrators. Callbacks are divided into 3 types:
- Result callbacks not nullable (OnComplete, OnResult). Integrators will always have result.
- OnError callbacks nullable for getters (getQueues, getVisitorInfo, getUnreadMessageCount, etc). In case integrators need the result, but don't want to handle the errors.
- make the OnError callback not nullable for methods that update something or perform some action (init, updateVisitorInfo, etc). Important actions should always be handled.

**Release notes:**

 - [x] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/app/src/main/java/com/glia/exampleapp/VisitorInfoFragment.kt
+++ b/app/src/main/java/com/glia/exampleapp/VisitorInfoFragment.kt
@@ -118,10 +118,10 @@ class VisitorInfoFragment : Fragment() {
 
     private fun getVisitorInfo() {
         saveButton.text = getString(R.string.visitor_info_loading)
-        val onResult: OnResult<VisitorInfo?> = OnResult { result ->
+        val onResult: OnResult<VisitorInfo> = OnResult { result ->
             view?.post {
                 saveButton.text = getString(R.string.visitor_info_save)
-                result?.let { showVisitorInfo(it) }
+                showVisitorInfo(result)
             }
         }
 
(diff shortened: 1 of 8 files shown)


DavDo (approved):
Good job 👍

app/src/main/java/com/glia/exampleapp/MainFragment.kt
```diff
@@ -540,7 +540,7 @@ class MainFragment : Fragment() {
                 view?.post { initMenu() }
             },
             onError = { error ->
-                error?.let { showToast(it.message.toString()) }
+                showToast(error.message.toString())
```
andrews-moc, line 543: I made [the same change](https://github.com/salemove/android-sdk-widgets/pull/1307/files#diff-e9302efccb9d7391ea69b478fceca3c273d1eb69d236afe0184f8d6f76b25bf2R542) in my PR. 🙂 👍

DavDo (approved):

andrews-moc (approved):

Andrii-Horishnii-Glia:
!fixup

Andrii-Horishnii-Glia:
!squash

app/src/main/java/com/glia/exampleapp/MainFragment.kt
Andrii-Horishnii-Glia: OK, I'm removing it from my PR :)

PR #1321, opened 2025-05-21
Add visitor ID handling to push notification and authentication flow

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-4384

**What was solved?**

**Release notes:**

 - [x] Feature(PN)
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/di/GliaCoreImpl.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/di/GliaCoreImpl.kt
@@ -19,6 +19,7 @@ import com.glia.androidsdk.queuing.QueueTicket
 import com.glia.androidsdk.secureconversations.SecureConversations
 import com.glia.androidsdk.site.SiteInfo
 import com.glia.androidsdk.visitor.Authentication
+import com.glia.androidsdk.visitor.Visitor
 import com.glia.androidsdk.visitor.VisitorInfo
 import com.glia.androidsdk.visitor.VisitorInfoUpdateRequest
 import com.glia.widgets.engagement.MediaType
@@ -151,5 +152,11 @@ internal class GliaCoreImpl : GliaCore {
 
     override fun unsubscribeFromQueueUpdates(onError: Consumer<GliaException>?, callback: Consumer<Queue>) {
         Glia.unsubscribeFromQueueUpdates(onError, callback)
+    }
+
+    override fun getCurrentVisitor(onSuccess: (Visitor) -> Unit, onError: () -> Unit) {
+        Glia.getCurrentVisitor { visitor, gliaException ->
+            visitor?.let { onSuccess(it) } ?: onError()
+        }
     }
 }
--- a/widgetssdk/src/main/java/com/glia/widgets/di/GliaCoreImpl.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/di/GliaCoreImpl.kt
@@ -154,9 +154,7 @@ internal class GliaCoreImpl : GliaCore {
         Glia.unsubscribeFromQueueUpdates(onError, callback)
     }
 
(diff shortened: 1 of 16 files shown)


widgetssdk/src/main/java/com/glia/widgets/di/GliaCoreImpl.kt
```diff
@@ -152,4 +156,10 @@ internal class GliaCoreImpl : GliaCore {
     override fun unsubscribeFromQueueUpdates(onError: Consumer<GliaException>?, callback: Consumer<Queue>) {
         Glia.unsubscribeFromQueueUpdates(onError, callback)
     }
+
+    override fun getCurrentVisitor(onResult: OnResult<Visitor>, onError: OnError) {
+        Glia.getCurrentVisitor { visitor, gliaException ->
+            visitor?.let(onResult::onResult) ?: onError.onError(gliaException.toWidgetsType("Failed to get current visitor"))
```
andrews-moc, line 162: I think the idea is that `GliaCoreImpl` works with Core types for exception and callbacks. Types are converted to Widgets types in `GliaWidgets` class. Have a look at how the other functions are implemented. 😉

DavDo: Yes! But what's wrong with this code?
I just use these callbacks because they fit my use case. The `RequestCallback` is quite uncomfortable to use.

widgetssdk/src/main/java/com/glia/widgets/push/notifications/RequestPushNotificationDuringAuthenticationUseCase.kt
```diff
@@ -42,7 +46,16 @@ internal class RequestPushNotificationDuringAuthenticationUseCaseImpl(
             permissionManager.shouldShowPermissionRationale(Manifest.permission.POST_NOTIFICATIONS) -> {
                 // Show rationale dialog
                 uiComponentsDispatcher.showNotificationPermissionDialog(onAllow = {
+                    Logger.i(
+                        TAG,
+                        "Visitor's decision for intermediate push notification permission is 'allowed'"
```
andrews-moc, line 51: I also see the `Visitor's decision for push notification permission during auth. is 'allowed=$it'`. Does this mean that this event would be logged twice or these are two different places that will not happen at the same time?

widgetssdk/src/main/java/com/glia/widgets/di/GliaCoreImpl.kt
andrews-moc: Let me think...

andrews-moc: Yes, the only concern is regarding consistency. Your approach is good but it's not consistent with the way how other functions in `GliaCoreImpl` are implemented. The matter is that we don't need to expose `getCurrentVisitor()` for integrators through `GliaWidgets`. That's the reason why this confusion appears. Otherwise you might add it to `GliaWidgets` and use it with Widgets callbacks.

Anyway, I would prefer to keep consistency in `GliaCoreImpl` although it's a bit inconvenient, agree. It's solely my opinion.

DavDo: `GliaCoreImpl` is internal.
But I think, I now understand what you mean. At least I can change the `onError` callback to not map it to Widgets type and then just ignore.

andrews-moc: Yes, it's internal, but it didn't have `OnResult`, `OnError`, `toWidgetsType` imports so far.

DavDo: fixed!

andrews-moc (approved):

PR #1331, opened 2025-05-28
Update push notification channel names

gugalo:
**Jira issue:**

**What was solved?**
After discussing this topic during a meeting, we agreed that channels should not have the  words 'notification' or 'channel' in them since it is already clear from context, and that it should be in the plural form

Last screenshot from the discussion:
Screenshot 2025-05-23 at 12 03 39

How it looks now:
Screenshot_20250528_111306

### DIFF
--- a/widgetssdk/src/main/res/values/donottranslate.xml
+++ b/widgetssdk/src/main/res/values/donottranslate.xml
@@ -5,7 +5,7 @@
 
     <string name="android_notification_audio_call_channel_name" translatable="false">@string/glia_notification_call_channel_name</string>
     <string name="android_notification_screen_sharing_channel_name" translatable="false">@string/glia_notification_screen_sharing_channel_name</string>
-    <string name="android_notification_secure_messaging_channel_name" translatable="false">Messaging notification channel</string>
+    <string name="android_notification_secure_messaging_channel_name" translatable="false">New secure messages</string>
 
     <!-- Survey -->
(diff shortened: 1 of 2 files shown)


widgetssdk/src/main/res/values/donottranslate.xml
```diff
@@ -5,7 +5,7 @@
 
     <string name="android_notification_audio_call_channel_name" translatable="false">@string/glia_notification_call_channel_name</string>
     <string name="android_notification_screen_sharing_channel_name" translatable="false">@string/glia_notification_screen_sharing_channel_name</string>
-    <string name="android_notification_secure_messaging_channel_name" translatable="false">Messaging notification channel</string>
+    <string name="android_notification_secure_messaging_channel_name" translatable="false">New Secure Messages</string>
```
1toomas, line 8: I'm not sure if "Secure Messages" is name here or simply a text (i.e. should it be "New Secure Messages" or "New secure messages"). I asked other TWs and am waiting for their opinion.

1toomas (commented):
Here are my thoughts but I'm waiting for a second opinion.

widgetssdk/src/main/res/values/donottranslate.xml
gugalo: The feature name would be secure conversations and not the secure messages. So I guess it would be okay to go with "New secure messages"

PR #1338, opened 2025-06-09
Fix engagement end during de-authentication

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-4431

**What was solved?**
According to [documentation](https://docs.glia.com/glia-mobile/docs/ios-authentication#end-ongoing-engagement) for authentication, when the authentication behavior is `ALLOWED_DURING_ENGAGEMENT`, the engagement ends after the visitor becomes deauthenticated.
But widgets ends engagement regardless of behvaior type.

❗ Will add test for AuthManager in next commit.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/authentication/Authentication.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/authentication/Authentication.kt
@@ -33,12 +33,7 @@ interface Authentication {
      * <br></br> [GliaWidgetsException.Cause.INTERNAL_ERROR] - when internal error occurs
      * <br></br> [GliaWidgetsException.Cause.AUTHENTICATION_ERROR] - when authentication fails
      */
-    fun authenticate(
-        jwtToken: String,
-        externalAccessToken: String?,
-        onComplete: OnComplete,
-        onError: OnError
-    )
+    fun authenticate(jwtToken: String, externalAccessToken: String?, onComplete: OnComplete, onError: OnError)
 
     /**
      * De-authenticates the visitor.
@@ -53,21 +48,12 @@ interface Authentication {
      * <br></br> [GliaWidgetsException.Cause.INTERNAL_ERROR] - when internal error occurs
      * <br></br> [GliaWidgetsException.Cause.AUTHENTICATION_ERROR] - when authentication fails
      */
-    fun deauthenticate(
-        stopPushNotifications: Boolean,
-        onComplete: OnComplete,
-        onError: OnError
-    )
+    fun deauthenticate(stopPushNotifications: Boolean, onComplete: OnComplete, onError: OnError)
 
     /**
      * Same as [deauthenticate] but with default value `false` for `stopPushNotifications` parameter.
      */
-    fun deauthenticate(
-        onComplete: OnComplete,
-        onError: OnError
-    ) {
-        deauthenticate(false, onComplete, onError)
-    }
+    fun deauthenticate(onComplete: OnComplete, onError: OnError) = deauthenticate(false, onComplete, onError)
 
     /**
      * Check if Visitor is authenticated in Glia using the external authentication.
@@ -91,12 +77,7 @@ interface Authentication {
      * <br></br> [GliaWidgetsException.Cause.INTERNAL_ERROR] - when internal error occurs
      * <br></br> [GliaWidgetsException.Cause.AUTHENTICATION_ERROR] - when authentication fails
      */
-    fun refresh(
-        jwtToken: String,
-        externalAccessToken: String?,
-        onComplete: OnComplete,
-        onError: OnError
-    )
+    fun refresh(jwtToken: String, externalAccessToken: String?, onComplete: OnComplete, onError: OnError)
 
     /**
      * Behavior for authentication and de-authentication.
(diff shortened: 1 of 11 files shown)


widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
```diff
@@ -220,13 +220,14 @@ internal class EngagementRepositoryImpl(
             unsubscribeFromEvents(it)
             resetState()
 
-            it.end { ex -> ex?.also { Logger.d(TAG, "Ending engagement failed") } }
+            //End the current engagement only when it is ended by the visitor.
+            if (endedBy == EndedBy.VISITOR) {
+                it.end { ex -> ex?.also { Logger.d(TAG, "Ending engagement failed") } }
+            }
```
andrews-moc, line 226: So, the endEngagement() will not end engagement if it's called programmatically (not be visitor), do I understand correctly? Could you please describe the reason for this change?

widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
```diff
@@ -220,13 +220,14 @@ internal class EngagementRepositoryImpl(
...
 
-            val state = State.EngagementEnded(
-                it.isCallVisualizer,
-                endedBy,
-                it.state.actionOnEnd
-            ) { onSuccess, onError -> fetchSurvey(it, onError, onSuccess) }
+            val state = State.EngagementEnded(it.isCallVisualizer, endedBy, it.state.actionOnEnd) { onSuccess, onError ->
+                fetchSurvey(it, onError, onSuccess)
+            }
```
andrews-moc, line 230: So `state` will be set to `State.EngagementEnded` even in case `it.end` was not called (`endedBy` is not `EndedBy.VISITOR`). Looks unlogical to me. 🤔

widgetssdk/src/main/java/com/glia/widgets/internal/authentication/AuthenticationManager.kt
```diff
@@ -27,65 +27,56 @@ internal class AuthenticationManager(
...
-                         onError: OnError
-    ) {
+    override fun refresh(jwtToken: String, externalAccessToken: String?, onComplete: OnComplete, onError: OnError) {
         Logger.i(TAG, "Refresh authentication")
+
         authentication.refresh(jwtToken, externalAccessToken) { _, gliaException ->
             if (gliaException != null) {
                 onError.onError(gliaException.toWidgetsType())
+            } else {
+                onComplete.onComplete()
```
andrews-moc, line 78: 👍

widgetssdk/src/main/java/com/glia/widgets/authentication/Authentication.kt
```diff
@@ -33,12 +33,7 @@ interface Authentication {
      * <br></br> [GliaWidgetsException.Cause.INTERNAL_ERROR] - when internal error occurs
      * <br></br> [GliaWidgetsException.Cause.AUTHENTICATION_ERROR] - when authentication fails
      */
-    fun authenticate(
-        jwtToken: String,
-        externalAccessToken: String?,
-        onComplete: OnComplete,
-        onError: OnError
-    )
+    fun authenticate(jwtToken: String, externalAccessToken: String?, onComplete: OnComplete, onError: OnError)
```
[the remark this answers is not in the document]
DavDo: but it fits our configurations `max_line_length = 150`

widgetssdk/src/main/java/com/glia/widgets/internal/authentication/AuthenticationManager.kt
```diff
@@ -27,65 +27,56 @@ internal class AuthenticationManager(
...
+
         authentication.authenticate(jwtToken, externalAccessToken) { _, gliaException ->
+            Dependencies.controllerFactory.pushClickHandlerController.onAuthenticationAttempt()
+
             if (gliaException != null) {
                 onError.onError(gliaException.toWidgetsType())
-            }
-            //Here we need to subscribe to secure conversations repository to get the data for authenticated visitors
-            repositoryFactory.secureConversationsRepository.subscribe()
+            } else {
```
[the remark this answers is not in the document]
DavDo: We need to call this function for both cases, to reset PN state when authentication fails.

widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
DavDo: I like the idea, but I think that's not in the scope of this fix.
I will add tests to this PR, so we can include this fix to 3.1.1 release.
Sine the suggested improvements will touch multiple layers with business logic, I'll create separate PR for that, this way we will test changes until the next release.

widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepositoryImpl.kt
andrews-moc: I think I was missing that in all other cases engagement is already ended. This function needs to call `it.end()` ONLY if visitor initiated this action(event). So should be fine.

andrews-moc (approved):
I'm approving but it might be good to test Secure Conversations + deauthentication + push notifications manually. 

The changes in these classes are the main focus for test cases:
- EngagementRepositoryImpl.kt
- AuthenticationManager.kt

DavDo:
ATs with this custom build succeeded.
I also tested manually and everything works as expected.

PR #1341, opened 2025-06-09
Partially revert the fix for engagement during deauthentication

andrews-moc:
**What was solved?**
1. Revert "Fix engagement end during de-authentication" (separate PR)
2. Move the authenticate(), deauthenticate() and refresh() functions code that is related to a successful operation to the 'else' block to avoid it's execution in case of an error.

ℹ️ Engagement is ended on de-authentication even if behavior is `FORBIDDEN_DURING_ENGAGEMENT`. This fix needs more careful changes and will be delivered separetely.

**Release notes:**

Move the authenticate(), deauthenticate() and refresh() functions code that is related to a successful operation to the 'else' block to avoid it's execution in case of an error.

 - [ ] Feature
 - [ ] Ignore

**Additional info:**


DavDo (approved):

DavDo:
Added tests for `AuthenticationManager.kt` and move PushClickHandler call out of the success block

andrews-moc:
> Added tests for `AuthenticationManager.kt` and moved `PushClickHandler` call out of the success block

Thank you @DavDo! Looks good to me. 👍

PR #1345, opened 2025-06-11
Refactor: Update engagement end handling and remove EndedBy references

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-4431

> [!NOTE]
> The tests are in a separate commit. This makes it easier to review them separately.

**What was solved?**
The functions for handling end-engagement requests initiated by visitors and integrators have been separated.
The engagement-end handling has also been improved.

**Release notes:**
Improved the end engagement experience during de-authentication.

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**


Andrii-Horishnii-Glia (approved):

widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepository.kt
```diff
@@ -47,7 +47,10 @@ internal interface EngagementRepository {
     fun initialize()
     fun reset()
     fun resetQueueing()
-    fun endEngagement(endedBy: EndedBy)
+    //used for ending engagement after visitor ends engagement explicitly
+    fun endEngagement()
+    //used for ending engagement after integrator ends the engagement from outside of SDK, for example [GliaWidgets.endEngagement]
```
andrews-moc, line 51: The comment is a bit confusing to me because it says `from outside of SDK` but mentions `[GliaWidgets.endEngagement]` which is an SDK function. Maybe it makes sense to rephrase it.

widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepository.kt
```diff
@@ -47,7 +47,10 @@ internal interface EngagementRepository {
...
+    fun endEngagementSilently()
```
andrews-moc, line 52: Regarding the function name. I'm thinking if we can find something more meaningful instead of `Silently`? Do I understand correctly that this function is intended to be called after an engagement is actually ended and Widgets SDK just needs to update UI, etc? If so, maybe an alternative name could be `onEngagementEnded()` or whatnot? 🤔

andrews-moc: Ah, you see - I misunderstood it looking at its name... 🤔  I would expect something like `forceEndEngagement()`, not sure if it's only my brain's problem. Maybe some additional comment with this info might be useful.

andrews-moc (approved):
Good job. I left some comments, they are not critical, so I'm approving the PR.

.idea/codeStyles/Project.xml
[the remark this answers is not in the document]
andrews-moc: Ah, you are right, sorry.

widgetssdk/src/main/java/com/glia/widgets/engagement/EngagementRepository.kt
DavDo: 🤔

DavDo: `terminateEngagement()`?

andrews-moc: I'm kay with `terminateEngagement`.

Having looked at the edited comments, I've got an alternative idea - `endEngagementByVisitor()` and `endEngagementByIntegrator()`.

But it's up to you. I just shared an idea, you can ignore it.

DavDo: I'd prefer `terminateEngagement` because I think it describes the nature of the function. `endEngagementByVisitor()` and `endEngagementByIntegrator()` are more about the reason for ending the engagement.

andrews-moc: Okay. 👍

PR #1352, opened 2025-07-02
Run automated AndroidStudio code refactoring tool

gugalo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-4490

**What was solved?**
It takes into account setting in .editorconfig file

I reverted some of the automated changes (less then 10%) because they made code less readable.

**PS:** You don't have to review everything line by line, because all the changes follow the same schemas defined in .editorconfig and default Android settings. But you should try to review the types/cases instead 

**Additional info:**

**Screenshots:**


andrews-moc (approved):
You haven't added labels to PR, but I also still forget doing this. 🙂

PR #1372, opened 2025-08-27
Add the GliaWidgets.init() error causes

Andrii-Horishnii-Glia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-4590

**What was solved?**
Added additional error causes to understand better the reason why the initialization failed.

The Core SDK causes:
- `NETWORK_TIMEOUT`: any network issue,
- `INVALID_INPUT`: wrong credentials (the [changes on the Core SDK](https://github.com/salemove/android-sdk/pull/879) are needed),
- `FORBIDDEN`:  if credentials are correct, but they don't have enough permissions for other initial requests.

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.kt
@@ -173,10 +173,32 @@ object GliaWidgets {
                         onComplete.onComplete()
                     } else {
                         Logger.i(TAG, "Glia Widgets SDK initialization failed")
-                        val invalidInputError = GliaWidgetsException(
-                            "Failed to initialise Glia Widgets SDK. Please check credentials.",
-                            GliaWidgetsException.Cause.INVALID_INPUT
-                        )
+                        val invalidInputError = when (exception.cause) {
+                            GliaException.Cause.NETWORK_TIMEOUT -> {
+                                GliaWidgetsException(
+                                    "Network timeout. Please check the Internet connection.",
+                                    GliaWidgetsException.Cause.NETWORK_TIMEOUT
+                                )
+                            }
+                            GliaException.Cause.INVALID_INPUT -> {
+                                GliaWidgetsException(
+                                    "Failed to initialise Glia Widgets SDK. Invalid input. Please check credentials.",
+                                    GliaWidgetsException.Cause.INVALID_INPUT
+                                )
+                            }
+                            GliaException.Cause.FORBIDDEN -> {
+                                GliaWidgetsException(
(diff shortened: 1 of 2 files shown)


widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.kt
```diff
@@ -173,10 +173,29 @@ object GliaWidgets {
...
+                                GliaWidgetsException(
+                                    "Failed to initialise Glia Widgets SDK. Please check credentials.",
+                                    GliaWidgetsException.Cause.INVALID_INPUT
+                                )
+                            }
+
+                            else -> {
+                                GliaWidgetsException(
+                                    "Failed to initialise Glia Widgets SDK. Please check credentials.",
+                                    GliaWidgetsException.Cause.INTERNAL_ERROR
```
gugalo, line 195: I might be mistaken, but I think SDK should not throw `GliaWidgetsException.Cause.INTERNAL_ERROR` for internal errors. As mentioned in casse heigher it should be `GliaWidgetsException.Cause.INVALID_INPUT`. Maybe we could change this error message?
```suggestion
                                GliaWidgetsException(
                                    "Failed to initialise Glia Widgets SDK. Internal error, please check logs.",
                                    GliaWidgetsException.Cause.INTERNAL_ERROR
```

DavDo (approved):

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.kt
```diff
@@ -173,10 +173,29 @@ object GliaWidgets {
...
+                        val invalidInputError = when (exception.cause) {
+                            GliaException.Cause.NETWORK_TIMEOUT -> {
+                                GliaWidgetsException(
+                                    "Network timeout. Please check the Internet connection.",
+                                    GliaWidgetsException.Cause.NETWORK_TIMEOUT
+                                )
+                            }
+
+                            GliaException.Cause.INVALID_INPUT,
+                            GliaException.Cause.FORBIDDEN -> {
```
andrews-moc, line 185: Could you please explain why you decided to add `GliaException.Cause.FORBIDDEN`?

widgetssdk/src/main/java/com/glia/widgets/GliaWidgets.kt
Andrii-Horishnii-Glia: Unfortunately, conditions above don't cover all credential cases.
Also, it is not a good idea to direct integrators to logs, because logs say nothing.

andrews-moc:
!squash

PR #1379, opened 2025-09-09
Refactor subscriptions to use safe subscribe method

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-4638

**What was solved?**

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/controller/CallVisualizerController.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/callvisualizer/controller/CallVisualizerController.kt
@@ -1,5 +1,6 @@
 package com.glia.widgets.callvisualizer.controller
 
+import android.annotation.SuppressLint
 import com.glia.widgets.engagement.EndAction
 import com.glia.widgets.engagement.domain.EngagementRequestUseCase
 import com.glia.widgets.engagement.domain.EngagementStateUseCase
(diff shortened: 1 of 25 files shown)


widgetssdk/src/main/java/com/glia/widgets/callvisualizer/controller/CallVisualizerController.kt
```diff
@@ -68,11 +68,12 @@ internal class CallVisualizerController(
...
 
+    @SuppressLint("CheckResult")
     private fun registerCallVisualizerListeners() {
-        engagementRequestUseCase().unSafeSubscribe { onEngagementRequested() }
-        onIncomingEngagementRequestTimeoutUseCase().unSafeSubscribe { onIncomingEngagementRequestTimeout() }
+        engagementRequestUseCase().subscribe { onEngagementRequested() }
+        onIncomingEngagementRequestTimeoutUseCase().subscribe { onIncomingEngagementRequestTimeout() }
         dialogController.addCallback(::handleDialogState)
-        engagementStartFlow.unSafeSubscribe { dismissVisitorCodeDialog() }
+        engagementStartFlow.subscribe { dismissVisitorCodeDialog() }
```
andrews-moc, line 76: I expected to see adding to disposable. Seems you just switched from `unSafeSubscribe()` to `subscribe()`, isn't it?

PR #1385, opened 2025-09-15
Disable item animator for chat RecyclerView to avoid crashes.

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-4595

**What was solved?**
This PR resolves a `RecyclerView` crash `(IllegalArgumentException: Tmp detached view)`. The crash occurred due to a race condition between item animations and data updates. The fix disables the `ItemAnimator` by setting it to `null`, which prevents the conflicting state and stabilizes the view.

**Additional info:**
**Disabled Animation**
[Screen_recording_20250916_095703.webm](https://github.com/user-attachments/assets/b44bf417-ba49-47e8-ac31-b01116390a75)

**Enabled Animation**
[Screen_recording_20250916_095938.webm](https://github.com/user-attachments/assets/edaf9c4e-d720-4500-b01e-5be00d900269)

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/ChatView.kt
@@ -348,6 +348,7 @@ internal class ChatView(context: Context, attrs: AttributeSet?, defStyleAttr: In
     }
 
     override fun emitItems(items: List<ChatItem>) {
+        binding.chatRecyclerView.itemAnimator?.endAnimations()
         adapter.submitList(items)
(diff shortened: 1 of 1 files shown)


DavDo:
> > Added videos to the description, to me the difference is minimal without visible visual effects.
> 
> With animation, it looks smooth. I can't say the difference is minimal.

Is the difference significant enough to justify not releasing this change?

Andrii-Horishnii-Glia:
Let's ask the designer. Raise this topic on slack.

PR #1504, opened 2026-05-04
Fix IllegalStateException on operator transfer in ChatRecyclerView

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-5208

**What was solved?**
During an engagement transfer in chat, the previous operator's Connected and Joined chat history items are removed when the Transferring status is inserted. Additionally, all OperatorStatusItem subtypes share the same stable id (operator_status_item) crashes application with the IllegalStateExecption.

- Added UUIDs for OperatorStatusItems
- Now the old OperatorStatusItems preserve in chat when the "TransFerring" appears. 

**Additional info:**

**Screenshots:**

### DIFF
--- a/widgetssdk/src/main/java/com/glia/widgets/chat/ChatManager.kt
+++ b/widgetssdk/src/main/java/com/glia/widgets/chat/ChatManager.kt
@@ -437,9 +437,13 @@ internal class ChatManager(
 
     @VisibleForTesting
     fun mapTransferring(state: State): State = state.apply {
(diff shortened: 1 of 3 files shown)


Andrii-Horishnii-Glia (approved):
It seems that solves the issues that I discussed with the designers.

DavDo:
> It seems that solves the issues that I discussed with the designers.

Seems so! I didn't knew that ticket existed.

PR #1508, opened 2026-06-17
Add CallForegroundService to protect active call media process in background

Andrii-Horishnii-Glia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-5356

**What was solved?**
Removes the API 31+ early-return from showAudioCallNotification/showVideoCallNotification (introduced in ee596aca as a crash workaround for MOB-3358) and replaces it with CallForegroundService — a foreground service with type microphone|camera that starts when a call connects and stops when it ends. This keeps the media process alive when the app is backgrounded, fixing audio cutout on Samsung and other aggressive OEM devices.

**Additional info:**

**Screenshots:**


DavDo (approved):
Good job!

PR #1519, opened 2026-07-23
Increment Telemetry lib version to 1.0.6

sm-deployer:


DavDo (approved):

DavDo:
The release has compileSdk 37 which is incompatible with the current Widgets version

PR #1522, opened 2026-07-29
Restore Paparazzi snapshot test execution after AGP 9 migration

DavDo:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-5458

**What was solved?**

After the AGP 9.3 migration the Paparazzi snapshot tests silently stopped running, in two independent ways:

1. AGP 9 creates unit-test components only for the single `android.testBuildType` (default `debug`), so the snapshot-variant tasks (`verifyPaparazziSnapshot` / `recordPaparazziSnapshot`) were no longer registered.
2. The Bitrise step upgrade to `gradle-unit-test@2` renamed its inputs (`unit_test_task` → `test_task`), so the old input was silently ignored and CI ran the plain `test` task — reporting green while verifying zero snapshots.

This PR restores honest snapshot verification:

- `testBuildType` is now switchable via `-PtestBuildType` or `local.properties` (the latter enables Android Studio IDE support for `src/testSnapshot`), defaulting to `debug`.
- New `verifySnapshots` / `recordSnapshots` wrapper tasks run the real Paparazzi tasks in a nested build with the property pre-set — same commands locally and in CI.
- Fixed the unit-test exclusion for the snapshot variant: under AGP 9 built-in Kotlin, `test.kotlin.srcDirs` must be cleared alongside `test.java.srcDirs`.
- Golden images moved to `src/testSnapshot/snapshots` (Paparazzi alpha05 derives the golden dir from the test source set); Git LFS pattern updated; all goldens re-recorded (432/442 changed by benign layoutlib anti-aliasing shifts, visually verified on sample deltas).
- `SnapshotAttachment` fixed for Core 4.0.1 `AttachmentFile` (new `url` and `interactionEntryIdentifiers` members).
- Bitrise snapshot step points at `verifySnapshots` using the v2 input schema; removed the Paparazzi HTML report workaround (fixed in alpha05, and the CI report export depends on the HTML report).
- Bitrise stack updated to `ubuntu-latest-stable` on `g2.linux.medium`.

Verified locally: `verifySnapshots` runs 434 snapshot tests with 0 failures; the debug unit-test suite (1070 tests) is unaffected.

**Additional info:**

**Screenshots:**

N/A — golden image changes are anti-aliasing-level only (no visual layout changes).


Andrii-Horishnii-Glia (commented):
Could you make git move instead of git remove and add for those files?
```
widgetssdk/src/test/snapshots/images/*.png
widgetssdk/src/testSnapshot/snapshots/images/*.png
```

Andrii-Horishnii-Glia (approved):
Looks good!
It would be great to see file renamings instead of removing and adding the same ones.

DavDo:
> Could you make git move instead of git remove and add for those files?

Done — split into two commits: a pure move (f045456f, all 442 files show as renames) and the re-record (5ab8893d, content changes at the new paths).

Note: the per-commit view shows the move correctly, but the PR-level "Files changed" tab will still show most snapshot images as removed+added — 432 of 442 were re-recorded after the move (layoutlib rendering changed in the Paparazzi upgrade), and since these are LFS pointer files, any content change rewrites the whole pointer, so GitHub's similarity-based rename detection can't pair them.

